PR #11, opened 2024-08-22
chore(draft): switch to Rolldown

AdrianGonz97:
This PR replaces Rollup for Rolldown. Currently a **WIP**, but it's in a functional state (locally).

## Notes
- Rolldown doesn't seem to know how to resolve import paths that use `.js` extensions to reference a TS file. Using a `.ts` extension seems to work instead, but for now, I just removed all extensions
- Packages that are written in CJS needed to be externalized as Rolldown doesn't transform `require` calls into ESM imports at this time (tracking: https://github.com/rolldown/rolldown/issues/2041)
- Dynamic import vars aren't fully supported yet, so they've been replaced for normal dynamic imports with static paths

## TODO
- [ ] Rolldown doesn't have a built-in watcher yet, so we'll have to create our own for the time being (tracking: https://github.com/rolldown/rolldown/issues/2028)
- [ ] Investigate why Rolldown is transforming `dedent` into malformed code (it's being externalized for now)
- [ ] Move `ast-manipulation` and `ast-tooling` directly into `core`

### DIFF
--- a/rolldown.config.js
+++ b/rolldown.config.js
@@ -19,19 +19,13 @@ function getConfig(project) {
 	if (project === 'core') inputs.push(`./packages/${project}/internal.ts`);
 
 	const projectRoot = path.resolve(path.join(outDir, '..'));
-	// fs.rmSync(outDir, { force: true, recursive: true });
+	fs.rmSync(outDir, { force: true, recursive: true });
 
 	/** @type {PackageJson} */
 	const pkg = JSON.parse(fs.readFileSync(path.join(projectRoot, 'package.json'), 'utf8'));
 	const externalDeps = getExternalDeps(pkg);
 
 	const external = [
-		// As of now, Rolldown doesn't transform cjs `require` calls into esm imports, so we need to
-		// manually externalize packages that use cjs.
-		/commander/,
-		/picocolors/,
-		/postcss/,
-		/recast/,
 		// TODO: investigate why Rolldown is transforming dedent into malformed code
 		/dedent/,
 		/@svelte-cli\/core/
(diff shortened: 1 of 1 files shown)


hyf0:
Happen to saw this PR. Happy to share some updates and thoughts.

> Rolldown doesn't seem to know how to resolve import paths that use .js extensions to reference a TS file.

This is supported in nightly version.

> Packages that are written in CJS needed to be externalized as Rolldown doesn't transform require calls into ESM imports at this time

This is solved too. We now polyfill `require` if the format is `cjs` and platform is `node`.

> Dynamic import vars aren't fully supported yet, so they've been replaced for normal dynamic imports with static paths

We have a builtin plugin to do it. But not sure if it fits this use case. 

https://github.com/rolldown/rolldown/blob/de64ad96e4c827cf0908e64ca879a9c8837c9869/packages/rolldown/src/experimental-index.ts#L9

> Rolldown doesn't have a built-in watcher

Yeah. Watcher is supported.

> however i have some reservations against using rolldown to build sv while it is still under rapid prerelease development. Unless we have tests that ensure the output is always correct and we don't accidentally release a version of sv that is broken due to a change in rolldown, this would be very risky

@dominikg

Yeah. This is a very legit worry.  We have done some process to ensure correctness, such as that rolldown is currently built by itself for every commit and run all tests.

We are planning a 1.0.0 beta release this month and certainly we would have a ecosystem CI to run before publishing any new version.

The ecosystem CI will clone the repo that uses rolldown, run the build command and test command.

Not intend to pushing anything. But if you decide to switched to rolldown, we surely will add this repo as a part of ecosystem CI.

PR #100, opened 2024-10-13
chore(DRAFT): use `acorn`, `estree` and `esrap`

### LINKED ISSUE #44: File created with mix of single and double quotes
Can we just use single quotes everywhere?

The Drizzle adder creates the `drizzle.config.ts` below currently:
```
if (!process.env.DATABASE_URL) throw new Error('DATABASE_URL is not set');

export default defineConfig({
  schema: "./src/lib/server/db/schema.ts",

  dbCredentials: {
    url: process.env.DATABASE_URL
  },

  verbose: true,
  strict: true,
  dialect: "sqlite"
});
```

### LINKED ISSUE #94: Use `{ parse } from 'svelte/compiler'`
recast is using babel, which is much less used in the Svelte ecosystem, so folks wouldn't generally be as familiar with it

manuel3108:
Closes #94 
Closes #44

Benefits
* closer to other tools in the `svelte` ecosystem
* Faster bundling (~33% speed improvement locally from 3s to 2s)
* Smaller bundle (not tested, but dropping babel should dramatically decrease the bundle size)
* more consistent formatting
* we are in full control of the printing, as we own `esrap`

Downsides
* Potentially to consistent formatting, as it looks like it's completely ignore the user input (tabs vs spaces, etc), but that's probably fixable in `esrap`
* Custom typing for special typescript types (see packages/ast-tooling/ts-estree.ts) and https://github.com/Rich-Harris/esrap/pull/13 for reasoning and considered alternatives.

Depends on https://github.com/Rich-Harris/esrap/pull/13. Currently published as `esrap-typescript-temp` for ease of use for testing. Otherwise, this would require using `pnpm link` locally with the branch mentioned before.

If we want to proceed with this (which i think we should), this needs to be done
* [ ] publish `packages/ast-tooling/ts-estree.ts` seperately so that it can also be used inside `esrap`. Later they should probably live inside `acorn-typescript` once we fork it
* [ ] make `esrap` understand what a `TSModuleDeclaration` is and how to print it
* [ ] update types inside `esrap` according to the types above. This will probably also require way more types in this new package
* [ ] merge the pr inside `esrap` and publish it
* [ ] fix todos
* [ ] Testing. I have done a few tests here and there and generally the formatting is way more consistent and matches what i would expect.

### DIFF
--- a/packages/adders/drizzle/index.ts
+++ b/packages/adders/drizzle/index.ts
@@ -229,8 +229,8 @@ export default defineAdder({
 						url: common.expressionFromString('process.env.DATABASE_URL'),
 						authToken
 					}),
-					verbose: { type: 'BooleanLiteral', value: true },
-					strict: { type: 'BooleanLiteral', value: true },
+					verbose: { type: 'Literal', value: true },
+					strict: { type: 'Literal', value: true },
 					driver
 				});
 
--- a/packages/adders/eslint/index.ts
+++ b/packages/adders/eslint/index.ts
@@ -9,7 +9,6 @@ import {
 	functions,
 	imports,
 	object,
-	type AstKinds,
 	type AstTypes
 } from '@sveltejs/cli-core/js';
 import { parseJson, parseScript } from '@sveltejs/cli-core/parsers';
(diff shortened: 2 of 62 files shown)


manuel3108:
Closing, as this PR is hopelessly outdated. The esrap TS PR has been merged, but esrap is just not ready enough for handling such stuff, especially regarding tabs vs spaces and single vs double quoutes.

PR #177, opened 2024-10-19
docs: fix link to SvelteKit docs

kevinji:
The original link returns a 404.


benmccann:
Thanka. This will be the new URL soon. I thought the site would go live slightly sooner 😆 But I'll close this since the link will fix itself in a couple of days

PR #181, opened 2024-10-20
chore(POC): simpler / more flexible api

### LINKED ISSUE #85: Consider a simpler API
It feels like we can make the API simpler. Why isn't the community template just this, for example?

```js
import { defineIntegration } from '@sveltejs/cli/integration';

export default defineIntegration({
  description: 'An adder template demo',
  environments: { kit: true, svelte: true },
  options: {
    demo: {
      question: 'Do you want to use a demo?',
      type: 'boolean',
      default: false
    }
  },
  files: ({ options }) => [
    {
      name: 'adder-template-demo.txt',
      content: (file) => {
        if (options.demo) {
          return 'This is a text file made by the integration template demo!'
        }

        return file.content;
      }
    },
    {
      name: 'src/DemoComponent.svelte',
      content: (file) => {
        file.addImport('../adder-template-demo.txt?raw', 'default', 'Demo');
      }
    {
  ]
});
```

Things to note:

- it's a single file. To me this is the difference between 'oh hey I can build one of these!' and 'ugh. strap in, I guess'. It probably doesn't even need a `src` directory
- `id` and `name` have been removed. They don't need to be separate concepts — it's less confusing if we just use an `id` consistently, instead of leaving people who see 'Foo Bar' listed somewhere to wonder whether it's `foobar` or `foo-bar` or whatever. And for stuff on `npm`, the package name _is_ the `id`, I'd argue we don't need another (that could potentially be something different than the package name, which is another source of confusion). _Maybe_ there's an argument for mandating an `id` for programmatic usage but I think it's probably unnecessary
- `description` and `environments` moved up to the top level instead of being in `metadata`
- no `packages` array. Surely `package.json` can be just like any other JSON file? The smaller the API surface area the better
- `files` is a function that returns an array, instead of an array of objects. That way it's easier to put logic in one place (e.g. you can do `const ext = options.typescript ? 'ts' : 'js'` once instead of in each `name` method, and easier to return a different set of files entirely under certain conditions
- `contentType` is inferred from file extension (I don't know if this is possible to do reliably from a types perspective, just an idea)
- utilities like `addImport` are passed in rather than imported. The big downside here is that it _slightly_ discourages you from extracting the logic out into separate functions, but I'd argue a) this is a context where less indirection is better anyway, and b) it's outweighed by the benefits of discoverability and not having to deal with the exposed guts of things like `jsAst` unless you absolutely have to

Related note: What's the technical reason for the prohibition on dependencies? Seems very restrictive

manuel3108:
Closes #85 

This is a POC to start visualizing stuff and initiate discussions. Besides of some new types & adapting the `paraglide` adder, I haven't done any implementation. So expect plenty of type errors and failing pipelines.

Up until now I was keeping my mouth shut about #85 as I had many doubts if this was the right way going forward. Also I had a hard time imagining how this would look in a real adder. Now that this is done with this POC it was eye opening how many flexibilities this will bring us & the community. Besides copying some code, there are nearly zero changes that affect the functionality of the adders. Since most of the functionality already exists it should be straight forward to implement this, once we come up with the "final" api. Most of the work probably needs to be done inside the adders, as there is lots of stuff that needs to be moved.

### DIFF
--- a/packages/adders/paraglide/index.ts
+++ b/packages/adders/paraglide/index.ts
@@ -10,27 +10,11 @@ import {
 	object,
 	variables,
 	exports,
-	kit
+	kit as kitJs
 } from '@sveltejs/cli-core/js';
 import * as html from '@sveltejs/cli-core/html';
 import { parseHtml, parseJson, parseScript, parseSvelte } from '@sveltejs/cli-core/parsers';
 import { addToDemoPage } from '../common.ts';
-
-const DEFAULT_INLANG_PROJECT = {
-	$schema: 'https://inlang.com/schema/project-settings',
-	modules: [
-		'https://cdn.jsdelivr.net/npm/@inlang/message-lint-rule-empty-pattern@1/dist/index.js',
-		'https://cdn.jsdelivr.net/npm/@inlang/message-lint-rule-identical-pattern@1/dist/index.js',
-		'https://cdn.jsdelivr.net/npm/@inlang/message-lint-rule-missing-translation@1/dist/index.js',
-		'https://cdn.jsdelivr.net/npm/@inlang/message-lint-rule-without-source@1/dist/index.js',
-		'https://cdn.jsdelivr.net/npm/@inlang/message-lint-rule-valid-js-identifier@1/dist/index.js',
-		'https://cdn.jsdelivr.net/npm/@inlang/plugin-message-format@2/dist/index.js',
-		'https://cdn.jsdelivr.net/npm/@inlang/plugin-m-function-matcher@0/dist/index.js'
-	],
-	'plugin.inlang.messageFormat': {
-		pathPattern: './messages/{languageTag}.json'
-	}
-};
 
 export const options = defineAdderOptions({
 	availableLanguageTags: {
@@ -66,20 +50,28 @@ export default defineAdder({
 	environments: { svelte: false, kit: true },
 	homepage: 'https://inlang.com',
 	options,
-	packages: [
-		{
-			name: '@inlang/paraglide-sveltekit',
-			version: '^0.11.1',
-			dev: false
-		}
-	],
-	files: [
-		{
-			// create an inlang project if it doesn't exist yet
-			name: () => 'project.inlang/settings.json',
-			condition: ({ cwd }) => !fs.existsSync(path.join(cwd, 'project.inlang/settings.json')),
-			content: ({ options, content }) => {
+	run: ({ files, cwd, options, typescript, kit, packages, dependencyVersion }) => {
+		packages.dependency('@inlang/paraglide-sveltekit', '^0.11.1');
+
+		if (!fs.existsSync(path.join(cwd, 'project.inlang/settings.json'))) {
+			files.update('project.inlang/settings.json', (content) => {
 				const { data, generateCode } = parseJson(content);
+
+				const DEFAULT_INLANG_PROJECT = {
+					$schema: 'https://inlang.com/schema/project-settings',
+					modules: [
+						'https://cdn.jsdelivr.net/npm/@inlang/message-lint-rule-empty-pattern@1/dist/index.js',
+						'https://cdn.jsdelivr.net/npm/@inlang/message-lint-rule-identical-pattern@1/dist/index.js',
+						'https://cdn.jsdelivr.net/npm/@inlang/message-lint-rule-missing-translation@1/dist/index.js',
+						'https://cdn.jsdelivr.net/npm/@inlang/message-lint-rule-without-source@1/dist/index.js',
+						'https://cdn.jsdelivr.net/npm/@inlang/message-lint-rule-valid-js-identifier@1/dist/index.js',
+						'https://cdn.jsdelivr.net/npm/@inlang/plugin-message-format@2/dist/index.js',
+						'https://cdn.jsdelivr.net/npm/@inlang/plugin-m-function-matcher@0/dist/index.js'
+					],
+					'plugin.inlang.messageFormat': {
+						pathPattern: './messages/{languageTag}.json'
+					}
+				};
 
 				for (const key in DEFAULT_INLANG_PROJECT) {
 					data[key] = DEFAULT_INLANG_PROJECT[key as keyof typeof DEFAULT_INLANG_PROJECT];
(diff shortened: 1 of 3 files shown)


packages/adders/paraglide/index.ts
```diff
@@ -10,28 +10,12 @@ import {
 	object,
 	variables,
 	exports,
-	kit
+	kit as kitJs
```
manuel3108, line 13: This is bad, but works for the moment. This is required since our `workspace` has a `kit` property and this import is also named `kit`

packages/adders/paraglide/index.ts
```diff
@@ -66,21 +50,29 @@ export default defineAdder({
...
-			dev: false
-		}
-	],
-	files: [
-		{
-			// create an inlang project if it doesn't exist yet
-			name: () => 'project.inlang/settings.json',
-			condition: ({ cwd }) => !fs.existsSync(path.join(cwd, 'project.inlang/settings.json')),
-			content: ({ options, content }) => {
+	run: ({ api, cwd, options, typescript, kit, dependencyVersion }) => {
```
manuel3108, line 53:
```suggestion
	run: ({ files, packages, scripts, cwd, options, typescript, kit, dependencyVersion }) => {
```
Open for discussion: Only one export, or multiple? I had implemented both locally, but decided to go with `api` as `files` and `scripts` only had one property. It might be more future proof, if we export all three things separately.

AdrianGonz97: i would probably name them as just `sv` and `file`. `sv` would encapsulate `packages` and `scripts` and `file` would just be for file handling. 

but maybe having a unified `sv` that does it all isn't so bad... hmmm, need to think on it.

AdrianGonz97:
this really does look and feel much nicer!!

there are still things i would love to improve after this, such as using an ast manipulating api like this:
```js
const { script, css, generateCode } = parseSvelte(content);
script.importNamed('../adder-template-demo.txt?raw', { demo: 'Demo' });
css.addClass('foobar', { 'background-color': 'blue' });
return generateCode({ script: script.generateCode(), css: css.generateCode() });
```

after that, i think we'll be super close to having a library that's ready for community integrations

packages/adders/paraglide/index.ts
manuel3108: interesting approach to just name it `sv`, i like that. 
Although i wouldn't be sure about combining `packages` and `scripts` and leaving `files` separately. I would suggest that we do all or nothing here.

manuel3108:
Regarding that bit of code that you removed back to the top: I moved it on purpose because it's only used in that one place and nowhere else, so i thought it was unnecessary polluting global stuff. But whatever works for you, i don't have a strong opinion on that.

> ```js
> const { script, css, generateCode } = parseSvelte(content);
> script.importNamed('../adder-template-demo.txt?raw', { demo: 'Demo' });
> css.addClass('foobar', { 'background-color': 'blue' });
> return generateCode({ script: script.generateCode(), css: css.generateCode() });
> ```

I agree, there is lot's of improvement potential in the ast apis as well. Overall i like where this would be going. But as you said, one step at a time.

AdrianGonz97:
> Regarding that bit of code that you removed back to the top: I moved it on purpose because it's only used in that one place and nowhere else, so i thought it was unnecessary polluting global stuff. But whatever works for you, i don't have a strong opinion on that.

yea, i much prefer keeping actual constants in the top scope, especially in this case so it's not polluting the sections with logic

packages/adders/paraglide/index.ts
AdrianGonz97: the issue with not letting it have it's own namespace is that each method now gets a bit more verbose

ex:
```js
sv.updateFile
sv.createOrUpdateFile
sv.createFile
```
as opposed to:
```js
file.update
file.createOrUpdate
sv.create
```

AdrianGonz97:
so one issue with this new api is that we'll no longer be able to determine if an integration has been previously installed as the packages are no longer static values that we can easily examine. 

see:
https://github.com/sveltejs/cli/blob/4812956d80a2d38e034da0a91a756e517af63871/packages/cli/commands/add/index.ts#L309-L327

the `lucia` integration relies on this to determine whether we should prompt to also install `drizzle`. this prevents the `drizzle` adder from being needlessly executed again if it's already installed.

not exactly sure how this can be worked around yet, but this is something we definitely needs to solve

packages/adders/paraglide/index.ts
manuel3108: But wouldn't we have exactly the same problem with `script` and `package`?
Additionally, `package` is a reserved keyword in strict mode, which would make it necessary to use the plural or something completely different like `dependencies`

manuel3108:
> so one issue with this new api is that we'll no longer be able to determine if an integration has been previously installed as the packages are no longer static values that we can easily examine.

I wasn't aware anymore that we had this in place.

I was initially going to suggest something like this
```js
run: ({ sv, cwd, options, typescript, kit, dependencyVersion }) => {
    if(!!dependencyVersion('drizzle-orm'))
        sv.dependsOn('drizzle)

    // do other file updates
}
```

That would increase flexibility as well. But this won't work because we will only be able to determine the integrations it depends on while executing the adder. But in general I do think a functional approach like this would be best, maybe by converting our existing `dependsOn` field into a function and basically doing the same thing as above.

packages/adders/paraglide/index.ts
manuel3108: Another option would be to allow `sv.file.update()` to group it even more. But in that case i would probably start destructuring the object, basically leaving us with an option we already have.

packages/adders/paraglide/index.ts
```diff
@@ -61,24 +45,35 @@ export const options = defineAdderOptions({
...
-		{
-			// create an inlang project if it doesn't exist yet
-			name: () => 'project.inlang/settings.json',
-			condition: ({ cwd }) => !fs.existsSync(path.join(cwd, 'project.inlang/settings.json')),
-			content: ({ options, content }) => {
+	run: ({ sv, cwd, options, typescript, kit, dependencyVersion }) => {
+		const ext = typescript ? 'ts' : 'js';
+		if (!kit) throw new Error('SvelteKit is required');
+
+		sv.dependency('@inlang/paraglide-sveltekit', '^0.11.1');
```
benmccann, line 73: perhaps it should be called `addDependency`?

packages/adders/paraglide/index.ts
```diff
@@ -61,24 +45,35 @@ export const options = defineAdderOptions({
...
+
+		if (!fs.existsSync(path.join(cwd, 'project.inlang/settings.json'))) {
+			sv.updateFile('project.inlang/settings.json', (content) => {
```
benmccann, line 76: I think we could combine these into a single API that adds the file if it doesn't exist. (maybe `addFile`, `updateFile`, and `upsertFile`? or `updateFile` could take an option describing whether to throw an error or overwrite if an existing file is found?)
```suggestion
		if (!fs.existsSync(path.join(cwd, 'project.inlang/settings.json'))) {
			sv.updateFile('project.inlang/settings.json', (content) => {
```

packages/adders/paraglide/index.ts
benmccann: `pkg` or `pkgJson` or something would be another option

I haven't made up my mind about splitting vs combining, but do feel we should be consistent. maybe we should ask Rich for feedback after svelte.dev is updated as he's quite good with API design

packages/adders/paraglide/index.ts
AdrianGonz97: i think a singular API makes sense (similar to how the `name: () => '...'` field operates today where it'll only create the file if the returned string is _not empty_). it could also have a second parameter that indicates the status:
```ts
sv.file('project.inlang/settings.json', (content: string, kind: 'create' | 'update') => {
	if (kind === 'create') {
		// ...
	}
	if (kind === 'update') {
		// ...
	}
});
```
most of the time though, you wouldn't care about whether the file is created or updated, so it can simply be this:
```ts
sv.file('project.inlang/settings.json', (content: string) => {
	// ...
});
```

packages/adders/paraglide/index.ts
AdrianGonz97: thinking on it more, I think a single api via `sv` should be fine:
```ts
sv.file // creates or updates files - see other comment down below
sv.dependency // adds dep
sv.devDependency // adds devDep
sv.execute // runs any scripts
sv.dependsOn // runs another adder(?)
```

manuel3108: Generally, I agree.

The problem I have is with the `dependsOn` as described somewhere else.
We need to know the adders something depends on before running the `run` method, so that we can ask the appropriate questions

After some rapid prototyping I came up with something like this:
```js
// we need to collect all todos to detect the `dependsOn` dependencies
// so that we can ask the relevant questions beforehand
const todos: Array<{
	type: 'file' | 'dependency' | 'script' | 'dependsOn';
	data: SvApi['file'] | SvApi['execute'] | SvApi['dependsOn'] | Dependency;
}> = [];
```
The problem with this is, that it's still pretty basic, and I would still need to store the adder information and the (potentially) workspace somewhere. After collecting them for all adders we could filter out the `type: 'dependsOn'` ones and ask the required questions. Then we could run the `todos` in the same order we collected them to preserver consistency.

There are two alternatives i see:
* The adder that is being applied (`lucia` in our case) is responsible for asking the relevant `drizzle` questions, based on some sort of condition
* We make the current `dependsOn` array a function, and remove it from the sv api, so that we can collect such stuff beforehand.

manuel3108:
FYI we have talked about nuking the `routify` adder in the past multiple times. Instead of rewriting it, that's what i have done right now.

manuel3108:
Even if we are not 100% sure what the new api is going to look like, as of the discussion above we are all fine with having a single method that handles most of the stuff. Even if decided to rename a few methods, or change the params in some way or another, i have already migrated all adders to the new api syntax. This should make all future approvements way simpler.

packages/adders/paraglide/index.ts
manuel3108: I have just moved `dependsOn` out of the `sv` api, because the problems mentioned above. For now I have decided to make `dependsOn` a function accepting a workspace, so that every add-on is responsible for deciding if it needs to run or not. This works great, and I think this is the best solution we can find at this point.

manuel3108:
Functionally this should now be equal to the current master branch. But we will need for sure, once we have implemented back our test suite.

packages/core/adder/config.ts
```diff
@@ -25,18 +25,21 @@ export type Scripts<Args extends OptionDefinition> = {
 	condition?: ConditionDefinition<Args>;
 };
 
+export type SvApi = {
```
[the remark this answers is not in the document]
manuel3108: All of the things that are in there require context of some sort. like the package manager, the cwd and the workspace. Everything else should be importet.

manuel3108:
Rebase this and implement `setup` hook as discussed in https://github.com/sveltejs/cli/pull/252. We might be able combine `dependsOn` and `setup`

manuel3108:
I have currently merged #275 in here, to be able to solve all those merge conflicts and code adaptions. Once #275 merging `main` here, should reduce the lines changed by roughly 50%

benmccann:
I think some of the API methods are a bit inconsistent at the moment. Maybe we should prefix all methods with `add` (or `set` or `update`, etc. as appropriate) and make sure they're all adding the code instead of returning it for consistency?

> common.addJsDocTypeComment(exportDeclaration, "import('tailwindcss').Config");

adds the code and has `add` in the name

> root = common.satisfiesExpression(rootExport, 'Config');

returns the code

> sv.dependency

adds the code, but doesn't have `add` in the name

> object.property(pluginsObject, 'tailwindcss', object.createEmpty());

adds the code, but doesn't have `add` in the name

> imports.addDefault

adds the code and has `add` in the name

AdrianGonz97:
yea, we definitely need to cleanup the naming of AST related functions, but i would much prefer the pattern of excluding the `add`:
```ts
common.jsDocTypeComment(exportDeclaration, "import('tailwindcss').Config");

common.satisfiesExpression(rootExport, 'Config');

sv.dependency()

object.property(pluginsObject, 'tailwindcss', object.createEmpty());

imports.default()
```
perhaps this should be tackled separately since it's unrelated to the core of this pr? especially as we have plans to overhaul how they work anyway

manuel3108:
Yeah, totally agree with what @AdrianGonz97, this is required before we open up, but we can handle that later.

manuel3108:
Update PR description to not close the referenced issue, so that we can focus on the other stuff later on.

packages/adders/storybook/index.ts
```diff
@@ -2,16 +2,9 @@ import { defineAdder } from '@sveltejs/cli-core';
...
-	scripts: [
-		{
-			description: 'applies storybook',
-			args: ['storybook@8.3.6', 'init', '--skip-install', '--no-dev'],
-			stdio: 'inherit'
-		}
-	],
-	files: []
+	run: async ({ sv }) => {
+		await sv.execute(['storybook@8.4', 'init', '--skip-install', '--no-dev'], 'inherit');
```
benmccann, line 8: it seems funny to halfway pin it like this. I'd remove the `@8.4` so that folks get the latest when 8.5 comes out

packages/core/adder/config.ts
```diff
@@ -25,18 +20,22 @@ export type Scripts<Args extends OptionDefinition> = {
...
+	file: (path: string, content: (content: string) => string) => string;
```
AdrianGonz97, line 24: perhaps this should return `void` instead of a `string`? not sure of a use-case where returning a `string` would help
```suggestion
	file: (path: string, edit: (content: string) => string) => void;
```

packages/cli/commands/add/preconditions.ts
```diff
@@ -38,13 +39,9 @@ export function getGlobalPreconditions(
...
 				run: () => {
-					const addersForInvalidEnvironment = adders.filter((a) => {
-						const supportedEnvironments = a.environments;
-						if (projectType === 'kit' && !supportedEnvironments.kit) return true;
-						if (projectType === 'svelte' && !supportedEnvironments.svelte) return true;
-
-						return false;
-					});
+					const addersForInvalidEnvironment = adders.filter(
+						(a) => !adderSetupResult[a.id].available
```
AdrianGonz97, line 43: it might be better if `available` was replaced with something like `adderSetupResult[a.id].unsupported` that holds an array of messages that we collect from all of the add-ons and then throw at the end:
```ts
const messages = adders.flatMap((a) => adderSetupResult[a.id].unsupported.map((msg) => `${a.id}: ${msg}`));

// ...

throw new UnsupportedError(messages.join('\n'));
```

Usage in the add-on would then be:
```ts
setup: ({ kit, unsupported, dependencyVersion }) => {
	if (!kit) unsupported("Requires SvelteKit");
	if (!dependencyVersion('eslint')) unsupported("Requires eslint");
	// ...
},
```

manuel3108: `unsupported` is wayyy better than `unavailable` 👍 

And adding a message sound like a perfect idea.
Working on it.

AdrianGonz97 (approved):
pretty satisfied with how this has came out! I think it's finally ready to be merged

note: I decided that renaming everything to `addon` should probably happen in a follow up PR so we don't further pollute this one

packages/adders/lucia/index.ts
```diff
@@ -35,321 +34,315 @@ const options = defineAdderOptions({
...
+			return code;
+		});
+
+		sv.file(`${kit?.libDirectory}/server/auth.${ext}`, (content) => {
+			const { ast, generateCode } = parseScript(content);
+
+			js.imports.addNamespace(ast, '$lib/server/db/schema', 'table');
+			js.imports.addNamed(ast, '$lib/server/db', { db: 'db' });
+			js.imports.addNamed(ast, '@oslojs/encoding', {
+				encodeBase64url: 'encodeBase64url',
```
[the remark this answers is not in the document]
manuel3108: @jerriclynsjohn Please see this discussion and PR where this was changed: https://github.com/sveltejs/cli/pull/254#discussion_r1822837180

PR #206, opened 2024-10-22
chore: color help screen

### LINKED ISSUE #195: adding some color
Are you guys opposed to adding some color?   If used well, a boost to the look/feel can also improve readability.  I'd be happy to take a stab if it makes sense.

The `astro` and `nuxt` cli's are nice examples (though I prefer the former):

Screenshot 2024-10-21 at 9 01 16 PM

manuel3108:
Closes #195

Adds some basic color support. Colors can now easily be changed. 
The problem is that this currently not supported by `commander`m but a PR is on it's way: https://github.com/tj/commander.js/pull/2251
In the meantime it was necessary to pull in a small amount of code from commander to make this work. We might need to debate if it's worth it to pull in that code or to wait.

From a readability pov, i do think this is a major improvement.

image

image

### DIFF
--- a/.changeset/tough-gorillas-switch.md
+++ b/.changeset/tough-gorillas-switch.md
@@ -1,0 +1,5 @@
+---
+'sv': patch
+---
+
+chore: color help screen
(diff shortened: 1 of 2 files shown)


manuel3108:
Does not seem to be as simple as expected and adds a lot of overhead. We will need for the mentioned PR to be implemented to make this work in a better way.

PR #212, opened 2024-10-23
Clarify CLI instructions

sacrosanctic:
related conversation
https://discord.com/channels/457912077277855764/1298479734606528628

### DIFF
--- a/packages/cli/commands/add/index.ts
+++ b/packages/cli/commands/add/index.ts
@@ -45,6 +45,8 @@ const communityDetails: AdderWithoutExplicitArgs[] = [];
 const defaultPkgPath = pkg.up();
 const defaultCwd = defaultPkgPath ? path.dirname(defaultPkgPath) : undefined;
 
+const multiselectHint = pc.dim('(use arrow keys / space bar)');
+
 export const add = new Command('add')
 	.description('applies specified integrations into a project')
 	.argument('[integration...]', 'integrations to install')
@@ -185,7 +187,7 @@ export async function runAddCommand(
 		}));
 
 		const selected = await p.multiselect({
-			message: 'Which community tools would you like to add to your project?',
+			message: `Which community tools would you like to add to your project? ${multiselectHint}`,
 			options: promptOptions,
 			required: false
 		});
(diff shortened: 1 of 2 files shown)


AdrianGonz97 (approved):
thanks!

packages/cli/commands/add/index.ts
```diff
@@ -284,7 +284,7 @@ export async function runAddCommand(
 			.filter((a) => !!a);
 
 		const selected = await p.multiselect({
-			message: 'What would you like to add to your project?',
+			message: `What would you like to add to your project? ${pc.dim('(use arrow keys/space bar)')}`,
```
benmccann, line 287:
```suggestion
			message: `What would you like to add to your project? ${pc.dim('(use arrow keys / space bar)')}`,
```

benmccann:
Do we need the same hint if you run `create`, etc ?

AdrianGonz97:
this change is already shared with `create` since as it invokes `runAddCommand`

benmccann (changes requested):
Oh, that's pretty weird. You'd want it to show up in the first create question - not the third

AdrianGonz97:
it only shows up on the multi-select where there's a difference between `space` and `enter`. the others are just normal select prompts

manuel3108:
I also understand it the way that the core problem is the multiselect. 
But I think we should be consistent and also add it to the tailwind plugin multiselect:

image

AdrianGonz97:
I thought about that and figured that it's fine to exclude it from proceeding multi-select prompts as you've been "taught" what to do by the time you reach this prompt

sacrosanctic:
What is the reason for not adding hints to subsequent multi-selects? Is it because of clutter?

Isn't it always better to provide visual aid, I understand it's squares for multi-select and circles for single-select. But I find that hardly noticeable.

benmccann:
I'd have to see what it looks like, but my first reaction is that it'd be too noisy and we should mention it only once

AdrianGonz97:
seeing it now, I still think it adds unnecessary noise. this noisiness is also exacerbated when the instructions are kept on screen long after prompt submission

sacrosanctic:
Is it possible to add a check to only show the hint on the first multi-select  the user encounters, or would that increase the code complexity.

If possible, this would reduce clutter and still have hints regardless of which branch the user takes.

manuel3108 (approved):

manuel3108:
Let's only add this to the integration-selection question for now, as everyone seems to agree there. Let's get this merged and hear for user feedback

Issue #82, opened 2024-10-10
Rename 'adders' to add-ons


Rich-Harris:
An adder is a type of snake. More generously, it's a thing that adds, not the thing being added.

'Plugins' is probably the right name, since it can cover both integrations with external libraries _and_ things that affect the behaviour of an app without relying on libraries (like, idk, adding some logging middleware), which may become more relevant if we add plugin hooks to SvelteKit in future

benmccann:
I _really_ don't like "extension". Everywhere I've heard that term, it's synonymous with plugin. These aren't extensions in the sense of browser extensions or VS Code extensions since there's no SvelteKit extension API that these are leveraging.

Back to the thesaurus game: `sweetners`, `additions`, `renovations`

rdytogokev:
Is `packages` too generic? 

This is, in fact, what would be added. It would account for any purpose, from a lonely 'robots.txt' file to a full-on middleware.

PR #226, opened 2024-10-25
chore: rename integrations to add-ons

benmccann:
matches the language used in the announcement blog post

closes https://github.com/sveltejs/cli/issues/82


manuel3108:
Is there a reason you decided to go with `add-ons` instead of `addons`? Probably from the blog post I imagine. Googeling around a bit it looks like both things are commonly used and i was unable to find any preference for one thing.

I noticed this while looking at the new args and `--no-add-ons` feels wired to me.

Also this PR "only" covers the public facing stuff. We should think about renaming our currently internal stuff to addons as well and potentially create a follow up. Especially in regard to the public facing API i think stuff like `defineAdderOptions` and `defineAdder` do not make much sense.

Rich-Harris:
Add-on is correct!

> `--no-add-ons` feels wired to me

We already use hyphenated flags elsewhere (`--compiler-warnings`, `--diagnostic-sources` etc) so I think it's fine to have another. (Though we do also have `--preserveWatchOutput` which should be updated to match the conventions everywhere else)

benmccann:
`--compiler-warnings`, etc. are replacing a space with a hyphen, which makes sense. I guess the question is if you start out with a hyphen, do you just remove it or leave it as a hyphen? I'm not opposed to `--no-addons`. It does look a bit nicer to me. But I agree that everywhere in text I would leave it as "Add-on", which is much more commonly used

manuel3108 (approved):

Rich-Harris:
great news, merging this caused https://github.com/sveltejs/svelte.dev/pull/729 to update. PRs will now automatically be created whenever docs changes are merged to main

PR #231, opened 2024-10-28
Tailwindcss default export should use `satisfies` instead of `as`

rcoopr:
The previous adder would create the following config file for tailwindcss:

```ts
import type { Config } from 'tailwindcss';

export default {
	content: ['./src/**/*.{html,js,svelte,ts}'],

	theme: {
		extend: {}
	}
} as Config;
```

With these changes, we can use the `satisfies` operator instead of `as`, and gain all the type hints which that brings.

There are probably other adders that will benefit from using `satisfies`

### DIFF
--- a/packages/ast-tooling/index.ts
+++ b/packages/ast-tooling/index.ts
@@ -16,7 +16,7 @@ import {
 import * as fleece from 'silver-fleece';
 import * as Walker from 'zimmerframe';
 import type { namedTypes as AstTypes } from 'ast-types';
-import type * as AstKinds from 'ast-types/gen/kinds.d.ts';
+import * as AstKinds from 'ast-types/gen/kinds';
 
 /**
  * Most of the AST tooling is pretty big in bundle size and bundling takes forever.
--- a/packages/ast-tooling/package.json
+++ b/packages/ast-tooling/package.json
@@ -1,38 +1,38 @@
 {
-	"name": "@sveltejs/ast-tooling",
-	"private": true,
-	"version": "0.0.0",
-	"type": "module",
-	"license": "MIT",
-	"repository": {
-		"type": "git",
-		"url": "https://github.com/sveltejs/cli/tree/main/packages/ast-tooling"
-	},
-	"bugs": "https://github.com/sveltejs/cli/issues",
-	"scripts": {
-		"check": "tsc",
-		"format": "pnpm lint --write",
-		"lint": "prettier --check . --config ../../prettier.config.js --ignore-path ../../.gitignore --ignore-path .gitignore --ignore-path ../../.prettierignore"
-	},
-	"files": [
-		"dist"
-	],
-	"exports": {
-		".": {
-			"types": "./dist/index.d.ts",
-			"default": "./dist/index.js"
-		}
-	},
-	"devDependencies": {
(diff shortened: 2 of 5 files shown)


packages/ast-tooling/index.ts
```diff
@@ -16,7 +16,7 @@ import {
 import * as fleece from 'silver-fleece';
 import * as Walker from 'zimmerframe';
 import type { namedTypes as AstTypes } from 'ast-types';
-import type * as AstKinds from 'ast-types/gen/kinds.d.ts';
+import * as AstKinds from 'ast-types/gen/kinds';
```
rcoopr, line 19: I had to update the import path here after updating the package, I'm not sure if this breaks anything but it seemed fine to me

packages/ast-tooling/index.ts
```diff
@@ -16,7 +16,7 @@ import {
 import * as fleece from 'silver-fleece';
 import * as Walker from 'zimmerframe';
 import type { namedTypes as AstTypes } from 'ast-types';
-import * as AstKinds from 'ast-types/gen/kinds';
+import type * as AstKinds from 'ast-types/gen/kinds';
```
rcoopr, line 19: I had to update the import path here after updating the package, I'm not sure if this breaks anything but it seemed fine to me

packages/ast-tooling/index.ts
rcoopr: I had to update the import path here after updating the package, I'm not sure if this breaks anything but it seemed fine to me (I fixed the type import in the next commit)

packages/ast-tooling/index.ts
```diff
@@ -16,7 +16,7 @@ import {
 import * as fleece from 'silver-fleece';
 import * as Walker from 'zimmerframe';
 import type { namedTypes as AstTypes } from 'ast-types';
-import type * as AstKinds from 'ast-types/gen/kinds.d.ts';
+import type * as AstKinds from 'ast-types/gen/kinds';
```
rcoopr, line 19: I had to update the import path here after updating the package, I'm not sure if this breaks anything but it seemed fine to me

packages/ast-tooling/package.json
```diff
@@ -1,38 +1,38 @@
...
-		"dom-serializer": "^2.0.0",
-		"domhandler": "^5.0.3",
-		"domutils": "^3.1.0",
-		"htmlparser2": "^9.1.0",
-		"postcss": "^8.4.38",
-		"recast": "^0.23.7",
-		"silver-fleece": "^1.2.1",
-		"zimmerframe": "^1.1.2"
-	}
+  "name": "@sveltejs/ast-tooling",
```
benmccann, line 2: can you revert the spacing changes in this file?

AdrianGonz97 (changes requested):
There was a specific purpose for using `as` over `satisfies` in this case, but I can't recall it at the moment. Going to prevent a merge for now so I can take a second to look back

packages/ast-tooling/package.json
benmccann: not sure since it seems specific to your setup and I don't know what editor you use, etc. we do have prettier setup in this repo, which I think should format things appropriately

benmccann (approved):

PR #234, opened 2024-10-28
Update docs

Rich-Harris:
Updates the docs, in preparation for adding them to the site. You can see how it looks here: https://svelte-dev-git-add-cli-docs-svelte.vercel.app/docs/cli/overview

I replaced the tables with normal text — it's less compact, but far more readable and doesn't mess up the layout on smaller screens. Also makes it possible to add code blocks etc.


manuel3108 (approved):
Makes sense, looks better. This is also how a lot of other docs are done ([`pnpm` sample](https://pnpm.io/cli/install#options))

PR #237, opened 2024-10-28
chore: add `postpublish` script and manual template updater

AdrianGonz97:


AdrianGonz97:
self-merging to test if the update script is working

PR #252, opened 2024-10-29
chore: remove routify

manuel3108:
As we decided to ditch #32 for an upcoming more flexible PR, we should remove routify separately. 
Reasoning: Routify is nearly unused on npm, and we do recommend creating full kit applications


AdrianGonz97 (approved):

benmccann:
we can remove the `svelte` environment now as this was the only one that wouldn't run in a kit project

manuel3108:
Do you mean completely removing the `svelte` property, or completely removing the `environment` property?

AdrianGonz97:
We can just have the `kit` field where it indicates that SvelteKit is _required_. if `kit === true` and we're _not_ in a SvelteKit project, then the precondition error is thrown

manuel3108:
Should this still be an object while we are going to be refactoring that? Can we think of any other `environments` that might appear in the future?

Should we rename `kit` to `requiresKit` to make things more clear?

AdrianGonz97:
Maybe we should just get rid of the field entirely and add a `setup` hook as part of #181? I don't think we should be blocking this PR from merging as we can address this separately.

benmccann:
Yeah, I was talking about the `svelte` environment, which has always felt weird to me because everything is a svelte project. Maybe we should just rename that one to `vite` instead?

> Can we think of any other environments that might appear in the future?

`vite` would probably be the main one. Like I bet the Tailwind adder isn't actually a `svelte` adder, but is a `vite` adder. Probably if you tried to set it up in a webpack project it doesn't work because you need postcss setup correctly

benmccann:
yeah, it doesn't have to block this PR

I'd thought about the setup hook too. The trade-off there is that a hook is more flexible, but can't be displayed easily via a website directory

manuel3108:
But `kit` project are also using `vite`.

`setup` hooks sounds fine, and aligns with our goal of a more flexible api. But let's implement this in #181

PR #254, opened 2024-10-30
fix: update lucia adder

### LINKED ISSUE #122: lucia: add `setSessionTokenCookie` and `deleteSessionTokenCookie` methods
It would help following the guide to have `setSessionTokenCookie` and `deleteSessionTokenCookie` methods:
https://github.com/lucia-auth/example-sveltekit-email-password-2fa/blob/main/src/lib/server/session.ts#L57

pilcrowonpaper:
- Adds `setSessionTokenCookie()` and `deleteSessionTokenCookie()` (closes #122)
- Fixes sessions to be consistent with lucia-auth.com. This isn't a security issue, but the hashing part wasn't implement properly which negated the benefits it provided.
- Update `generateUserId()` to be consistent with `generateSessionToken()` and easier to deal with.
- Removed return types to be consistent with the rest of the codebase.

I haven't added change sets yet and I'm not sure how to test the adder.

### DIFF
--- a/packages/adders/lucia/index.ts
+++ b/packages/adders/lucia/index.ts
@@ -230,9 +230,9 @@ export default defineAdder({
 				if (!ms.original.includes('export const sessionCookieName')) {
 					ms.append("\n\nexport const sessionCookieName = 'auth-session';");
 				}
-				if (!ms.original.includes('function generateSessionToken')) {
+				if (!ms.original.includes('export function generateSessionToken')) {
 					const generateSessionToken = dedent`					
-						function generateSessionToken() {
+						export function generateSessionToken() {
 							const bytes = crypto.getRandomValues(new Uint8Array(20));
 							const token = encodeBase32LowerCaseNoPadding(bytes);
 							return token;
--- a/packages/adders/lucia/index.ts
+++ b/packages/adders/lucia/index.ts
@@ -365,7 +365,6 @@ export default defineAdder({
 			content: ({ content, typescript }) => {
 				const { ast, generateCode } = parseScript(content);
 				imports.addNamespace(ast, '$lib/server/auth.js', 'auth');
-				imports.addNamed(ast, '$app/environment', { dev: 'dev' });
 				kit.addHooksHandle(ast, typescript, 'handleAuth', getAuthHandleContent());
 				return generateCode();
 			}
@@ -394,7 +393,6 @@ export default defineAdder({
 					import { encodeBase32LowerCaseNoPadding } from '@oslojs/encoding';
 					import { fail, redirect } from '@sveltejs/kit';
 					import { eq } from 'drizzle-orm';
-					import { dev } from '$app/environment';
 					import * as auth from '$lib/server/auth';
 					import { db } from '$lib/server/db';
 					import * as table from '$lib/server/db/schema';
--- a/packages/adders/lucia/index.ts
+++ b/packages/adders/lucia/index.ts
@@ -212,7 +212,7 @@ export default defineAdder({
 				imports.addNamespace(ast, '$lib/server/db/schema', 'table');
 				imports.addNamed(ast, '$lib/server/db', { db: 'db' });
 				imports.addNamed(ast, '@oslojs/encoding', {
-					encodeBase32LowerCaseNoPadding: 'encodeBase32LowerCaseNoPadding',
+					encodeBase64url: 'encodeBase64url',
 					encodeHexLowerCase: 'encodeHexLowerCase'
 				});
 				imports.addNamed(ast, '@oslojs/crypto/sha2', { sha256: 'sha256' });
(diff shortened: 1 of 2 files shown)


packages/adders/lucia/index.ts
```diff
@@ -217,6 +217,9 @@ export default defineAdder({
 				});
 				imports.addNamed(ast, '@oslojs/crypto/sha2', { sha256: 'sha256' });
 				imports.addNamed(ast, 'drizzle-orm', { eq: 'eq' });
+				if (typescript) {
```
pilcrowonpaper, line 220: I'm not sure if this is the correct way to add type imports

manuel3108:
Cool, thanks!

You can test the adder by running `pnpm sv add lucia` insider the `pnpm workspace`. 
Or use `pnpx https://pkg.pr.new/sveltejs/cli/sv@254 add lucia` which was published instants ago

packages/adders/lucia/index.ts
```diff
@@ -365,7 +392,7 @@ export default defineAdder({
 				const [ts] = utils.createPrinter(typescript);
 				return dedent`
 					import { hash, verify } from '@node-rs/argon2';
-					import { generateRandomString } from '@oslojs/crypto/random';
+					import { encodeBase32LowerCaseNoPadding } from '@oslojs/encoding';
```
benmccann, line 395: thanks! are there some docs where I can learn more about this method and others that are available? I didn't see it mentioned on https://oslo.js.org/reference/encoding/

pilcrowonpaper: Sorry, that's the "v1" docs. The "v2" successor is available at https://encoding.oslojs.dev and it's documented here

https://encoding.oslojs.dev/reference/main/encodeBase32LowerCaseNoPadding

pilcrowonpaper: Base32 is much more compact than hex encoding (used by UUID etc) but it's also case insensitive and doesn't use special characters unlike base64

packages/adders/lucia/index.ts
```diff
@@ -300,14 +294,47 @@ export default defineAdder({
...
+				}
+				if (!ms.original.includes('export function deleteSessionTokenCookie')) {
+					const deleteSessionTokenCookie = dedent`					
+						${ts('', '/** @param {import("@sveltejs/kit").RequestEvent} event */')}
+						export function deleteSessionTokenCookie(event${ts(': RequestEvent')}, token${ts(': string')}, expiresAt${ts(': Date')}) {
+							event.cookies.set(sessionCookieName, '', {
+								expires: expiresAt,
+								path: '/'
+							});
+						}`;
```
AdrianGonz97, line 335: any particular reason to use `event.cookies.set` over `event.cookies.delete`?
```suggestion
						export function deleteSessionTokenCookie(event${ts(': RequestEvent')}, expiresAt${ts(': Date')}) {
							event.cookies.delete(sessionCookieName, { path: '/' });
						}`;
```

packages/adders/lucia/index.ts
AdrianGonz97: this is good 👍

packages/adders/lucia/index.ts
pilcrowonpaper: Should be fixed now

packages/adders/lucia/index.ts
```diff
@@ -229,19 +232,20 @@ export default defineAdder({
 				}
 				if (!ms.original.includes('function generateSessionToken')) {
 					const generateSessionToken = dedent`					
-						${ts('', '/** @returns {string} */')}
-						function generateSessionToken()${ts(': string')} {
+						function generateSessionToken() {
```
AdrianGonz97:
```suggestion
						export function generateSessionToken() {
```

pilcrowonpaper:
The generated TypeScript code works. I can't seem to get `sv` to generate JSDoc code tho. Is this expected?

packages/adders/lucia/index.ts
pilcrowonpaper: Just fixed this as well

AdrianGonz97:
That's expected. The JSDoc in this instance is only included when `typescript` isn't present in the project.

pilcrowonpaper:
Tested out with the published package and all the types seems to be working in both the TS and JSDoc version

AdrianGonz97 (approved):
much appreciated!

manuel3108 (approved):
Any reason this is still a draft?

packages/adders/lucia/index.ts
benmccann: I would lean towards the base 64 version to maintain the previous behavior as it's a much more compact representation and more closely matches popular implementations like https://github.com/ai/nanoid

pilcrowonpaper: I personally prefer IDs that don't contain special characters and this provides about the same entropy as nanoid with a similar number of characters (24 vs 21). I don't mind changing it though. Do you want me to change `generateSessionToken()` to use base64 as well?

benmccann: Yeah, I guess it makes sense to remain consistent

I see the input was changed to 15 characters. How do we know how many characters the output will be?

pilcrowonpaper: The generated ID will actually be 24 characters. In base32, every 5 bits (that's where the 32 comes from) are converted to a corresponding character (e.g. 0b00001 => A). So you get 8 characters for every 5 bytes, or 24 characters for a 15 byte input.

For base64, it's 6 bits instead of 5, so you get 4 characters for every 3 bytes.

benmccann:
thank you!! this is a big help

we're hoping to add more things like oauth in the future. we're working on cleaning up some of our APIs to hopefully make it easier to add and maintain add-ons.

really appreciate you taking a look!

PR #255, opened 2024-10-30
fix docs

Rich-Harris:
The `--types` docs are out of date post-#249 (ideally we'd update the docs at the same time as we update the functionality, though it's easy to forget)

Also updates a link that's currently breaking https://github.com/sveltejs/svelte.dev/pull/725 — the same change as in #226


AdrianGonz97 (approved):
oops, thanks!!

PR #273, opened 2024-11-07
feat: `migrate` allow passing path or component

manuel3108:
Relates https://github.com/sveltejs/cli/issues/278

Obviously only merge once the issue above is implemented.

### DIFF
--- a/packages/cli/commands/migrate.ts
+++ b/packages/cli/commands/migrate.ts
@@ -7,6 +7,7 @@ import { getUserAgent } from '../common.ts';
 export const migrate = new Command('migrate')
 	.description('a CLI for migrating Svelte(Kit) codebases')
 	.argument('<migration>', 'migration to run')
+	.argument('[path]', 'component path or directory to migrate')
 	.option('-C, --cwd <path>', 'path to working directory', process.cwd())
 	.configureHelp({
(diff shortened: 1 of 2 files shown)


paoloricciuti:
> Obviously only merge once the issue above is implemented.

You can't tell me what to do [merge]

😄 😄 😄

packages/cli/commands/migrate.ts
```diff
@@ -7,6 +7,7 @@ import { getUserAgent } from '../common.ts';
 export const migrate = new Command('migrate')
 	.description('a CLI for migrating Svelte(Kit) codebases')
 	.argument('<migration>', 'migration to run')
+	.argument('[path]', 'component path or directory to migrate')
```
benmccann, line 10: This seems confusing to me because I doubt that all the migrations will accept a path. I think the args should be dependent on the migration. For now I think I'd just pass the args through without trying to list them here

maybe after https://github.com/sveltejs/cli/pull/274 is merged we can use the same CLI parsing library for both and share the args across packages or something

manuel3108: That's indeed a valid argument. Now that the migrate code is here, we can think about this directly while implementing #278 

Closing here.

PR #275, opened 2024-11-07
feat: simple testing library

AdrianGonz97:
Alternative to #32 

This takes more of a lightweight approach where we provide a bare minimum api for testing rather than controlling the whole testing pipeline. This should allow for better flexibility as we're no longer tied to any particular testing framework.

Notes:
- Had to rename the directory of the vitest addon as it was _somehow_ interfering with vitest or tinypool during runtime, causing shit to randomly break
- Storybook has been pinned to `v8.3.x` as `v8.4.x` flat out doesn't work. I haven't had a moment to fully investigate the issue yet.

TODO:
- Snapshots?
- look into using vitest workspaces
- Should we continue to use `degit` to clone the vite templates? If so, should we pin it to a specific commit hash ([see here](https://github.com/sveltejs/cli/blob/6a139e257605a6cc8ec68e5bfcde978c122f74d0/packages/cli/testing.ts#L49-L51))? Or should this be something we manage fully (perhaps as part of `create`)?


package.json
```diff
@@ -39,8 +43,5 @@
 		"unplugin-isolated-decl": "^0.6.5",
 		"vitest": "^2.0.5"
```
benmccann, line 44: https://github.com/sveltejs/cli/security/dependabot shows some issues from vite. Perhaps we could upgrade it to address those. I didn't want to do it while you were working on testing stuff, but if it doesn't break things, you might want to try doing it here
```suggestion
		"vitest": "^2.1.4"
```

.gitignore
```diff
@@ -7,3 +7,6 @@ yarn.lock
 vite.config.js.timestamp-*
 /packages/create-svelte/template/CHANGELOG.md
 .test-tmp
+.test-output
+
+*storybook.log
```
benmccann, line 12: nit: it'd be nice to add a newline to the end of this file and the `.gitignore` in the community template

AdrianGonz97: huh, looks like storybook added that. should be fixed now 👍

packages/cli/lib/install.ts
```diff
@@ -0,0 +1,92 @@
...
+		workspace.options = options[addon.id];
+
+		const files = await runAddon(workspace, addon);
+		files.forEach((f) => filesToFormat.add(f));
+	}
+
+	return Array.from(filesToFormat);
+}
+
+async function runAddon(
```
manuel3108, line 43: nit: not sure these fits in a file called `install.ts`. Also it looks like this file is the only one in `lib` and has only to do with adders. What about merging this with `commands/add/utils.ts` or `utils/adders.ts`?

manuel3108 (changes requested):
Overall this looks quite good, more on that later. I added a comment for every small thing i noticed, so this might look a bit overwhelming at first. Below is a list of everything that i was not able to directly tie to the code.

---

On windows after the initial test run im getting this error (on the second test run so to say):
```
Error: EPERM: operation not permitted, unlink 'D:\dev\web\svelte-cli\packages\adders\.test-output\_tests\drizzle\node_modules\.pnpm\@esbuild+win32-x64@0.21.5\node_modules\@esbuild\win32-x64\esbuild.exe'
```
Avoiding to delete the `node_modules` folder and letting pnpm figure out the rest worked flawlessly for me: 

https://github.com/sveltejs/cli/blob/ff96d27ae52cea8113b24f845d27defa7923f25e/packages/adder-testing-library/utils.ts#L168-L171

--- 

`community-adder-template/tests/setup/suite.ts` This looks like a lot of code is duplicated from our internal global adder setup. I think we should provide a common abstraction over this, while allowing such detailed control if required. I don't think community adders will require such detailed control in 95% of the cases

---

Why not place the adder tests inside the pre-defined adder directory and instead nest them away in the `_tests/adder/test.ts` file?

---

I find that the export `sv/test` is weird, as its only providing testing stuff for addons. Maybe something like `sv/add-on/test` would be a better fit. Initially i was irritated because i did not find `test.ts` and need to look at the export map to see that it's actually referencing `testing.ts`. This might only be a me problem.

---

`storybook` ~~and `drizzle`~~ tests are timing out for me locally, will investigate later. I even accidentally increased the timeout to 85 minutes but cut it of after about 20 minutes where it still hadn't finished.

---

> Snapshots?

I do think they would be particularly useful for stuff like #100 or #181, but we can always add that later.

> Should we continue to use degit to clone the vite templates? If so, should we pin it to a specific commit hash ([see here](https://github.com/sveltejs/cli/blob/6a139e257605a6cc8ec68e5bfcde978c122f74d0/packages/cli/testing.ts#L49-L51))? Or should this be something we manage fully (perhaps as part of create)?

I would personally be in favor of controlling this stuff ourselves, see #170.

Overall, I don't have really strong preferences between #275 and #32 both have different pros and cons.

manuel3108:
Ohhh and there is another thing I just noticed: How are the directory names inside `.test-output/_tests` generated? Here are my directories inside `drizzle`
```
-1930824653_0
-1930824653_1
-1930824653_2
-1930824653_3
-1930824653_4
-1930824653_5
-1930824653_6
-1930824653_7
```
They are semi-useful at least. In #32 I generated them based on the options passed for each test, so that if a test failed you could identify it by directory name and run it by hand.

community-adder-template/tests/custom-addon.test.ts
```diff
@@ -0,0 +1,28 @@
...
+	const cwd = await ctx.run(variant, { [id]: { demo: true } });
+
+	// ...add files
+	if (variant.startsWith('kit')) {
+		const target = path.resolve(cwd, 'src', 'routes', '+page.svelte');
+		fixture({ name: '+page.svelte', target });
+	} else {
+		const target = path.resolve(cwd, 'src', 'App.svelte');
+		fixture({ name: 'App.svelte', target });
+	}
```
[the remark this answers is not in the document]
AdrianGonz97: The import specifiers for the demo are different. Though, we could simplify it to just modify the contents of the file like we do in the mdsvex test

packages/adders/_config/official.ts
```diff
@@ -9,7 +9,7 @@ import playwright from '../playwright/index.ts';
 import prettier from '../prettier/index.ts';
 import storybook from '../storybook/index.ts';
 import tailwindcss from '../tailwindcss/index.ts';
-import vitest from '../vitest/index.ts';
+import vitest from '../vitest-addon/index.ts';
```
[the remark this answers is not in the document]
AdrianGonz97: we should, yea. just haven't gotten around to making a repro for it yet

community-adder-template/tests/setup/global.ts
```diff
@@ -0,0 +1,29 @@
+import path from 'node:path';
+import { setup, type ProjectVariant } from 'sv/test';
+import type { GlobalSetupContext } from 'vitest/node';
+
+const variants: ProjectVariant[] = ['kit-js', 'kit-ts', 'vite-js', 'vite-ts'];
```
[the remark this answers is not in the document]
AdrianGonz97: honestly, I think `vite-` and `kit-` are perfectly fine. i don't think anyone is going to be confused by this distinction in practice

community-adder-template/.gitignore
```diff
@@ -1,3 +1,4 @@
 node_modules
 temp
 .outputs
+.test-output
```
benmccann, line 4: nit: it'd be nice to add a newline to the end of this file as well

AdrianGonz97:
> How are the directory names inside `.test-output/_tests` generated? Here are my directories inside `drizzle`

they're based on the task id provided by vitest. Each test gets their own `meta.json` (similar to how we had it in `svelte-add`) that documents the options that were passed in.

> They are semi-useful at least. In #32 I generated them based on the options passed for each test, so that if a test failed you could identify it by directory name and run it by hand.

not sure if that's a scalable solution. imagine having multiple test cases that share the same addon options but they evaluate different things

packages/cli/testing.ts
```diff
@@ -99,8 +106,29 @@ export async function startPreview({ cwd, command = 'npm run preview' }: Preview
...
+async function terminate(pid: number) {
+	const children = await getProcessTree(pid);
+	// the process tree is ordered from parents -> children,
+	// so we'll iterate in the reverse order to terminate the children first
+	for (let i = children.length - 1; i >= 0; i--) {
+		const child = children[i];
+		const pid = Number(child.PID);
+		process.kill(pid);
+	}
+	process.kill(pid);
```
manuel3108, line 133: Interesting, because this looks exactly like the selling point of [`terminate`](https://www.npmjs.com/package/terminate)

> If you then want to restart your app you naively (we did this!) terminate the "main" process and expect all the "child" processes to be ended too.

> 
> Sadly this results in [Zombie Processes](https://stackoverflow.com/questions/27381163) which you would have to either manually kill or restart the machine/VM to clear.
Instead you need to find all the Process IDs for the child processes that your app created and terminate those before you can re-start your app.

> 
> Using terminate you no longer have this problem.

But whatever works is great!

packages/adders/vitest.config.ts
```diff
@@ -7,10 +6,9 @@ const ONE_MINUTE = 1000 * 60;
 export default defineConfig({
 	test: {
 		include: ['_tests/**/test.{js,ts}'],
-		testTimeout: ONE_MINUTE * 2,
+		testTimeout: ONE_MINUTE * 3,
 		hookTimeout: ONE_MINUTE * 3,
-		maxConcurrency: cpus().length,
-		pool: 'forks',
+		maxConcurrency: 10,
```
manuel3108: Is there a reason for the `maxConcurrency` simplification? Having only 6 cores and 12 threads this overloads my computer and results in many timeouts. Reducing it down to 6 solved most occurences of these issues

packages/cli/lib/install.ts
manuel3108: this is no longer true in my opionion

packages/adders/vitest.config.ts
```diff
@@ -7,10 +7,10 @@ export default defineConfig({
...
 		name: 'adders',
 		include: ['_tests/**/test.{js,ts}'],
+		globalSetup: ['_tests/_setup/global.ts'],
 		testTimeout: ONE_MINUTE * 3,
 		hookTimeout: ONE_MINUTE * 3,
-		maxConcurrency: 10,
-		globalSetup: ['_tests/_setup/global.ts'],
-		retry: env.CI ? 3 : 0
+		retry: env.CI ? 3 : 0,
+		pool: 'threads'
```
manuel3108, line 14: I know, you are going to love this. This causes problems on windows :D.

Running the tests this line produces the following error for basically all the tests:
```
TypeError: process.chdir() is not supported in workers
 - /node_modules/.pnpm/playwright-core@1.48.2/node_modules/playwright-core/lib/utilsBundleImpl/index.js:43:461
 - /node_modules/.pnpm/playwright@1.48.2/node_modules/playwright/lib/common/expectBundleImpl.js:45:6553
 - /node_modules/.pnpm/degit@2.8.4/node_modules/degit/dist/index-688c5d50.js:7940:9
 - /node_modules/.pnpm/tinyexec@0.3.1/node_modules/tinyexec/dist/main.js:207:20
 - /node_modules/.pnpm/tinyexec@0.3.1/node_modules/tinyexec/dist/main.js:212:12
 - /node_modules/.pnpm/tinyexec@0.3.1/node_modules/tinyexec/dist/main.js:270:14
 - /node_modules/.pnpm/tinyexec@0.3.1/node_modules/tinyexec/dist/main.js:277:13
 - /node_modules/.pnpm/tinyexec@0.3.1/node_modules/tinyexec/dist/main.js:298:26
 - /node_modules/.pnpm/tinyexec@0.3.1/node_modules/tinyexec/dist/main.js:544:83
```

So we either have to remove all those tools from the testing happy path, or just remove that line

AdrianGonz97: how lovely. will fix momentarily

manuel3108 (approved):

Issue #283, opened 2024-11-08
The link to `svelte-migrate` in the docs is broken


fnimick:
Link on: https://svelte.dev/docs/cli/sv-migrate

target: https://github.com/sveltejs/kit/blob/main/packages/migrate (404)

PR #284, opened 2024-11-08
docs: fix broken link

benmccann:
closes https://github.com/sveltejs/cli/issues/283

### DIFF
--- a/documentation/docs/20-commands/40-sv-migrate.md
+++ b/documentation/docs/20-commands/40-sv-migrate.md
@@ -2,7 +2,7 @@
 title: sv migrate
 ---
 
-`sv migrate` migrates Svelte(Kit) codebases. It delegates to the [`svelte-migrate`](npmjs.com/package/svelte-migrate) package.
+`sv migrate` migrates Svelte(Kit) codebases. It delegates to the [`svelte-migrate`](https://www.npmjs.com/package/svelte-migrate) package.
 
 Some migrations may annotate your codebase with tasks for completion that you can find by searching for `@migration`.
 
--- a/documentation/docs/20-commands/40-sv-migrate.md
+++ b/documentation/docs/20-commands/40-sv-migrate.md
@@ -2,7 +2,7 @@
 title: sv migrate
 ---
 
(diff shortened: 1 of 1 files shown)


Rich-Harris:
preview: https://svelte-dev-git-preview-cli-284-svelte.vercel.app/

documentation/docs/20-commands/40-sv-migrate.md
```diff
@@ -2,7 +2,7 @@
 title: sv migrate
 ---
 
-`sv migrate` migrates Svelte(Kit) codebases. It delegates to the [`svelte-migrate`](https://github.com/sveltejs/kit/blob/main/packages/migrate) package.
+`sv migrate` migrates Svelte(Kit) codebases. It delegates to the [`svelte-migrate`](npmjs.com/package/svelte-migrate) package.
```
Conduitry, line 5:
```suggestion
`sv migrate` migrates Svelte(Kit) codebases. It delegates to the [`svelte-migrate`](https://www.npmjs.com/package/svelte-migrate) package.
```

But also: Would it be more helpful to link to the location of the package in this repo, rather than to npm?

benmccann: Thanks for catching this!

> But also: Would it be more helpful to link to the location of the package in this repo, rather than to npm?

I don't think so, which is why I changed it. Is there a reason you think that might be more helpful?

manuel3108: Is it even worth to mention that in the docs? It's more of an internal thing for me at least. Also does not look like we have a similar sentence for `check` in the docs.

benmccann: There are lots of articles, blog posts, tweets, etc. mentioning the two, so I think it's helpful for users to know what the difference is between them and how they relate

benmccann: Hmm. But I guess all the docs have been updated to refer to sv, so maybe the more important place to put this is the svelte-migrate readme

manuel3108 (approved):

PR #290, opened 2024-11-11
feat: add `jsconfig.json` to the 'no type checking' template

### LINKED ISSUE #267: Add jsconfig.json even for people not wanting any type checking
Right now the `jsconfig.json` is omitted if people want no type checking at all. This has a few unwanted side effects
- intellisense isn't as good anymore. The VS Code extension will make the whole thing a inferred project which has some limitations
- stuff like aliases (`$lib` for example) doesn't work, in the sense of if you click on the component / import path it won't take you to the definition, because intellisense has no way of knowing how to resolve it
- other quirks like https://github.com/sveltejs/language-tools/issues/2557

For this reason, the CLI should add a jsconfig regardless, just with the type checking option turned off (i.e. `"checkJs": false`)

AdrianGonz97:
closes #267

### DIFF
--- a/packages/create/index.ts
+++ b/packages/create/index.ts
@@ -98,7 +98,7 @@ function matches_condition(condition: Condition, options: Options) {
 	if (condition === 'demo' || condition === 'minimal' || condition === 'library') {
 		return options.template === condition;
 	}
-	if (condition === 'typescript' || condition === 'checkjs') {
+	if (condition === 'typescript' || condition === 'checkjs' || condition === 'none') {
 		return options.types === condition;
 	}
 	return !!options[condition];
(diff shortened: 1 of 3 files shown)


packages/create/index.ts
```diff
@@ -98,7 +98,7 @@ function matches_condition(condition: Condition, options: Options) {
 	if (condition === 'demo' || condition === 'minimal' || condition === 'library') {
 		return options.template === condition;
 	}
-	if (condition === 'typescript' || condition === 'checkjs') {
+	if (condition === 'typescript' || condition === 'checkjs' || condition === 'none') {
 		return options.types === condition;
 	}
 	return !!options[condition];
```
benmccann, line 104: would this be equivalent? I think all possible values are in this `if` statement
```suggestion
	return options.types === condition;
```

AdrianGonz97: not exactly since it's seemingly possible for the condition to exist as a key of `options`. i may be able to simplify it a tiny bit though

manuel3108 (approved):

PR #292, opened 2024-11-11
fix: ignore path prompt if user provided path in `create`

### LINKED ISSUE #291: sv create ignore when the path given is the current directory
image
I'm looking to skip/automate all prompts. In the curent situation it's impossible to skip the path prompt when targeting the current directory

manuel3108:
Closes #291

### DIFF
--- a/.changeset/popular-cows-swim.md
+++ b/.changeset/popular-cows-swim.md
@@ -1,0 +1,5 @@
+---
+'sv': patch
+---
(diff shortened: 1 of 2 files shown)


benmccann:
I liked the name `userProvidedPath` better, so might revert the last commit :laughing:

AdrianGonz97 (approved):

Issue #296, opened 2024-11-12
Unpin `storybook@8.4.0`


manuel3108:
In #275 we found out that the latest `storybook@8.4.2` appear to be broken. Same goes for `storybook@8.3.7`. Once they are working again, we should unpin this.

benmccann:
I can't reproduce any issue running the following commands:
```
npx sv create myapp
cd myapp
npx storybook init --skip-install --no-dev
npm install
npm run storybook
```

It installs and runs storybook 8.4.2 just fine for me

AdrianGonz97:
Here's the stack trace:
```shell
➜  npx storybook@latest init --skip-install --no-dev
/home/koala/.npm/_npx/6b1ed7b66a76da7f/node_modules/create-storybook/bin/index.cjs:23
  throw error;
  ^

Error: Cannot find module '@storybook/core/common'
Require stack:
- /home/koala/.npm/_npx/6b1ed7b66a76da7f/node_modules/storybook/core/common/index.cjs
- /home/koala/.npm/_npx/6b1ed7b66a76da7f/node_modules/create-storybook/dist/bin/index.cjs
- /home/koala/.npm/_npx/6b1ed7b66a76da7f/node_modules/create-storybook/bin/index.cjs
    at Module._resolveFilename (node:internal/modules/cjs/loader:1225:15)
    at Module._load (node:internal/modules/cjs/loader:1051:27)
    at Module.require (node:internal/modules/cjs/loader:1311:19)
    at require (node:internal/modules/helpers:179:18)
    at Object.<anonymous> (/home/koala/.npm/_npx/6b1ed7b66a76da7f/node_modules/storybook/core/common/index.cjs:1:18)
    at Module._compile (node:internal/modules/cjs/loader:1469:14)
    at Module._extensions..js (node:internal/modules/cjs/loader:1548:10)
    at Module.load (node:internal/modules/cjs/loader:1288:32)
    at Module._load (node:internal/modules/cjs/loader:1104:12)
    at Module.require (node:internal/modules/cjs/loader:1311:19) {
  code: 'MODULE_NOT_FOUND',
  requireStack: [
    '/home/koala/.npm/_npx/6b1ed7b66a76da7f/node_modules/storybook/core/common/index.cjs',
    '/home/koala/.npm/_npx/6b1ed7b66a76da7f/node_modules/create-storybook/dist/bin/index.cjs',
[...]

Issue #293, opened 2024-11-11
fix `storybook` tests on windows in ci


manuel3108:
In #275 we had severe problems with the storybook tests on windows. They were failing with obscure error message while running `pnpm dlx storybook@8.3.6`. Those failing tests mostly occurred on windows, but from time to time also on ubuntu. Never on mac.

Tests work flawlessly locally on windows; thus we decided to skip them ci for now

JReinhold:
Let me know if I can assist in any way here.

AdrianGonz97:
It's tough to pinpoint the exact issue, but some error info we've been able to get: https://github.com/sveltejs/cli/actions/runs/11806879168/job/32892430989?pr=181#step:8:168

AdrianGonz97:
@JReinhold 

I managed to reduce it down into a minimal repro: https://github.com/AdrianGonz97/storybook-ci-repro

It appears that it doesn't like to run separate processes of `storybook init` concurrently.

manuel3108:
> I managed to reduce it down into a minimal repro: https://github.com/AdrianGonz97/storybook-ci-repro
> 
> It appears that it doesn't like to run separate processes of `storybook init` concurrently.

Great. Was also able to reproduce this in a linux based codespace just now!

AdrianGonz97:
> Should we try running the storybook test non concurently?

Already covered in https://github.com/sveltejs/cli/pull/181/files#diff-88b82b064ede32a2332fd0d35f0f01326479ad0ff68248247c513c40838072caR10-R13

manuel3108:
Ohh great Since this is also happening on ubuntu from time to time, should we just remove the concurrent tests for `storybook` altogether?

AdrianGonz97:
I think we can just leave it alone for now. If it starts to get annoying again on the other platforms, then we can start considering to only running it sequentially in CI

manuel3108:
And for the sake of protocolling it somewhere, this just happend to me for the first time on `macOs` as well. https://github.com/sveltejs/cli/actions/runs/11843814850/job/33005611663?pr=303

JReinhold:
Are you okay with fixing it in your CI, not running them concurrently?

Because it doesn't sound like anything a regular user would do (you're probably the first ever to discover this), so sinking time into figuring out the issue on the Storybook side is a hard sell.

AdrianGonz97:
> Because it doesn't sound like anything a regular user would do (you're probably the first ever to discover this)

[...]

PR #298, opened 2024-11-13
chore: update `storybook` to v8.4.0

manuel3108:
Relates #296


AdrianGonz97:
`storybook` has already been set to `@latest` as part of #181, so this PR isn't necessary unless we're wanting to pin it to a specific version again

JReinhold:
If CI has been working with latest, maybe it was a temporary issue?

AdrianGonz97:
#296 was strictly a local issue and was unrelated to CI. In fact, I don't think we ever saw a CI failure as a result of using `@latest`. 

However, #293 _is_ a CI related issue with storybook where we noticed that we can't run our storybook tests concurrently on windows in CI (though, it works fine locally).

manuel3108:
> `storybook` has already been set to `@latest` as part of #181, so this PR isn't necessary unless we're wanting to pin it to a specific version again

Uhh, somehow didn't notice that!

Issue #297, opened 2024-11-12
better add command


cognominal:
Setting up stuff was a nightmare for people not familiar with the web ecosystem. sv/cli doing it is great. Thx.
The implementation of my proposition would help the beginners without bothering the experts:

* For each package to add, provide a oneliner description
* indicate the packages already installed by a blue checkbox

manuel3108:
> * For each package to add, provide a oneliner description

We have done so many revision of that, that i stopped counting. The current implementation is both minimal and gives you a direct link to the site, in case you don't know what this or that tool is doing.

One of our previous iterations was the one below, that you can try yourself with `pnpx svelte-add`. Please don't run it further than the initial questions tho, as the tool is outdated.
![image](https://github.com/user-attachments/assets/70e50874-d1a0-47b2-a5a1-04079c75778c)

The problem with this is that it feels clunky and requires much vertical space. Adding a description to the line might work, but why not link to their website directly, where they can describe themself.

> * indicate the packages already installed by a blue checkbox

This might be an interesting one. For some tools, this is pretty easy to detect based on the installed packages, but if I do remember correctly from the past, we had a problem with at least one tool, where this was not straightforward. Thus we decided to leave it.

Issue #247, opened 2024-10-28
Rename lucia adder to auth


benmccann:
Folks are getting really confused and thinking it's using a deprecated library

Rich-Harris:
It does feel a bit 'one true auth solution' if we do that, and I'm not sure that it is — I can't log in with google or github or biometrics or a passkey or whatever, there's no bot detection or email verification or password resets or the many other things that go into authentication, and we don't have anything to say about authorization. And it's tied to one specific ORM.

I'm not saying `lucia` is a name that we should keep, but I'm a little wary of people's expectations around `auth`. I can imagine Laravel developers saying 'aww, that's cute'.

manuel3108:
Maybe `basic-auth` or something

benmccann:
I think we can add those things over time. We need to start somewhere

`basic-auth` feels weird to me because then as we add features, at what point do we remove `basic-` from the name?

Conduitry:
'Basic auth' very much has an existing precise meaning already, which I really don't think we should try to go against.

AdrianGonz97:
How about `lucia-guide` or `lucia-v4`? We could also add `lucia` as an alias so that people can still write `npx sv add lucia`

benmccann:
yeah, someone had suggested `lucia-guide` on discord - though someone else then responded that it sounds like you're setting up documentation or something along those lines

I kind of want to send a PR for a feature that's not in the Lucia guide just so we can stop calling it anything related to Lucia :laughing:

AdrianGonz97:
`lucia-inspired` then?

gives us leeway to change things in the future while also referencing where it loosely came from

benmccann:
sounds fine to me

Rich-Harris:
`email-password-auth`? It's not like Lucia is doing anything particularly unusual

Rich-Harris:
(or maybe not that exact thing but you get the idea — something that says what it does)

AdrianGonz97:
i think at that point we should just call it `auth` so that it prompts a multiselect for the different auth methods like: `email-password`, `oauth`, `passkey`, etc

```
┌  Welcome to the Svelte CLI! (v0.5.11)
│
◇  What would you like to add to your project?
│  auth
│
◆  Which auth methods would you like to include?
│  ◻ email-password
│  ◻ oauth
│  ◻ passkey
│  ...
└

AdrianGonz97:
[...]

Issue #217, opened 2024-10-24
Remove lucia


zlstringham:
The author of lucia is deprecating it. https://github.com/lucia-auth/lucia/discussions/1714

Given that, I think it's a disservice to devs to bootstrap their applications with it today. I'd suggest just removing it. Alternatively, a "lucia-styled" auth could be scaffolded following https://lucia-auth.com/sessions/migrate-lucia-v3

AdrianGonz97:
thanks, but I welcome you to run the integration and see what it does

Issue #189, opened 2024-10-21
`lucia` is deprecated and should not be recommended by `sv create`


phenomen:
https://github.com/lucia-auth/lucia/discussions/1714

Since @pilcrowonpaper deprecated `lucia` package, it should not be a recommended way of adding auth to the new project.

benmccann:
It doesn't use the Lucia package. It uses the new guide

PR #299, opened 2024-11-13
feat: short description for each add-on

benmccann:
closes https://github.com/sveltejs/cli/issues/297

I've seen this feedback a number of times now, so think it's worth doing something here if we can keep the design clean enough

Screenshot from 2024-11-13 09-39-37


AdrianGonz97:
I don't know, #297 is the first time that I've seen a complaint relating to this. I'm also having a hard time finding any similar complaints when I search through the discord and reddit. Not sure if this is a real issue that people actually face

packages/core/adder/config.ts
```diff
@@ -29,6 +29,7 @@ export type Adder<Args extends OptionDefinition> = {
 	id: string;
 	alias?: string;
 	environments: Environments;
+	oneOrTwoWordDescription?: string;
```
manuel3108, line 32:
```suggestion
	shortDescription?: string;
```

manuel3108:
Agree with @AdrianGonz97, haven't seen any feedback beside the one mentioned earlier

packages/core/adder/config.ts
benmccann: I had thought about calling is `shortDescription`, but I thought it was better to be extra explicit about the expected length for folks building community adders

benmccann:
I don't know if I'm going to be able to go back and find them, but there is https://github.com/sveltejs/cli/issues/253: 

> For some addons it's pretty obvious what they do, but for others a link or a description would be very helpful. For example, unless you already know what mdsvex is, you're unlikely to know that that's what you want to be able to use Markdown in your Svelte components.

That issue was referring to our docs, but it's just as true that we don't have a description in the CLI

And of course there have been lots of folks confused about the Lucia adder (e.g. https://github.com/sveltejs/cli/issues/217,  https://github.com/sveltejs/cli/issues/189, [on reddit](https://www.reddit.com/r/sveltejs/comments/1gai9vk/sv_the_svelte_cli/?rdt=63894), etc.). This would allow us to close https://github.com/sveltejs/cli/issues/247 as well, by clarifying that it's the guide in the hint.

packages/core/adder/config.ts
manuel3108: I don't think we need to be this explicit, as we are always going to review at least the official ones. If we are going to have something similar for community adders is another discussion. 

Another alternative could be `keyword`

manuel3108:
Already forgot about the the `lucia` stuff, convinced.

packages/core/adder/config.ts
AdrianGonz97: i prefer `shortDescription` as well. we have full control of this so we can certainly limit ourselves

packages/adders/drizzle/index.ts
```diff
@@ -66,6 +66,7 @@ const options = defineAdderOptions({
 export default defineAdder({
 	id: 'drizzle',
 	environments: { svelte: false, kit: true },
+	oneOrTwoWordDescription: 'ORM',
```
AdrianGonz97, line 69: this one's the odd one out given that `i18n` and `css` are all lowercase. `database orm` should be better

packages/core/adder/config.ts
benmccann: Yeah, I'm not worried about the official ones since we control those. We don't have any control over the community adders though and aren't they using the same interface?

https://github.com/sveltejs/cli/blob/e4c41d650a6031ded725d4fade7415a961358a3b/community-adder-template/src/index.js#L13

AdrianGonz97: They do, but I don't think we'll be displaying their descriptions anywhere in the CLI if we're only going invoke them via args:
```shell
sv add tailwindcss some-community-add-on another-one

# skips over the selection prompt

# ...add-on specific questions and so on
```

manuel3108:
Rebasing basically resulted in rewrite. I hate merge conflicts!

packages/core/adder/config.ts
benmccann: Hmm. I don't trust that people wouldn't fill it out with something long. We should perhaps have a different interface for them to use if we're not going to use the field. Or maybe have some validation that only shows it if it's short. Anyway, I can rename it for now

packages/core/addon/config.ts
```diff
@@ -29,6 +29,7 @@ export type SvApi = {
 export type Addon<Args extends OptionDefinition> = {
 	id: string;
 	alias?: string;
+	shortDescription: string;
```
benmccann, line 32: should this be optional like the others? especially if we don't want community adders to have to fill them out?
```suggestion
	shortDescription?: string;
```

manuel3108 (approved):

PR #303, opened 2024-11-14
chore: rename `adders` to `addons`

manuel3108:


README.md
```diff
@@ -13,7 +13,7 @@ Read the [SvelteKit documentation](https://svelte.dev/docs/kit) for more details
 
 ## Contributing
 
-Please file an issue for discussion before sending a PR for a new adder. Most new adders will likely be recommended to be implemented as community adders. This repository will only hold a very limited number of adders at the maintainers discretion that address widely held needs, are considered best-in-class, and are widely used in the Svelte community.
+Please file an issue for discussion before sending a PR for a new addon. Most new addons will likely be recommended to be implemented as community addons. This repository will only hold a very limited number of addons at the maintainers discretion that address widely held needs, are considered best-in-class, and are widely used in the Svelte community.
```
AdrianGonz97, line 16: this is where i was mainly conflicted - i was thinking that we should use `add-on` in all documentation and `addon` for 'internal' file names, dirs, code, etc

manuel3108: Ahh you are right. Forgot about that

manuel3108: Fixed. There are a few spots that feel quite strange, especially in `community-addon-template/README.md` as we are using stuff like `addon-name` while explaining a npx command as `npx degit sveltejs/cli/community-addon-template addon-name`. In that case I would count this as "code" and leave it as is (it's in a code-block for this reason :D)

Another thing I haven't addressed are the inline code comments. I would argue using `addons` there beside the code makes more sense than referencing `add-ons` in comments and `addon` inside code. But this is open for discussion.

AdrianGonz97 (approved):

PR #307, opened 2024-11-15
fix: limit dependency install's output window size

AdrianGonz97:
Fixes a small bug where the output would grow larger than the current window's height, leaving undesirable text behind.


manuel3108 (approved):
This works great btw!

PR #311, opened 2024-11-15
feat: `vitest` use client and server side testing

### LINKED ISSUE #270: testing of client and server code with vitest
`test:unit:client` could run `.svelte` and `.svelte.js`/`.svelte.ts` files with jsdom and svelte testing library (which sets `resolve: browser`)

`test:unit:server` could run the remaining files without those

manuel3108:
Closes #270


manuel3108:
The small problem I see here, is that we now have more or less completely different code for `vite` and `kit`. But since there is nearly no overlap, i don't think there is a good way around this.

packages/addons/vitest-addon/index.ts
```diff
@@ -38,63 +38,143 @@ export default defineAddon({
...
-				const idxToRemove = ast.body.indexOf(defineConfigImportDecl);
-				ast.body.splice(idxToRemove, 1);
-			} else {
-				// otherwise, just remove the `defineConfig` specifier
-				const idxToRemove = defineConfigImportDecl?.specifiers?.findIndex(
-					(s) => s.type === 'ImportSpecifier' && s.imported.name === 'defineConfig'
-				);
-				if (idxToRemove) defineConfigImportDecl?.specifiers?.splice(idxToRemove, 1);
-			}
+		if (kit) {
```
dominikg, line 41: why only with kit? you can use component testing where you'd need testing library outside of kit too

packages/addons/vitest-addon/index.ts
```diff
@@ -38,63 +38,143 @@ export default defineAddon({
...
+					import { describe,test, expect } from 'vitest';
+					import { render, screen } from '@testing-library/svelte';
+					import Page from  './+page.svelte';
 
-			const test = object.create({
-				include: common.expressionFromString("['src/**/*.{test,spec}.{js,ts}']")
+					describe('/+page.svelte',()=>{
+						test('should render h1',()=>{
+							render(Page);
+							expect(screen.getByRole('heading',{level:1})).toBeInTheDocument();
```
benmccann, line 57:
```suggestion
							expect(screen.getByRole('heading', {level:1})).toBeInTheDocument();
```

packages/addons/vitest-addon/index.ts
```diff
@@ -38,63 +38,143 @@ export default defineAddon({
...
+			sv.file('vitest.workspace.ts', (content) => {
+				const { ast, generateCode } = parseScript(content);
+
+				imports.addNamed(ast, 'vitest/config', { defineWorkspace: 'defineWorkspace' });
+				imports.addNamed(ast, '@testing-library/svelte/vite', { svelteTesting: 'svelteTesting' });
+
+				const clientObjectExpression = object.create({
+					extends: common.createLiteral(`./vite.config.${ext}`),
+					plugins: common.expressionFromString(
+						'[svelteTesting({ resolveBrowser: true, autoCleanup: true })]'
```
dominikg, line 82: these 2 might be defaults so we could just use `svelteTesting()` maybe ping someone from testing library svelte here?

packages/addons/vitest-addon/index.ts
```diff
@@ -38,63 +38,143 @@ export default defineAddon({
...
+				if (content) return content;
+
+				return dedent`
+					import '@testing-library/jest-dom/vitest'
+
+					// add global mocks here, i.e. for sveltekit '$app/stores'
+				`;
+			});
+
+			sv.file('vitest.workspace.ts', (content) => {
```
dominikg, line 73: how does this work for non-typescript projects?

dominikg:
requires an update to .gitignore for `vitest.workspace` temporary bundle files, similar to vite.config and vitest.config

depending on which tsconfig is used, vitest.config.ts needs to be added to that too

packages/addons/vitest-addon/index.ts
manuel3108: `vitest` just eats everything. But i have still modified to use `js` if we are in a non `ts` project

manuel3108:
> requires an update to .gitignore for `vitest.workspace` temporary bundle files, similar to vite.config and vitest.config

Does this file get's generated for you? For me i only see `vite.config.ts.timestamp-*` being generated, even while running `pnpm test`. Additionally it seems like `vite` recently merged a [PR](https://github.com/vitejs/vite/pull/18509) that's going to generate those files outside of the project root either way. Looks like this will be a change that will be introduced in [`vite@6`](https://github.com/vitejs/vite/blob/main/packages/vite/CHANGELOG.md#600-beta8-2024-11-01)

> depending on which tsconfig is used, vitest.workspace.ts needs to be added to that too

Why should this be necessary?

manuel3108:
Requires further modifications as pointed out by @dominikg here https://github.com/dominikg/vitest-example-svelte5/pull/1

dominikg:
> Requires further modifications as pointed out by @dominikg here [dominikg/vitest-example-svelte5#1](https://github.com/dominikg/vitest-example-svelte5/pull/1)

those changes are required for vite-6, however given vitest only works with an override (no-go for our cli) or as a beta with that, i suggest we roll back the vite version in our templates to vite-5 until vitest@3 was released and we were able to ensure the config we create actually works.

By the time vitest@3 is out, i hope that users that created projects from now till then can just keep using the setup and update versions. If a breaking change in vitest@3 requires a config change, i wonder if we have the ability to run an `upgrade` command as well?

mxfactorial:
> If a breaking change in vitest@3 requires a config change, i wonder if we have the ability to run an upgrade command as well?

will settle for a closed issue documenting the error and new config

may have to share the link a few times but we'll catch up

manuel3108:
Needs the following things to be done
* use `vite@6` in `create` templates again
* wait for `vitest@3` release

manuel3108:
> Needs the following things to be done
> 
>     * use `vite@6` in `create` templates again
> 
>     * wait for full `vitest@3` release

`vitest@3` just got released and we should theoretically be able to go ahead with this. If anyone want's to pick up where I left of, please go ahead. I won't have any time for this in the next 2-3 weeks.

benmccann:
Vitest 3 upgrade is probably blocked by https://github.com/sveltejs/vite-plugin-svelte/issues/1068

dominikg:
vitest 3.0.1 fixes the issue mentioned above

manuel3108:
@dominikg updated this pr with main, where we are already using the respective versions mentioned above. Made some further tweaks to fix typing and linting issues. 

Would you mind checking if everything works as expected now? For me it looks good now.
```
pnpx https://pkg.pr.new/sveltejs/cli/sv@311 add vitest
```

dominikg:
we might be able to use vite.config.js to define vitest workspaces, vitest 3 added a way to do this. this would save one file. what do you think?

dominikg:
PR that changes my example setup to use vite.config.ts instead of vitest.workspace.ts: https://github.com/dominikg/vitest-example-svelte5/pull/2/files#diff-6a3b01ba97829c9566ef2d8dc466ffcffb4bdac08706d3d6319e42e0aa6890dd

manuel3108:
> we might be able to use vite.config.js to define vitest workspaces, vitest 3 added a way to do this. this would save one file. what do you think?

This is great, updated.

The only thing that's bothering me is the hack for `matchMedia` to make it work. I'm pretty sure that this is going to cause some users to ask questions. But I also saw your recent discord message, so we can just go with it for now.

This other new line that you added though is causing some problems with `eslint`. Any reason for adding that? I think everything was working perfectly before?
https://github.com/dominikg/vitest-example-svelte5/pull/2/files#diff-9dde8fde39332d4adc9f376ff5576f21765bdfebfe03f2e25767feecc0b54a06R5
```
/home/manuel/dev/svelte-cli-temp/vitest-setup-client.ts
  5:5   error  Unexpected constant binary expression. Compares constantly with the right-hand side of the `!=`  no-constant-binary-expression
  5:22  error  Invalid typeof comparison value 
```

dominikg:
oh, i added that when i was experimenting with vitest browser mode where matchMedia is present. sorry, that if wrapper can be removed.

manuel3108:
done

PR #314, opened 2024-11-16
chore: update dependencies script

manuel3108:
Adds a script `pnpm update-dependencies` to automatically update the dependencies. This script focuses on 
- [x] updating `package.json` dependencies (using `pnpm update --recursive`)
- [x] updating dependencies that are used while scaffolding `addons` (using `regex` and `fetch` to npm)

As discussed, updating the actual dependencies (packages and addons) should happen separately, since both also require code changes, although minimal.

### DIFF
--- a/scripts/update-addon-dependencies.js
+++ b/scripts/update-addon-dependencies.js
@@ -19,7 +19,7 @@ async function updateAddonDependencies() {
 		let content = fs.readFileSync(filePath, { encoding: 'utf8' });
 
 		// regex to extract package name and version from `sv.dependency` and `sv.devDependency`
-		const svDepRegex = /sv\.(?:dependency|devDependency)\(['"]([^'"]+)['"],\s*['"]([^'"]+)['"]\)/g;
+		const svDepRegex = /sv\.(?:dependency|devDependency)\('([^']+)',\s*'([^']+)'\)/g;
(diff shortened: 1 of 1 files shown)


AdrianGonz97 (approved):
This is looking great! Ended up making a few tweaks, but this should be ready to go:
- Removed the script that updated the repo's deps since that's something that shouldn't necessarily be tied to updating add-on deps
- renamed the script and file
- added a bit more info to the output and some color so it's easier to visually parse
- added a regex for packages that are defined in object literals (take our [tailwind add-on](https://github.com/sveltejs/cli/blob/622e856d3f8d8b11627e47f289b6feaadddea327/packages/addons/tailwindcss/index.ts#L14-L33) as an example)

manuel3108:
> * Removed the script that updated the repo's deps since that's something that shouldn't necessarily be tied to updating add-on deps

Sounds logical, this was setup before i figured out that `pnpm update -r` exists and i was always trying to remember an external tool name. 

> * added a regex for packages that are defined in object literals (take our [tailwind add-on](https://github.com/sveltejs/cli/blob/622e856d3f8d8b11627e47f289b6feaadddea327/packages/addons/tailwindcss/index.ts#L14-L33) as an example)

Great point!

PR #323, opened 2024-11-22
chore: remove @types/eslint package from eslint addon

BlankParticle:
removed `@types/eslint` package from eslint addon as it is not required anymore
ref https://github.com/DefinitelyTyped/DefinitelyTyped/pull/70735


manuel3108 (approved):

AdrianGonz97 (approved):
thank you!

Issue #248, opened 2024-10-28
Adder for sveltekit adapter


benmccann:
auto should be the default. also offer other official adders

can also use it to customize ignores. see https://github.com/sveltejs/cli/pull/335

manuel3108:
Should this be an adder? Or a question while creating a new project, similar to the `types` question? How would we call the adder?

benmccann:
Interesting. I hadn't considered offering it only during the project setup. I guess that's an option. Though switching out adapter-auto with an adder seems nice

We could call it `sveltekit-adapter`

PR #335, opened 2024-11-27
feat(addons): Respect `.gitignore` in eslint addon

### LINKED ISSUE #334: feat: Add `.vercel` to eslint addon ignores
The `@sveltejs/adapter-vercel` defaults it's output to `.vercel` in the root of the project. The current ignore rules don't ignore this.

Current:
```ts
const ignoresConfig = object.create({
	ignores: common.expressionFromString('["build/", ".svelte-kit/", "dist/"]')
});
```

Happy to fill out a PR if this makes sense to add!

ieedan:
Fixes #334 

Searched the docs and tried the adapters out ended up also adding `.wrangler` and `.netlify`

### DIFF
--- a/.changeset/beige-peas-swim.md
+++ b/.changeset/beige-peas-swim.md
@@ -2,4 +2,4 @@
 'sv': patch
 ---
 
-feat: respect `.gitignore` in eslint addon
+feat: respect `.gitignore` in `eslint` add-on
--- a/.changeset/beige-peas-swim.md
+++ b/.changeset/beige-peas-swim.md
@@ -1,5 +1,5 @@
 ---
-'@sveltejs/addons': patch
+'@sveltejs/addons': minor
 ---
 
 feat: add `@sveltejs/<adapter>` out directories to eslint ignores
--- a/.changeset/beige-peas-swim.md
+++ b/.changeset/beige-peas-swim.md
@@ -2,4 +2,4 @@
 '@sveltejs/addons': patch
 ---
 
-Add `.vercel` to eslint ignore directories.
+Add `@sveltejs/<adapter>` out directories to eslint ignores.
--- a/.changeset/beige-peas-swim.md
+++ b/.changeset/beige-peas-swim.md
@@ -2,4 +2,4 @@
 'sv': patch
 ---
 
-feat: add `@sveltejs/<adapter>` out directories to eslint ignores
+feat: respect `.gitignore` in eslint addon
--- a/.changeset/beige-peas-swim.md
+++ b/.changeset/beige-peas-swim.md
@@ -1,5 +1,5 @@
 ---
-'@sveltejs/addons': minor
+'sv': patch
 ---
 
 feat: add `@sveltejs/<adapter>` out directories to eslint ignores
--- a/.changeset/beige-peas-swim.md
+++ b/.changeset/beige-peas-swim.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/addons': patch
+---
+
+Add `.vercel` to eslint ignore directories.
--- a/.changeset/beige-peas-swim.md
+++ b/.changeset/beige-peas-swim.md
@@ -2,4 +2,4 @@
 '@sveltejs/addons': patch
 ---
 
-Add `@sveltejs/<adapter>` out directories to eslint ignores.
+feat: add `@sveltejs/<adapter>` out directories to eslint ignores
--- a/.changeset/six-moose-admire.md
+++ b/.changeset/six-moose-admire.md
@@ -1,0 +1,5 @@
+---
+'sv': patch
+---
+
+chore: replace svelte-5-preview link
--- a/README.md
+++ b/README.md
@@ -6,10 +6,10 @@ Read the [SvelteKit documentation](https://svelte.dev/docs/kit) for more details
 
 ### Packages
 
-| Package                            | Changelog                                  |
-| ---------------------------------- | ------------------------------------------ |
-| [sv](packages/cli)                 | [Changelog](packages/cli/CHANGELOG.md)     |
-| [svelte-migrate](packages/migrate) | [Changelog](packages/migrate/CHANGELOG.md) |
+| Package                            | Changelog                                  | Documentation                                           |
+| ---------------------------------- | ------------------------------------------ | ------------------------------------------------------- |
+| [sv](packages/cli)                 | [Changelog](packages/cli/CHANGELOG.md)     | [Documentation](https://svelte.dev/docs/cli/overview)   |
+| [svelte-migrate](packages/migrate) | [Changelog](packages/migrate/CHANGELOG.md) | [Documentation](https://svelte.dev/docs/cli/sv-migrate) |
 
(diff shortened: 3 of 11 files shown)


AdrianGonz97:
thank you! however I'm not sure if having to manually update the `ignores` property for every new entry is the best solution (also, TIL that eslint v9 no longer takes the `.gitignore` file into consideration, which is fun). 

perhaps it would be nicer to have the `.gitignore` automatically included using the [`@eslint/compat` package](https://eslint.org/docs/latest/use/configure/ignore#including-gitignore-files)? 

It can be implemented as so:
```js
import { fileURLToPath } from "node:url";
import { includeIgnoreFile } from "@eslint/compat";

const gitignorePath = fileURLToPath(new URL("./.gitignore", import.meta.url));
export default {
 // ...
 includeIgnoreFile(gitignorePath)
}
```

thoughts? @manuel3108 @benmccann

ieedan:
> thank you! however I'm not sure if having to manually update the `ignores` property for every new entry is the best solution (also, TIL that eslint v9 no longer takes the `.gitignore` file into consideration, which is fun).
> 
> perhaps it would be nicer to have the `.gitignore` automatically included using the [`@eslint/compat` package](https://eslint.org/docs/latest/use/configure/ignore#including-gitignore-files)?
> 
> It can be implemented as so:
> 
> ```js
> import { fileURLToPath } from "node:url";
> import { includeIgnoreFile } from "@eslint/compat";
> 
> const gitignorePath = fileURLToPath(new URL("./.gitignore", import.meta.url));
> export default {
>  // ...
>  includeIgnoreFile(gitignorePath)
> }
> ```
> 
> thoughts? @manuel3108 @benmccann

I don't think this is part of the .gitignore? It should be the `ignores` key:
```jsonc
{
	ignores: ['build/', '.svelte-kit/', 'dist/', '.vercel/']
}
```

manuel3108:
Yeah, i think that approach would make sense, although we will have basically the same problem in the `.gitignore` file then. But personally, I don't think this should be implemented in a "hardcoded" way. I would prefer to see this implemented together with #248

manuel3108:
> the `.gitignore` already includes all of the different adapter output dirs anyway (or at least it _should_, in the case of netlify which needs to be fixed)

At this point for me it doesn't look like it's including the vercel output. So that's also missing. But I wouldn't be in favor of adding that globally to the templates, as explained above.

AdrianGonz97:
> At this point for me it doesn't look like it's including the vercel output. So that's also missing.

here's the currently generated `.gitignore`:
```
node_modules

# Output
.output
.vercel
/.svelte-kit
/build

# OS
.DS_Store
Thumbs.db

# Env
.env
.env.*
!.env.example
!.env.test

# Vite
vite.config.js.timestamp-*
vite.config.ts.timestamp-*

```

the only one that's missing is `.netlify` (and maybe wrangler? i normally use `@sveltejs/adapter-cloudflare` which outputs into `.svelte-kit/cloudflare` instead, so idk). 

> But I wouldn't be in favor of adding that globally to the templates, as explained above.

yea but that would only be helpful for people who are changing adapters via `add` and not manually. why not just get them all in one go instead? we're already most of the way there anyway since we're only missing 1 or 2 entries.

ieedan:
I just went through and ran all of the adapters to see what they output since I don't think any of those outputs are actually in the docs for the adapters at least that I could find.

AdrianGonz97:
here are the generated dirs in the root:
```shell
@sveltejs/adapter-node - /build
@sveltejs/adapter-static - /build
@sveltejs/adapter-cloudflare - /.wrangler 
@sveltejs/adapter-cloudflare-workers /.wrangler
@sveltejs/adapter-netlify - /.netlify
@sveltejs/adapter-vercel - /.vercel
```

not sure where `.output` dir that already exists in the `.gitignore` is supposed to be for, but we only need to add `.wrangler` and `.netlify` to the gitignore.

ieedan:
Is there a reason there can't just be a exported constant somewhere that has all of these ignores to be used everywhere?

.changeset/beige-peas-swim.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sveltejs/addons': patch
```
benmccann, line 2:
```suggestion
'@sveltejs/addons': minor
```

.changeset/beige-peas-swim.md
```diff
@@ -0,0 +1,5 @@
...
+---
+
+Add `@sveltejs/<adapter>` out directories to eslint ignores.
```
benmccann, line 5:
```suggestion
feat: add `@sveltejs/<adapter>` out directories to eslint ignores
```

.changeset/beige-peas-swim.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sveltejs/addons': minor
```
AdrianGonz97, line 2: sorry! `addons` is a private package so we'll need to increment `sv` instead, which is also in pre-release so it should also be a `patch`
```suggestion
'sv': patch
```

AdrianGonz97:
> Is there a reason there can't just be a exported constant somewhere that has all of these ignores to be used everywhere?

not exactly sure I follow - who should be exporting these? the individual adapter packages?

ieedan:
> > Is there a reason there can't just be a exported constant somewhere that has all of these ignores to be used everywhere?
> 
> not exactly sure I follow - who should be exporting these? the individual adapter packages?

I don't think it being a separate package would be practical but it may be nice to have an internal package to use for it? Like a @sveltejs/constants or something idk.

AdrianGonz97:
eh, that would be pretty niche and not particularly helpful when it comes to the `gitignore` file since there's no way to 'import' those constants into it

anyway, I think we're in agreement that implementing https://github.com/sveltejs/cli/pull/335#issuecomment-2504356903 would be good, so feel free to add it to this PR!

ieedan:
I meant more it could just be passed around to build whatever files need it.

Either way let me wrap a few things up and I can give it a shot!

ieedan:
This should be it but I cannot seem to get it to build or run tests on my machine I am probably just doing something wrong.

manuel3108 (approved):
Thanks, so much! This seems to be working great!

The only thing im unsure about, is the placement of the `includeIgnoreFile` call. According to the [readme](https://www.npmjs.com/package/@eslint/compat) I would understand that this should be placed first in the eslint array, but currently it's placed last. In my local testing though, this didn't make any difference at all.

ieedan:
> Thanks, so much! This seems to be working great!
> 
> The only thing im unsure about, is the placement of the `includeIgnoreFile` call. According to the [readme](https://www.npmjs.com/package/@eslint/compat) I would understand that this should be placed first in the eslint array, but currently it's placed last. In my local testing though, this didn't make any difference at all.

Yeah if you would rather have it at the beginning I can move it no problem at all!

ieedan:
Also should I update the template git ignores with the output files mentioned above?

AdrianGonz97:
Not at the moment, but we have plans on addressing issues similar to this in the future

PR #337, opened 2024-11-29
Paraglide: Fix typing for newLanguage in demo code

msgeissler:
The paraglide demo contains sample code for a language switcher with typing in JSDoc.

Currently the typing is not working, as it is missing curly brackets.


manuel3108 (approved):
Thanks!

PR #343, opened 2024-12-03
feat: set app templates to `private` by default

### LINKED ISSUE #342: Set `package.json` `private` to `true` by default
https://docs.npmjs.com/cli/v10/configuring-npm/package-json#private

Its a nice to have. It prevents dumb mistakes when running npm publish in the wrong terminal.

manuel3108:
Closes #342

### DIFF
--- a/.changeset/honest-bugs-brush.md
+++ b/.changeset/honest-bugs-brush.md
@@ -2,4 +2,4 @@
 'sv': patch
 ---
 
-fix: `create` mark created apps as private
+feat: set app templates to `private` by default
--- a/.changeset/honest-bugs-brush.md
+++ b/.changeset/honest-bugs-brush.md
@@ -1,0 +1,5 @@
+---
+'sv': patch
+---
+
+fix: `create` mark created apps as private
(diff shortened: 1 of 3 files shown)


packages/create/templates/demo/package.template.json
```diff
@@ -2,6 +2,7 @@
 	"name": "~TODO~",
 	"version": "0.0.1",
 	"type": "module",
+	"private": true,
```
benmccann, line 5: it's pretty subjective, but I usually put `private` right after `name` in my projects because it's so important to know if something will be published or not and you really don't want to miss whether or not it's there and accidentally publish something

benmccann (approved):

.changeset/honest-bugs-brush.md
```diff
@@ -0,0 +1,5 @@
+---
+'sv': patch
+---
+
+fix: `create` mark created apps as private
```
AdrianGonz97, line 5:
```suggestion
feat: set app templates to `private` by default
```

UltraCakeBakery:
Thanks for this!

PR #346, opened 2024-12-03
feat: adds `sveltekit-adapter` add-on

### LINKED ISSUE #248: Adder for sveltekit adapter

AdrianGonz97:
closes #248

`sveltekit-adapter` installs and replaces existing SvelteKit adapters.

usage:
```shell
sv add --sveltekit-adapter <node|static|vercel|cloudflare-pages|cloudflare-workers|netlify>

### DIFF
--- a/packages/addons/sveltekit-adapter/index.ts
+++ b/packages/addons/sveltekit-adapter/index.ts
@@ -28,7 +28,7 @@ const options = defineAddonOptions({
 
 export default defineAddon({
 	id: 'sveltekit-adapter',
-	alias: 'adapater',
+	alias: 'adapter',
 	shortDescription: 'deployment',
 	homepage: 'https://svelte.dev/docs/kit/adapters',
 	options,
--- a/packages/addons/sveltekit-adapter/index.ts
+++ b/packages/addons/sveltekit-adapter/index.ts
@@ -28,8 +28,7 @@ const options = defineAddonOptions({
 
 export default defineAddon({
 	id: 'sveltekit-adapter',
-	// TODO: description?
-	shortDescription: '',
+	shortDescription: 'deployment',
 	homepage: 'https://svelte.dev/docs/kit/adapters',
 	options,
 	setup: ({ kit, unsupported }) => {
@@ -38,7 +37,6 @@ export default defineAddon({
 	run: ({ sv, options }) => {
 		const adapter = adapters.find((a) => a.id === options.adapter)!;
 
-		// TODO: should we do this?
 		// removes previously installed adapters
 		sv.file('package.json', (content) => {
 			const { data, generateCode } = parseJson(content);
--- a/packages/addons/sveltekit-adapter/index.ts
+++ b/packages/addons/sveltekit-adapter/index.ts
@@ -1,0 +1,102 @@
+import { defineAddon, defineAddonOptions } from '@sveltejs/cli-core';
+import { exports, functions, imports, object, type AstTypes } from '@sveltejs/cli-core/js';
+import { parseJson, parseScript } from '@sveltejs/cli-core/parsers';
+
+type Adapter = {
+	id: string;
+	package: string;
+	version: string;
+};
+
+const adapters: Adapter[] = [
+	{ id: 'node', package: '@sveltejs/adapter-node', version: '^5.2.9' },
+	{ id: 'static', package: '@sveltejs/adapter-static', version: '^3.0.6' },
+	{ id: 'vercel', package: '@sveltejs/adapter-vercel', version: '^5.5.0' },
+	{ id: 'cloudflare-pages', package: '@sveltejs/adapter-cloudflare', version: '^4.8.0' },
+	{ id: 'cloudflare-workers', package: '@sveltejs/adapter-cloudflare-workers', version: '^2.6.0' },
+	{ id: 'netlify', package: '@sveltejs/adapter-netlify', version: '^4.4.0' }
+];
+
+const options = defineAddonOptions({
+	adapter: {
+		type: 'select',
+		question: 'Which SvelteKit adapter would you like to use?',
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

File: packages/addons/sveltekit-adapter/index.ts [hunk 1]
```diff
@@ -0,0 +1,102 @@
...
+			const devDeps = data['devDependencies'];
+
+			for (const pkg of Object.keys(devDeps)) {
+				if (pkg.startsWith('@sveltejs/adapter-')) {
+					delete devDeps[pkg];
+				}
+			}
+
+			return generateCode();
+		});
```
AdrianGonz97, line 54: This portion removes any previously installed adapters. Is this a good/bad idea? Thoughts?

packages/addons/sveltekit-adapter/index.ts
```diff
@@ -0,0 +1,102 @@
...
+		question: 'Which SvelteKit adapter would you like to use?',
+		options: adapters.map((p) => ({ value: p.id, label: p.id, hint: p.package })),
+		default: 'node'
+	}
+});
+
+export default defineAddon({
+	id: 'sveltekit-adapter',
+	// TODO: description?
+	shortDescription: '',
```
AdrianGonz97, line 32: Couldn't think of a good description of it besides repeating it's name `sveltekit adapters`. Any ideas?

benmccann: Maybe `deployment`?

packages/addons/sveltekit-adapter/index.ts
benmccann: It's going to be the right thing to do most of the time, so it's probably fine. It'd be annoying in the case where you don't want it, but I think it'd be pretty rare that you're deploying to multiple places and in that case the add-on isn't going to be able to help you much anyway as you're going to have to do something rather custom

packages/addons/sveltekit-adapter/index.ts
```diff
@@ -0,0 +1,100 @@
...
+};
+
+const adapters: Adapter[] = [
+	{ id: 'node', package: '@sveltejs/adapter-node', version: '^5.2.9' },
+	{ id: 'static', package: '@sveltejs/adapter-static', version: '^3.0.6' },
+	{ id: 'vercel', package: '@sveltejs/adapter-vercel', version: '^5.5.0' },
+	{ id: 'cloudflare-pages', package: '@sveltejs/adapter-cloudflare', version: '^4.8.0' },
+	{ id: 'cloudflare-workers', package: '@sveltejs/adapter-cloudflare-workers', version: '^2.6.0' },
+	{ id: 'netlify', package: '@sveltejs/adapter-netlify', version: '^4.4.0' }
+];
```
manuel3108, line 18: Did you had any specific ordering in mind? I would have expected the same ordering as on the website: https://svelte.dev/docs/kit/adapters.

Seems to be mostly ordered by popularity (expect netlify): https://npmcharts.com/compare/@sveltejs/adapter-node,@sveltejs/adapter-static,@sveltejs/adapter-vercel,@sveltejs/adapter-cloudflare,@sveltejs/adapter-cloudflare-workers,@sveltejs/adapter-netlify

Should there also be an option to adapter auto? I think that would be strange and we should avoid it.

packages/addons/sveltekit-adapter/index.ts
```diff
@@ -0,0 +1,100 @@
...
+	adapter: {
+		type: 'select',
+		question: 'Which SvelteKit adapter would you like to use?',
+		options: adapters.map((p) => ({ value: p.id, label: p.id, hint: p.package })),
+		default: 'node'
+	}
+});
+
+export default defineAddon({
+	id: 'sveltekit-adapter',
```
manuel3108, line 30:
```suggestion
	alias: 'adapter',
```
Should we add an alias?

AdrianGonz97: yea, I wasn't sure if this was a good idea or not. I'm not completely against it though. I _don't_ think we'll have another `X-adapter` add-on anyway, so we could probably get away with it

Rich-Harris:
preview: https://svelte-dev-git-preview-cli-346-svelte.vercel.app/

manuel3108:
> Also some we have the single vs. double quotes problem again, I think we already solved this somewhere else, but I don't remember where and how.

This still seems to be present, but looking at the code, no clue how this is happening and if this is fixable.

For whatever reason the following command is installing the `node` adapter, without asking any questions: 
```
pnpm sv add --sveltekit-adapter --cwd ../svelte-cli-temp
```

Also, it looks like there is an inconsistency with the aliases:
```bash
pnpm sv add sveltekit-adapter --cwd ../svelte-cli-temp     # works
pnpm sv add adapter --cwd ../svelte-cli-temp               # works
pnpm sv add --sveltekit-adapter --cwd ../svelte-cli-temp   # works
pnpm sv add --adapter --cwd ../svelte-cli-temp             # error: unknown option '--adapter'
```
The last one could be solved in a separate PR as this is assumably a broader issue

AdrianGonz97:
> This still seems to be present, but looking at the code, no clue how this is happening and if this is fixable.
> 
> For whatever reason the following command is installing the `node` adapter, without asking any questions:
> 
> ```
> pnpm sv add --sveltekit-adapter --cwd ../svelte-cli-temp
> ```

That's the intended behavior. If you're specifying the adapter via the flag, then any _unspecified_ add-on options will fallback to their defaults (which is `node` in this case).

> Also, it looks like there is an inconsistency with the aliases:

I don't think we ever went over if aliases should also get their own flags. At the moment, they're only useful when specifying add-ons as args (i.e. `sv add tailwind`).

manuel3108 (approved):

PR #354, opened 2024-12-12
feat: add app-state migration

dummdidumm:
A simple, string-based migration for https://github.com/sveltejs/kit/pull/13140 (tested on that repo, zero false-positives, less than five false negatives)

We can switch to using the Svelte parser if people prefer that to get a bit better with regards to only actually transforming identifiers, but to me this seems to work fine.


packages/migrate/migrations/page-state/migrate.js
```diff
@@ -0,0 +1,74 @@
...
+	if (!import_match) return code; // nothing to do
+
+	const stores = import_match[1].split(',').map((i) => i.trim());
+	let modified = code.replace('$app/stores', '$app/state');
+
+	for (const store of stores) {
+		// if someone uses that they're deep into stores and we better not touch this file
+		if (store === 'getStores') return code;
+
+		const regex = new RegExp(`\\b${store}\\b`, 'g');
```
paoloricciuti, line 42: I might be missing something since I'm looking at it from the phone but this doesn't account for

```ts
import { page as page_store } from "$app/stores";
```
right? It will try to substitute the whole `page as page_store` (and probably fail to do so)

dummdidumm: Yes but IMO that's fine, I haven't seen a single rename anywhere, and even if, it's just that one file that needs manual migration.

As I said in the pr description this is simple text replacement but it works surprisingly well given the unique constraints of dollar prefixes etc

paoloricciuti: Oh yeah...just thought it's probably decently easy to add since is still a determined syntax, but we can probably skip it

paoloricciuti: I did a quick search out of mere curiosity 

https://github.com/search?q=path%3A*.svelte+%2Fimport.%2Bas.%2Bfrom%5Cs%2B%28%5B%27%7C%22%5D%29%5C%24app%5C%2Fstores%2F&type=code

It's objectively rare (200 something instances on millions of files)

Rich-Harris:
preview: https://svelte-dev-git-preview-cli-354-svelte.vercel.app/

packages/migrate/migrations/page-state/migrate.js
dummdidumm: add support for aliasing

Rich-Harris:
Should we make this an `app-state` migration and include `navigating` and `updated`?

dummdidumm:
This already contains migrations for navigating (though the one with current on it) and updated - page-state is a misnomer, you're right that we should rename it to app-state

packages/migrate/migrations/app-state/migrate.js
```diff
@@ -0,0 +1,77 @@
...
+					match.index - count_removed <
+						/** @type {number} */ (import_match.index) + import_match[0].length
+				) {
+					continue; // this is the import statement
+				}
+
+				return code;
+			}
+
+			modified = before.slice(0, -1) + alias + (store === 'updated' ? '.current' : '') + after;
```
dummdidumm, line 71: For navigating this need to handle the `$navigating` case where just `navigating` would do nothing. I.e.

```diff
- $navigating
+ navigating.to
```

Rich-Harris: yeah... i don't really know what to do about this. we can't just substitute `navigating.to` everywhere we currently see `$navigating`. probably needs a `@migration task`, how would we go about adding one?

dummdidumm: Mhm you're right in case where it's not used as an if block guard but passed somewhere. So we could add a migration if we detect the prior string is `{#if `, and else bail and return `code`. If a migration task is wanted we could just prepend an html comment at the top

Rich-Harris (approved):

benmccann:
Heads up that this only found 9 out of 22 occurrences of `$app/stores` when I ran it on a real project. I filed an issue here: https://github.com/sveltejs/cli/issues/368

PR #360, opened 2024-12-16
fix(create): Resolve package manager commands correctly in next steps

### LINKED ISSUE #359: Typo: ```deno dev --open``` instead of ```deno task dev --open```
There is a typo in step 3 when choosing deno as the package manager.

image

AdrianGonz97:
closes #359

### DIFF
--- a/.changeset/chatty-queens-judge.md
+++ b/.changeset/chatty-queens-judge.md
@@ -1,0 +1,5 @@
+---
+'sv': patch
+---
+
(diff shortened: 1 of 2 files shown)


manuel3108 (approved):
TBH, I did not manage to test the deno specific case, as the cli is never showing `deno` in the next steps for me. But since this is also happening with the latest release, this seems to be a setup problem on my side.

I managed to confirm that the calls for `pnpm` and `npm` are printed correctly, so this should be fine.

PR #361, opened 2024-12-16
fix: use modern Spring implementation

Rich-Harris:
how does one go about testing these things locally? I tried doing `pnpm build:dist` and then running `sv create` like so...

```js
node ../../cli/packages/cli/dist/bin.js create spring-demo
```

...but it gave me the old files.

### DIFF
--- a/.changeset/tiny-colts-trade.md
+++ b/.changeset/tiny-colts-trade.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/create': patch
+---
+
+fix: use modern `Spring` implementation
--- a/packages/create/templates/demo/package.json
+++ b/packages/create/templates/demo/package.json
@@ -11,8 +11,8 @@
 	"devDependencies": {
 		"@fontsource/fira-mono": "^5.0.0",
 		"@neoconfetti/svelte": "^2.0.0",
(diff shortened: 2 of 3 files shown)


AdrianGonz97:
ah, yea... we should really make a `CONTRIBUTING.md` guide. thanks for the reminder 😅 

to test a local build, you can run the following from the root of the monorepo:
```shell
pnpm build

pnpm sv create temp/spring-demo
```

AdrianGonz97 (commented):
in addition to `Spring`, the [`reducedMotion` store](https://github.com/sveltejs/cli/blob/3b9fbcd04d9ad920121d9c483bcc081d05bc4935/packages/create/templates/demo/src/routes/sverdle/reduced-motion.ts#L11) in the demo could also be updated to use the new `MediaQuery` class as well!

manuel3108:
> in addition to `Spring`, the [`reducedMotion` store](https://github.com/sveltejs/cli/blob/3b9fbcd04d9ad920121d9c483bcc081d05bc4935/packages/create/templates/demo/src/routes/sverdle/reduced-motion.ts#L11) in the demo could also be updated to use the new `MediaQuery` class as well!

This is crazy and saves so much code, add this!

manuel3108 (approved):

PR #363, opened 2024-12-17
fix: make `lucia` validation error messages more descriptive

fbehrens:


fbehrens:
This message would have helped me to understand, what the problem was.
I did not realize that the file belongs to the svelte monorepo.
Probably the commit message is not adequate.
Anyway thanks for reading

manuel3108 (approved):
Good catch, thanks for your help!

PR #365, opened 2024-12-18
feat: incorporate `<svelte:html>` into templates

dummdidumm:
Related to https://github.com/sveltejs/kit/pull/13065

TBH I don't love the fact that it's one more file for the minimal template and the library template, but it is what it is.


manuel3108:
Marking as draft, as the other PR are drafts too, and have received at least some pushback. We cannot merge this without the other PRs.

PR #369, opened 2024-12-21
Deno also requires `deno task dev`

phocks:
The output says to run `3: deno dev --open` but it requires `deno task dev --open`

This is a quick fix.


manuel3108:
Duplicates #360, but thanks for trying to help!

PR #382, opened 2024-12-31
use Spring class

Thiagolino8:
The demo, by trying to use 2 different reactivity systems (runes and stores) needed an effect for synchronization and to ignore a warning about using state in a non-reactive way
Now that the Spring class exists, which uses runes and is much more powerful than the store, the demo can be simplified


manuel3108:
Hey, thanks for your work. We do already have a similar PR open, see #361 that we are likely to merge latest next week. 

Thanks for your effort tho!

Closing as duplicate of #361

PR #391, opened 2025-01-09
feat: allow sub path migration or none at all

### LINKED ISSUE #390: Specify paths to be ignored by svelte-migrate
`svelte-migrate` currently gives the option to select paths to be migrated, but those are only first-level child directories. I'd like to be able to specify paths to explicitly ignore.

**When is this useful?**
I've recently came across this when trying to migrate to `svelte@5` while using [shadcn-svelte](https://github.com/huntabyte/shadcn-svelte) - a library which saves components to the `src/lib/components` and migrating them may not be the desired behaviour.

A good approach to avoid any clutter in the cli would be to use an optional argumen/option when running `sv migrate` to explicitly ignore certain paths.

dummdidumm:
This adds two new prompts to the Svelte migration:
- custom folder prompt (from multiselect); if selected you can specify what folders exactly to migrate. closes #390
- adds a way to skip migrating your components if you want to do it incrementally but still benefit from having your package.json and entry points migrated automatically

### DIFF
--- a/packages/migrate/migrations/svelte-5/index.js
+++ b/packages/migrate/migrations/svelte-5/index.js
@@ -140,7 +140,7 @@ export async function migrate() {
 			.concat([
 				{
 					title: 'custom (overrides selection, allows to specify sub folders)',
-					value: ',',
+					value: ',', // a value that definitely isn't a valid folder name so it cannot clash
 					selected: false
 				}
 			])
@@ -154,7 +154,7 @@ export async function migrate() {
 		const custom = await prompts({
 			type: 'list',
 			name: 'value',
-			message: 'Specify folders (comma separated)'
+			message: 'Specify folder paths (comma separated)'
 		});
 
 		if (!custom.value) {
--- a/packages/migrate/migrations/svelte-5/index.js
+++ b/packages/migrate/migrations/svelte-5/index.js
@@ -18,10 +18,6 @@ export async function migrate() {
 		bail('Please re-run this script in a directory with a package.json');
 	}
 
-	console.log(
-		'This migration is experimental — please report any bugs to https://github.com/sveltejs/svelte/issues'
-	);
-
 	const pkg = JSON.parse(fs.readFileSync('package.json', 'utf8'));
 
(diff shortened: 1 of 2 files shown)


packages/migrate/migrations/svelte-5/index.js
```diff
@@ -141,12 +137,41 @@ export async function migrate() {
...
 
 	if (!folders.value?.length) {
 		process.exit(1);
 	}
 
+	if (folders.value.includes(',')) {
+		const custom = await prompts({
+			type: 'list',
+			name: 'value',
+			message: 'Specify folders (comma separated)'
```
paoloricciuti, line 157: Could `folders` be confusing here? As a user should i specify the name of the folder or the full path? I know it's the latter, but could it be ambiguous?

kristiankunc: Agree with this, feel like "path" or something in that fashion would be more descriptive.

packages/migrate/migrations/svelte-5/index.js
```diff
@@ -141,12 +137,41 @@ export async function migrate() {
 				(dir) => fs.statSync(dir).isDirectory() && dir !== 'node_modules' && !dir.startsWith('.')
 			)
 			.map((dir) => ({ title: dir, value: dir, selected: true }))
+			.concat([
+				{
+					title: 'custom (overrides selection, allows to specify sub folders)',
+					value: ',',
```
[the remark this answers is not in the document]
dummdidumm: The thing is, `custom` could be an actual directory name so I took a value that definitely can't be a valid folder. Should add a comment I guess

packages/migrate/migrations/svelte-5/index.js
benmccann: I would say "folder paths" assuming it has to be a folder and can't be a file

packages/migrate/migrations/svelte-5/index.js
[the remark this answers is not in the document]
manuel3108: Ohhh, I seee

manuel3108 (approved):

PR #395, opened 2025-01-12
feat: whitelist postinstall scripts for `pnpm`

### LINKED ISSUE #388: Run install scripts in `pnpm@10`
In [`pnpm@10`](https://github.com/pnpm/pnpm/releases/tag/v10.0.0), install scripts must be individually enabled.

> Lifecycle scripts of dependencies are not executed during installation by default!

If not, the following message will be logged.

> The following dependencies have build scripts that were ignored: @sveltejs/kit, esbuild

> To allow the execution of build scripts for these packages, add their names to "pnpm.onlyBuiltDependencies" in your "package.json", then run "pnpm rebuild"

What if this field is added to the `package.json` if the user selects pnpm?

```
│  pnpm
```

```json
{
  "pnpm": {
    "onlyBuiltDependencies": [
      "@sveltejs/kit",
      "esbuild"
    ]
  }
}
```

manuel3108:
Resolves #388

### DIFF
--- a/.changeset/selfish-pandas-fail.md
+++ b/.changeset/selfish-pandas-fail.md
@@ -1,0 +1,5 @@
+---
+'sv': patch
+---
+
+feat: whitelist postinstall scripts for `pnpm`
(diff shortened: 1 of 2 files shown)


manuel3108:
Potentially not necessary anymore, leaving it as draft for the moment.

manuel3108:
Based on #388 this is no longer the correct way.

PR #396, opened 2025-01-14
chore: utilize prepack lifecycle script

hyunbinseo:
The `prepack` lifecycle script runs before the [`npm pack`](https://docs.npmjs.com/cli/v10/using-npm/scripts#npm-pack) and `npm publish` commands.

By utilizing this:

1. the `prepublishOnly` lifecycle script can be removed, which contained an `npm` command.[^1]
2. it works with `npm pack`.

> [!NOTE]  
> The `publint@0.3` CLI does run the `pack` command, but it does not run the lifecycle scripts.
>
> So `&& publint` should be included at the end of the `prepack` script.
>
> This is correctly implemented from [`publint@0.3.2`](https://github.com/publint/publint/releases/tag/publint%400.3.2), thus the SemVer is updated to `^0.3.2`.
>
> - https://github.com/publint/publint/issues/129#issuecomment-2589151355

[^1]: I hope they are changed to [`node --run`](https://nodejs.org/api/cli.html#--run) in the future when SvelteKit requires Node.js 22+.

### DIFF
--- a/.changeset/pretty-balloons-leave.md
+++ b/.changeset/pretty-balloons-leave.md
@@ -1,5 +1,5 @@
 ---
-'@sveltejs/create': patch
+'sv': patch
 ---
 
 chore: utilize prepack lifecycle script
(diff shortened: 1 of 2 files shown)


manuel3108:
Thanks!

PR #376, opened 2024-12-24
chore: align `migrate` dependencies with `sv`

manuel3108:
Initial goal here was to unify the UX between stuff like `sv create` and `svelte-migrate`

done:
- use `picocolors` instead of `kleur`
- use `@clack/prompts` instead of `prompts`

todo:
- [x] print package version and intro similar to `sv`
- [x] print outro (also if action is cancelled or process is exited)
- [x] remove most `console.log` as they look ugly inside the prompts and convert them to notes or information etc.
- [x] unify code (there is so much trivially duplicated code)
- [x] closes #302 

I opted to not use our own `@sveltejs/clack-prompts` package and instead use the public one. Using our own package would either require us to publish the package, or include `svelte-migrate` inside our own `rollup` config to bundle it. Not a fan of both options, thus decided to use the public package instead.

PR #397, opened 2025-01-15
feat: add `sv kit` as alias for `svelte-kit` command

### LINKED ISSUE #101: Consider aliasing other Svelte CLIs
From @dominikg:
> svelte-kit cli we still have (mostly for the sync command). Do we have plans to alias that in sv as well? (what about svelte-package)

rChaoz:
(Partially?) closes #101

Adds a new alias command `sv kit` that calls `svelte-kit`, and documents it.

### DIFF
--- a/documentation/docs/20-commands/50-sv-kit.md
+++ b/documentation/docs/20-commands/50-sv-kit.md
@@ -1,0 +1,24 @@
+---
+title: sv kit
+---
+
+`sv kit` is a tiny CLI tool that helps you initialize and update your SvelteKit project.
+
+As SvelteKit projects use [Vite](https://vitejs.dev/), you'll mostly be using Vite's CLI commands to build and run your project:
+- `vite dev` — start a development server
+- `vite build` — build a production version of your app
+- `vite preview` — run the production version locally
+
+## Installation
+
+`sv kit` is available in SvelteKit projects, i.e. you will need the `@sveltejs/kit` package to use it. You can use `sv create` to [create and set up](sv-create) a new SvelteKit project. You can also check out SvelteKit's [documentation](/docs/kit) for more information.
+
+## Usage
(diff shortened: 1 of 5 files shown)


benmccann:
I'm not sure this really makes sense. We're executing sv via npx, but SvelteKit is installed. We had at one point talked about having sv installed as well in which case additional aliases would make more sense. But we didn't end up going that direction. We should probably close the issue instead

Conduitry:
Yeah, I'd be in favor of closing this and the corresponding issue. I don't think we want to have `sv` be a dependency of SvelteKit apps - there's no reason for them to have all the code needed to scaffold new projects or to migrate code. And if `sv` isn't a dependency, we don't want to encourage people to pull it down with `npx` just to have it call a dependency that they already have.

rChaoz:
> We're executing sv via npx, but SvelteKit is installed.

I had no idea, I always install `sv` instead, my `check` script even does `sv check` instead of `svelte-check`, I thought this was the intended direction; is this not why `sv check` exists?

If the intended direction is for `sv` to be just a tool for bootstrapping projects, I think this should be more clear, as currently it's [defined as](https://svelte.dev/docs/cli/overview):

> The command line interface (CLI), sv, is a toolkit for creating and maintaining Svelte applications.

manuel3108:
> I thought this was the intended direction; is this not why `sv check` exists?

It is. Though changes need time. We are intending to better integrate `migrate` and `check` commands to behave more similarly to the current `create` and `add` commands. For `migrate` there is an open pr with an initial attempt at #376. Something similar is probably going to happen to `check` in the future. 

As for the `kit` command, this is kinda special and I agree with the opinions above that this is not something we should consider right now. We might re-consider this in the future though.

Thanks for your effort though!

PR #400, opened 2025-01-16
fix: tailwind plugins as dev dependencies

hyunbinseo:
Tailwind CSS plugins should be installed as `devDependencies` alongside `tailwindcss`.

```shell
npm install -D @tailwindcss/forms
```

https://github.com/tailwindlabs/tailwindcss-forms#installation

### DIFF
--- a/.changeset/twelve-pets-film.md
+++ b/.changeset/twelve-pets-film.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/addons': patch
+---
+
+fix: tailwind plugins as dev dependencies
--- a/.changeset/twelve-pets-film.md
+++ b/.changeset/twelve-pets-film.md
@@ -1,5 +1,5 @@
 ---
-'@sveltejs/addons': patch
+'sv': patch
 ---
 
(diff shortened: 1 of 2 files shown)


benmccann (commented):
Thanks for catching this! Agreed!

We probably need to change the package in the changeset. I wonder if there's a way we can make the changeset CLI do the correct thing but default. Otherwise perhaps we should mention it in the PE template and contributors guide

.changeset/twelve-pets-film.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sveltejs/addons': patch
```
manuel3108, line 2:
```suggestion
'sv': patch
```

manuel3108:
> We probably need to change the package in the changeset. I wonder if there's a way we can make the changeset CLI do the correct thing but default. Otherwise perhaps we should mention it in the PE template and contributors guide

Agreed, opened #403 for tracking.

manuel3108 (approved):
Thanks!

PR #420, opened 2025-01-23
fix: sv check not allowing any arguments

### LINKED ISSUE #417: `sv check` doesn't accept any arguments
```shell
$ sv check --tsconfig ./tsconfig.json
> error: too many arguments for 'check'. Expected 0 arguments but got 2.
```

This happens when upgrading from 0.6.11 to 0.6.12. I tried looking at the [changes between these versions](https://github.com/sveltejs/cli/compare/sv%400.6.11...sv%400.6.12) but I can't figure out what's causing it. Looking at the code it seems like this line:

https://github.com/sveltejs/cli/blob/43f73a289b9aa2da685ef15b601ee9f5ffb5b924/packages/cli/commands/check.ts#L13

stopped doing it's job for whatever reason.

rChaoz:
Closes #417

### DIFF
--- a/.changeset/cold-ligers-brake.md
+++ b/.changeset/cold-ligers-brake.md
@@ -1,0 +1,5 @@
+---
+'sv': patch
+---
+
+fix: sv check not allowing any arguments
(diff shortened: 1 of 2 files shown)


AdrianGonz97 (approved):
Thank you!

PR #422, opened 2025-01-23
feat: add `tailwindcss` 4.0

jycouet:
- [x] Tests updated to `oklch` 
- [x] Typography is working with out any plugin 🥳

### DIFF
--- a/.changeset/empty-geese-draw.md
+++ b/.changeset/empty-geese-draw.md
@@ -1,5 +1,5 @@
 ---
-'@sveltejs/addons': patch
+'sv': minor
 ---
 
-using tailwindcss 4.0.0
+feat: update to `tailwindcss` v4.0.0
--- a/.changeset/empty-geese-draw.md
+++ b/.changeset/empty-geese-draw.md
@@ -1,0 +1,5 @@
+---
+'@sveltejs/addons': patch
+---
+
+using tailwindcss 4.0.0
--- a/.changeset/empty-geese-draw.md
+++ b/.changeset/empty-geese-draw.md
@@ -1,5 +1,5 @@
 ---
-'sv': minor
+'sv': patch
 ---
 
 feat: update to `tailwindcss` v4.0.0
--- a/packages/addons/tailwindcss/index.ts
+++ b/packages/addons/tailwindcss/index.ts
@@ -1,14 +1,6 @@
 import { defineAddon } from '@sveltejs/cli-core';
 import { addImports } from '@sveltejs/cli-core/css';
-import {
-	array,
-	common,
-	functions,
-	imports,
-	object,
-	variables,
-	exports
-} from '@sveltejs/cli-core/js';
+import { array, functions, imports, object, exports } from '@sveltejs/cli-core/js';
 import { parseCss, parseJson, parseScript, parseSvelte } from '@sveltejs/cli-core/parsers';
 import { addSlot } from '@sveltejs/cli-core/html';
 
--- a/packages/addons/tailwindcss/index.ts
+++ b/packages/addons/tailwindcss/index.ts
@@ -32,7 +32,7 @@ export default defineAddon({
 			const { ast, generateCode } = parseScript(content);
 
 			const vitePluginName = 'tailwindcss';
-			imports.addNamed(ast, '@tailwindcss/vite', { tailwindcss: vitePluginName });
+			imports.addDefault(ast, '@tailwindcss/vite', vitePluginName);
 
 			const { value: rootObject } = exports.defaultExport(ast, functions.call('defineConfig', []));
 			const param1 = functions.argumentByIndex(rootObject, 0, object.createEmpty());
--- a/packages/addons/tailwindcss/index.ts
+++ b/packages/addons/tailwindcss/index.ts
@@ -4,41 +4,41 @@ import { array, common, exports, imports, object } from '@sveltejs/cli-core/js';
 import { parseCss, parseScript, parseJson, parseSvelte } from '@sveltejs/cli-core/parsers';
 import { addSlot } from '@sveltejs/cli-core/html';
 
-type Plugin = {
-	id: string;
-	package: string;
-	version: string;
-	identifier: string;
-};
+// type Plugin = {
+// 	id: string;
+// 	package: string;
+// 	version: string;
+// 	identifier: string;
+// };
 
-const plugins: Plugin[] = [
-	{
-		id: 'typography',
-		package: '@tailwindcss/typography',
-		version: '^0.5.16',
-		identifier: 'typography'
-	},
-	{
-		id: 'forms',
-		package: '@tailwindcss/forms',
-		version: '^0.5.10',
-		identifier: 'forms'
-	},
-	{
-		id: 'container-queries',
-		package: '@tailwindcss/container-queries',
-		version: '^0.1.1',
-		identifier: 'containerQueries'
-	}
-];
+// const plugins: Plugin[] = [
+// 	{
+// 		id: 'typography',
+// 		package: '@tailwindcss/typography',
+// 		version: '^0.5.16',
+// 		identifier: 'typography'
+// 	},
+// 	{
+// 		id: 'forms',
+// 		package: '@tailwindcss/forms',
+// 		version: '^0.5.10',
+// 		identifier: 'forms'
+// 	},
+// 	{
+// 		id: 'container-queries',
+// 		package: '@tailwindcss/container-queries',
+// 		version: '^0.1.1',
+// 		identifier: 'containerQueries'
+// 	}
+// ];
 
 const options = defineAddonOptions({
-	plugins: {
-		type: 'multiselect',
-		question: 'Which plugins would you like to add?',
-		options: plugins.map((p) => ({ value: p.id, label: p.id, hint: p.package })),
-		default: []
-	}
+	// plugins: {
+	// 	type: 'multiselect',
+	// 	question: 'Which plugins would you like to add?',
+	// 	options: plugins.map((p) => ({ value: p.id, label: p.id, hint: p.package })),
+	// 	default: []
+	// }
 });
 
 export default defineAddon({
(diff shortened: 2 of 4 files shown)


packages/addons/tailwindcss/index.ts
```diff
@@ -1,121 +1,25 @@
...
-			object.property(pluginsObject, 'autoprefixer', object.createEmpty());
-			return generateCode();
-		});
+		if (prettierInstalled) sv.devDependency('prettier-plugin-tailwindcss', '^0.6.11');
 
 		sv.file('src/app.css', (content) => {
-			const layerImports = ['base', 'components', 'utilities'].map(
-				(layer) => `tailwindcss/${layer}`
-			);
+			const layerImports = ['tailwindcss'];
```
benmccann, line 22: you could probably simplify this section a bit to something like:
```
			if (content.includes('tailwindcss')) {
				return content;
			}

			const { ast, generateCode } = parseCss(content);
			const originalFirst = ast.first;

			const nodes = addImports(ast, ["'tailwindcss'"]);
```

packages/addons/tailwindcss/index.ts
```diff
@@ -1,121 +1,25 @@
...
-			object.property(themeObject, 'extend', object.createEmpty());
-
-			const pluginsArray = object.property(config, 'plugins', array.createEmpty());
-
-			for (const plugin of plugins) {
-				if (!options.plugins.includes(plugin.id)) continue;
-				imports.addDefault(ast, plugin.package, plugin.identifier);
-				array.push(pluginsArray, { type: 'Identifier', name: plugin.identifier });
-			}
+		sv.devDependency('@tailwindcss/vite', '^4.0.0');
```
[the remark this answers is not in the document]
jycouet: Thank you, should be better on this.

_(I'm fighting with my machine ATM 😒)_
Windows ARM WSL => I can't get `playwright` to work, some host (ubuntu) lib missing.
Windows ARM => vitest hang after runing tests when not completed.

endigo9740:
@benmccann @jycouet

Hey guys, this is Chris from [Skeleton](https://www.skeleton.dev/). We're super excited about the Tailwind v4 update as well, but we're preparing to launch our [next major release](https://next.skeleton.dev/) as soon as next week. As such, we have not had a chance to even touch Tailwind v4 yet to ensure Skeleton is compatible.

We're using `sv` as part of our [onboarding for SvelteKit projects](https://next.skeleton.dev/docs/get-started/installation/sveltekit) and we're defaulting to the Tailwind (v3) install by default. So if it's not too much trouble, would it be possible to add a prompt and/or flag to ensure folks can set which version is installed? Currently we're recommending the following in our guide:

```
npx sv create --template minimal --types ts my-skeleton-app
```

A flag to default to Tailwind v3 temporary would no doubt save us from a flood of support tickets!

Once we can confirm all is well with v4 then by all means, let's open that floodgate! But if it goes like I'm expecting, we're in for a heavy lift before it'll work 1:1. Namely we may have to rewrite our entire Tailwind plugin, which may even become the focus of our following major release. The changes are not trivial in this regard!

We're down to collaborate on whatever is required to make this happen. Just let me know. Thanks!

benmccann:
Could you just use a pinned version of `sv` like `npx sv@0.6.16 create ...`?

endigo9740:
@benmccann as long as users would get the latest packages for Svelte/SvelteKit that would probably be fine for now.

What I'm worried about is we might get in there and realize we have to completely replace our entire Tailwind Plugin to resolve compatibility. That's the worst case scenario and would likely have a long lead time.

But I'll have a definitive answer on compatibility tomorrow. I'm going to run through and test our RC version with Tailwind 4 and see how it goes. Ideally it just works much ado about nothing! *fingers crossed*

jycouet:
Pining sv sounds to be the right solution to me. Plus it's adding a motivation boost 😅💪

endigo9740:
> Pining sv sounds to be the right solution to me. Plus it's adding a motivation boost 😅💪

Ha, well this is the end of a 13 month dev cycle for our new update. It's been a massive undertaking. Not sure if I need a motivation, a vacation, or a straight jacket at this point! 😄

jycouet:
While svelte 5 was a big effort, I know that for libs it's also massive ! Congratz' 🚀🚀

jycouet:
Thank you for all these details & info.

I have to say that I started this PR more like _"Haaa cool, v4 is out, for new projects it could make sense to be on latest"_.
I understand the issue with `skeleton` and every other UI lib based on tailwind.

In your doc, if you pin `sv`, yes you will get drizzle and co of this version too. But I don't know how often all these gets updated ? In the end it's maybe not a big deal ? For maybe few weeks ?

In the end, I let repo owners guide me in giving the direction.

endigo9740:
Completely understand Jean, and I am terribly sorry if I have derailed any progress here. There's obviously plenty of folks that will want to use the latest and greatest version of Tailwind asap. I just wanted to make sure you guys were aware of the impact it could have on us.

Mostly this is just bad timing! We've been in crunch mode these last few weeks to polish off this big new update, so we completely missed the Tailwind v4 beta. That's 100% on me. But with a bit more time we'll get to the point we can fully support Tailwind v4. Which Im incredibly excited for. There's so many awesome features and many things will be much simpler for both us and end users.

I think we just need more hours in the day to get everything done! Heh

endigo9740:
I'll say this - go ahead and merge whenever you guys are read. We'll opt for pinning `sv` and then figure out what that means for our users in the meantime. It could be as simple as a message saying like:

```
Skeleton currently only supports Tailwind v3, so we pin `sv` to X version to ensure compatibility. This may mean some add-ons are are affected. Make sure to update those directly after setting up your new project.
```

jycouet:
@benmccann did you get the chance to see my updates ?
Let me know if you have some other comment.

packages/addons/tailwindcss/index.ts
```diff
@@ -1,130 +1,50 @@
...
-		version: '^0.5.10',
-		identifier: 'forms'
-	},
-	{
-		id: 'container-queries',
-		package: '@tailwindcss/container-queries',
-		version: '^0.1.1',
-		identifier: 'containerQueries'
-	}
-];
```
AdrianGonz97, line 33: amongst these 3 plugins, I believe `container-queries` is the only one that's now integrated in tailwind v4. we can keep the other two and add their plugin references to the stylesheet like so:
```css
@plugin "@tailwindcss/typography";
@plugin "@tailwindcss/forms";
```

.changeset/empty-geese-draw.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sveltejs/addons': patch
+---
+
+using tailwindcss 4.0.0
```
AdrianGonz97, line 5:
```suggestion
---
'sv': minor
---

feat: update to `tailwindcss` v4.0.0
```

AdrianGonz97 (changes requested):
Only have a couple of suggestions. Aside from those, the rest is excellent and should be good to go. Thank you!!

packages/addons/tailwindcss/index.ts
endigo9740: Hey @AdrianGonz97 I've been digging through the Tailwind v4 docs the last few days. There's no mention of plugins in the v4 docs as far as I can tell. They really aren't pushing them any more. My guess is more is (or will) be absorbed into core.

They confirmed container queries are now core:
https://tailwindcss.com/blog/tailwindcss-v4#container-queries
> We've brought container query support into core for v4.0, so you don't need the @tailwindcss/container-queries plugin anymore:

And at least from initial testing, form styles seemed improved by default, so those may have been added to the global reset. But not seeing anything official.

Typography I've not had a chance to test at all though.

benmccann:
I filed an issue to ask them about plugins. Hopefully they'll be able to share some information. https://github.com/tailwindlabs/tailwindcss-typography/issues/372

packages/addons/tailwindcss/index.ts
jycouet: In the v4 doc I saw: https://tailwindcss.com/docs/functions-and-directives#plugin-directive
image

And it's maybe not the best to bring in `sv` an option to have "legacy" included? (even in option?)

But, open to change my mind ✌️
Let me know

.changeset/empty-geese-draw.md
```diff
@@ -0,0 +1,5 @@
+---
+'sv': minor
```
manuel3108, line 2:
```suggestion
'sv': patch
```

we are still in pre 1.0 mode, so feats are handled as patch udpates

manuel3108:
Thank you everyone for all your work! This looks perfect, and CI is passing!

Apart from the discussion around `@tailwindcss/typography` and `@tailwindcss/forms` I don't see anything that's preventing us from merging this. 

Let's give them a few days so that we can hopefully get a response on the status of those two plugins. If we don't get a meaningful response I think using `tailwindcss@4` would outweigh the benefits of being able to add two plugins directly from the cli. Or am I missing something?

jycouet:
I think that you are not missing anything 👍

TGlide:
Just wanna say y'all are awesome. :heart:

packages/addons/tailwindcss/index.ts
manuel3108: > This is likely the most official source at the moment:
> [tailwindlabs/tailwindcss#15715 (comment)](https://github.com/tailwindlabs/tailwindcss/discussions/15715#discussioncomment-11946588)

Thanks for linking that discussion. I added my two cents a few days ago. It looks like there is currently no way to include these plugins without using the "legcy" api mentioned above. Let's give them a few more days.

Currently though I would be leaning towards keeping the two plugins inside the add-on and linking them based on the docs provided above.

jycouet: If everything would work with the `legacy` mode why not. But it seems that it's not fully compatible.
Looking at https://github.com/tailwindlabs/tailwindcss-typography/issues/372 issue.

manuel3108: Maybe I'm reading some of the comments wrongs in that issue, but apart from the missing documentation i'm unable to read from the comments that anyone using the `@plugin` api has problems with anything.

sujyotraut:
Quick question: does it work with `prettier-plugin-tailwindcss`? [prettier-plugin-tailwindcss docs](https://github.com/tailwindlabs/prettier-plugin-tailwindcss?tab=readme-ov-file#specifying-your-tailwind-stylesheet-path).
> When using Tailwind CSS v4 you must specify your CSS file entry point, which includes your theme, custom utilities, and other Tailwind configuration options. To do this, use the tailwindStylesheet option in your Prettier configuration.
```json
{
  "tailwindStylesheet": "./src/app.css"
}
```

packages/addons/tailwindcss/index.ts
jycouet: You need to use back a `js` config to be able to tune it.
I mean, at this point, typography is not "native" v4. _(they are working on it)_

When I look at `sv`, I think: "Put me in the right track for x, y, z". 
I don't expect to start with a legacy way for doing things. If people needs `typography` and v4, they can do what on their own (it's userland).
When typography will be back to "native v4", yes I would be good to be able to add it directly from sv cli (like today with v3)

manuel3108:
Based on the current state of plugins, and no official timeline provided, let's go with what we have here. As soon as the plugins are officially supported again, we will re-include them. 

Thanks for all your work @jycouet and all the others for providing feedback!

packages/addons/tailwindcss/index.ts
manuel3108: See https://github.com/sveltejs/cli/pull/422#issuecomment-2642758906

hyunbinseo:
As @sujyotraut mentioned, is there a reason why this is not added?

> When using Tailwind CSS v4 you must specify your CSS file entry point

```jsonc
// .prettierrc
{
  "tailwindStylesheet": "./src/app.css"
}
```

```
┌  Welcome to the Svelte CLI! (v0.7.0)
│
◇  What would you like to add to your project? (use arrow keys / space bar)
│  prettier, tailwindcss
```

manuel3108:
Must have missed that, https://github.com/sveltejs/cli/issues/507

PR #434, opened 2025-02-02
fix: changsets ignores unpublished packages

### LINKED ISSUE #403: Make `changesets` cli suggest correct packages
> We probably need to change the package in the changeset. I wonder if there's a way we can make the changeset CLI do the correct thing but default. Otherwise perhaps we should mention it in the PR template and contributors guide

Initially posted by @benmccann in https://github.com/sveltejs/cli/pull/400#pullrequestreview-2556086389

manuel3108:
closes #403

### DIFF
--- a/.changeset/config.json
+++ b/.changeset/config.json
@@ -7,5 +7,5 @@
 	"access": "public",
 	"baseBranch": "main",
 	"updateInternalDependencies": "patch",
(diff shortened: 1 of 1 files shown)


benmccann:
How is this different? Was the `!` syntax not working?

manuel3108:
exactly

PR #456, opened 2025-02-26
fix: allow `tailwindcss` plugins to be added after initial execution

AdrianGonz97:
Fixes a bug where tailwind plugins wouldn't be added on subsequent `sv add tailwindcss` runs.


manuel3108 (approved):
Good catch, works as expected now!

PR #459, opened 2025-02-27
fix: pass `schema` to `drizzle` client for better type generation

### LINKED ISSUE #452: Drizzle option generates code that did not work for me.
New to SK, Drizzle, and my TS is weak too.

So I generated a new SK app with `npx sv create whatever` and it generated this code in `src\lib\server\db\index.ts`:

```ts
import { drizzle } from 'drizzle-orm/postgres-js';
import postgres from 'postgres';
import { env } from '$env/dynamic/private';
if (!env.DATABASE_URL) throw new Error('DATABASE_URL is not set');
const client = postgres(env.DATABASE_URL);
export const db = drizzle(client);
```

I was unable to use `db` at all without errors IE: `await db.query.user.findMany();` would throw `TypeError: Cannot read properties of undefined (reading 'findMany')`.

I needed to use the following code... where I pass schema to drizzle.

```ts
import { drizzle } from 'drizzle-orm/postgres-js';
import postgres from 'postgres';
import { env } from '$env/dynamic/private';
import * as schema from './schema';

if (!env.DATABASE_URL) throw new Error('DATABASE_URL is not set');
const client = postgres(env.DATABASE_URL);
export const db = drizzle(client, {
    schema
});
```

Hope this helps someone...

manuel3108:
closes #452 

This PR basically changes `src\lib\server\db\index.ts` from 
```js
export const db = drizzle(client);
```

to 
```js
export const db = drizzle(client, {
    schema
});
```

For both mysql based  options (mysql2 and planetscale) another property [`mode`](https://orm.drizzle.team/docs/rqb#modes) is added to the config object. 

I' was unable to test the planetscale implementation though, as their pricing is off the charts.  We should consider dropping it in the future.

### DIFF
--- a/.changeset/swift-poets-wonder.md
+++ b/.changeset/swift-poets-wonder.md
@@ -1,0 +1,5 @@
+---
+'sv': patch
+---
+
+fix: pass `schema` to `drizzle` client for better type generation
--- a/packages/addons/drizzle/index.ts
+++ b/packages/addons/drizzle/index.ts
@@ -261,6 +261,7 @@ export default defineAddon({
 			const { ast, generateCode } = parseScript(content);
 
 			imports.addNamed(ast, '$env/dynamic/private', { env: 'env' });
+			imports.addNamespace(ast, './schema', 'schema');
 
 			// env var checks
 			const dbURLCheck = common.statementFromString(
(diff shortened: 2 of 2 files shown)


manuel3108:
Updated PR description to reflect our discussion yesterday.

PR #461, opened 2025-02-28
update to paraglide js v2

samuelstroschein:
Paraglide JS 2.0 is going to be released on Monday (March 3). This PR updates the addon to set up and use v2 of Paraglide JS. 

Pre-elimnary release notes for Paraglide JS v2 can be found [here]( https://github.com/opral/monorepo/blob/4e1a26dbeef4779bd36629a72b48c6477a7fb104/inlang/packages/paraglide/paraglide-js/v2-release-notes.md)

**Don't merge this before v2 got released (I will write a comment when this happened. Approx second week of march)**


packages/addons/paraglide/index.ts
```diff
@@ -17,24 +17,19 @@ import { addToDemoPage } from '../common.ts';
 const DEFAULT_INLANG_PROJECT = {
 	$schema: 'https://inlang.com/schema/project-settings',
 	modules: [
-		'https://cdn.jsdelivr.net/npm/@inlang/message-lint-rule-empty-pattern@1/dist/index.js',
-		'https://cdn.jsdelivr.net/npm/@inlang/message-lint-rule-identical-pattern@1/dist/index.js',
-		'https://cdn.jsdelivr.net/npm/@inlang/message-lint-rule-missing-translation@1/dist/index.js',
-		'https://cdn.jsdelivr.net/npm/@inlang/message-lint-rule-without-source@1/dist/index.js',
-		'https://cdn.jsdelivr.net/npm/@inlang/message-lint-rule-valid-js-identifier@1/dist/index.js',
```
samuelstroschein, line 24: Lint rules have been deprecated in favor of upcoming lix validation rules that are more generalizable https://github.com/opral/lix-sdk/issues/239

packages/addons/paraglide/index.ts
```diff
@@ -17,24 +17,19 @@ import { addToDemoPage } from '../common.ts';
...
+		pathPattern: './messages/{locale}.json'
 	}
 };
 
 const options = defineAddonOptions({
 	availableLanguageTags: {
 		question: `Which languages would you like to support? ${colors.gray('(e.g. en,de-ch)')}`,
 		type: 'string',
-		default: 'en',
+		default: 'en, de',
```
samuelstroschein, line 32: added a second language to have a better demo. i skipped through "which languages do you want to support?" several times. the demo will only have one language as a result leading to a "okay and now?" moment

packages/addons/paraglide/index.ts
```diff
@@ -62,18 +57,18 @@ const options = defineAddonOptions({
...
 		if (!kit) unsupported('Requires SvelteKit');
 	},
-	run: ({ sv, options, typescript, kit, dependencyVersion }) => {
+	run: ({ sv, options, typescript, kit }) => {
 		const ext = typescript ? 'ts' : 'js';
 		if (!kit) throw new Error('SvelteKit is required');
 
 		const paraglideOutDir = 'src/lib/paraglide';
 
-		sv.dependency('@inlang/paraglide-sveltekit', '^0.15.5');
```
samuelstroschein, line 76: the sveltekit adapter is not needed anymore. hallelujah

packages/addons/paraglide/index.ts
```diff
@@ -114,31 +109,17 @@ export default defineAddon({
...
-			const i18n = variables.declaration(ast, 'const', 'i18n', createI18nExpression);
-
-			const existingExport = exports.namedExport(ast, 'i18n', i18n);
-			if (existingExport.declaration !== i18n) {
-				log.warn('Setting up $lib/i18n failed because it already exports an i18n function');
-			}
-
-			return generateCode();
-		});
-
```
samuelstroschein, line 134: no more i18n.js needed in v2 🙏

packages/addons/paraglide/index.ts
```diff
@@ -153,41 +134,20 @@ export default defineAddon({
...
-			const templateCode = new MagicString(template.generateCode());
-			if (!templateCode.original.includes('<ParaglideJS')) {
-				templateCode.indent();
-				templateCode.prepend('<ParaglideJS {i18n}>\n');
-				templateCode.append('\n</ParaglideJS>');
-			}
-
-			return generateCode({ script: script.generateCode(), template: templateCode.toString() });
-		});
-
```
samuelstroschein, line 190: no more provider and layout needed in v2 🙏

packages/addons/paraglide/index.ts
```diff
@@ -204,8 +164,7 @@ export default defineAddon({
 			}
 			htmlNode.attribs = {
 				...htmlNode.attribs,
-				lang: '%paraglide.lang%',
-				dir: '%paraglide.textDirection%'
```
samuelstroschein, line 208: removing textDir because users reported issues with the [Intl API](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Locale/getTextInfo) behaving differently in different runtimes. 

it's better to have a working setup for everyone that users can improve themselves than a half broken setup

packages/addons/paraglide/index.ts
```diff
@@ -229,36 +188,14 @@ export default defineAddon({
...
-						${ts('', '/**')} 
-						${ts('', '* @param {import("$lib/paraglide/runtime").AvailableLanguageTag} newLanguage')} 
-						${ts('', '*/')} 
-						function switchToLanguage(newLanguage${ts(': AvailableLanguageTag')}) {
-							const canonicalPath = i18n.route(page.url.pathname);
-							const localisedPath = i18n.resolveRoute(canonicalPath, newLanguage);
-							goto(localisedPath);
-						}
-					`);
-				}
```
samuelstroschein, line 261: `setLocale()` just works in v2, irrespective of the chosen strategy

packages/addons/paraglide/index.ts
```diff
@@ -271,11 +208,15 @@ export default defineAddon({
...
-							`${templateCode.getIndentString()}<button onclick={() => switchToLanguage('${x}')}>${x}</button>`
+							`${templateCode.getIndentString()}<button onclick={() => setLocale('${x}')}>${x}</button>`
 					)
 					.join('\n');
 				templateCode.append(`<div>\n${links}\n</div>`);
 
+				templateCode.append(
+					'<p>\nIf you use VSCode, install the <a href="https://marketplace.visualstudio.com/items?itemName=inlang.vs-code-extension" target="_blank">Sherlock i18n extension</a> for a better i18n experience.\n</p>'
+				);
+
```
samuelstroschein, line 219: Moved the recommendation form the CLI into the paraglide js demo with a link to directly open the vscode marketplace

.changeset/eighty-parrots-grab.md
```diff
@@ -0,0 +1,5 @@
+---
+'sv': minor
```
manuel3108, line 2:
```suggestion
'sv': patch
```

we are still below 1.0

packages/addons/paraglide/index.ts
manuel3108: Agree, this makes sense. Second language was removed in #130 for whatever reason. Maybe we should use `es` instead, as spanish is one of the most spoken languages.

manuel3108 (approved):
This is so cool. So much less boilerplate required! Nice work.

Also thanks for the PR, everything seems to be working perfectly at a first glance. Apart from the tests of course, these are failing because they are using a non existing npm version of paraglide

samuelstroschein:
Ready to merge eta next week. Getting a few edge case Paraglide JS 2.0 bug reports in. Fixing them over the weekend.

packages/addons/paraglide/index.ts
manuel3108:
```suggestion
		default: 'en, es',
```

manuel3108:
Sure, just give me a heads up once it's ready!

samuelstroschein:
@manuel3108 press merge 🥳

manuel3108:
Mhmm, seems like the tests are still failing. Don't have the time to investigate this today.

manuel3108:
I still don't understand why the tests are failing here. Everything works just fine locally.

samuelstroschein:
Argh the gnarly cloud without possibilities to test locally is striking once again. @manuel3108 i remember that CI was green when i first opened it (ha, that doesn't help, does it?). 

lmk if i can do something

manuel3108:
> remember that CI was green when i first opened it 
i think this was because at that time it did not run (iirc), as this seems to be your first contribution to your project, and we need to "whitelist" each run

manuel3108:
Ok, i think I finally found the error. The message displayed in the github action has to be converted to a buffer, than a string. This is what it was printing:

> Scope: all 3 workspace projects
..                                       | Progress: resolved 1, reused 0, downloaded 0, added 0
/home/runner/work/cli/cli/.test-output/addons/paraglide/-1511660393_0:
 ERR_PNPM_UNSUPPORTED_ENGINE  Unsupported environment (bad pnpm and/or Node.js version)
>
> This error happened while installing the dependencies of @inlang/paraglide-js@2.0.1
 at @inlang/sdk@2.4.3
>
>Your Node version is incompatible with "@lix-js/sdk@0.4.3".
>
>Expected version: >=21
>Got: v18.20.7
>
>This is happening because the package's manifest has an engines.node field specified.
>To fix this issue, install the required Node version.

Locally i was already using a valid node version.
Update the node version to test against, and everything should be working now!

samuelstroschein:
@manuel3108 i am gonna invest 5 min if i can ship https://github.com/opral/lix-sdk/issues/277

manuel3108:
That would be great!
Because looks like v21 is unsupported by v-p-s

> Scope: all 5 workspace projects
 ERR_PNPM_UNSUPPORTED_ENGINE  Unsupported environment (bad pnpm and/or Node.js version)
>
>Your Node version is incompatible with "@sveltejs/vite-plugin-svelte@5.0.3(svelte@5.25.2)(vite@6.2.2)".
>
>Expected version: ^18.0.0 || ^20.0.0 || >=22
>Got: v21.7.3
>
>This is happening because the package's manifest has an engines.node field specified.
>To fix this issue, install the required Node version.

samuelstroschein:
@manuel3108 paraglide js 2.0.4 should work. try again in approx 10 mins when https://github.com/opral/monorepo/pull/3530 is deployed

manuel3108:
Everything is passing now :tada:. Thanks for your help!

PR #465, opened 2025-03-02
chore: detect package manager asynchronously

benmccann:


manuel3108:
I don't fully understand why this change is necessary. Previously we tried to keep as much code synchronously where possible (at least as I remember). Based on the [release notes](https://github.com/antfu-collective/package-manager-detector/releases) of `package-manager-detector` it does not look like the sync methods  are being deprecated or something.

But i'm probably missing something here.

benmccann:
Yeah, sorry, I should have provided a description. It's not necessary at the moment, but we are discussion removing the synchronous APIs in `package-manager-detector`, so that we don't have to maintain two versions of every method. I wanted to make sure that it wouldn't affect us much and would be easy enough to adjust to if we do so

benmccann:
There's a PR out now to drop CJS and sync APIs from `package-manager-detector`, so it'd probably make sense to move forward with this PR

PR #476, opened 2025-03-08
fix: add `playwright` setup command to download browsers

### LINKED ISSUE #316: When installing Playwright from the SvelteKit installer, it should offer to install the browsers
### Describe the problem

When I tried to run Playwright tests, they failed because the browsers were not installed. As Playwright isn't much use without them, it would be useful for the installer to run `npx playwright install` after installing Playwright. 

### Describe the proposed solution

If the user opts to install Playwright, the next question should be whether or not to install all the Playwright browsers. The default should be "yes" — if the user only wants to install selected browsers, they can choose "no" and then install the browsers manually.

### Alternatives considered

Typing `npx playwright install` after trying to run tests and having them fail to find a browser.

### Importance

nice to have

### Additional Information

Usually, you're not in a state to run `npx playwright test` until you're a fair way into the project, at which point you've forgotten that it didn't install the browsers at the same time as Playwright.

As Playwright installs its browsers per-user, not per-project (in ~/Library/Caches/ms-playwright on macOS) this doesn't affect anyone who's already installed the browsers as part of a previous project. I suspect most of the SvelteKit devs fall into this camp, which might be why nobody's noticed this 😁

manuel3108:
Closes #316


benmccann:
I'm not sure this is really helpful. When you try to run playwright the first time it will give you the install command

AdrianGonz97:
Agreed. I don't think we really want to pollute a user's `package.json` with scripts that they'll only be using very rarely.

manuel3108:
Sorry for not putting this into the description, i was initially thinking of only adding the `playwright install` command to the next steps as requested by #316. In order to do this correctly this will require a reference to `package-manager-detector` which is currently not accessible by the add-ons (although easily doable).

That's why i looked around for the other add-ons and saw that we had similar things for `drizzle-kit push`, `drizzle-kit migrate` and `drizzle-kit studio` so i thought this would align quite well.

Would you be open if it would only be a text in the next steps section?

PR #478, opened 2025-03-08
fix: insert the `tailwindcss` vite plugin at the start of the plugin array

Hugos68:
`tailwindcss` should be first in the `vite` plugin array because they both `enforce: "pre"`. The reason for this is because the tailwind class scanner should run before any transformations are done to the source files.


AdrianGonz97 (approved):

manuel3108 (approved):
Thanks!

PR #483, opened 2025-03-12
chore: disable `no-undef` ESLint rule in TypeScript project

baseballyama:
close: https://github.com/sveltejs/cli/issues/481

### DIFF
--- a/packages/addons/eslint/index.ts
+++ b/packages/addons/eslint/index.ts
@@ -85,11 +85,25 @@ export default defineAddon({
 			const globalsNode = common.createSpreadElement(common.expressionFromString('globals.node'));
 			const globalsObjLiteral = object.createEmpty();
 			globalsObjLiteral.properties = [globalsBrowser, globalsNode];
+			const off = common.createLiteral('off');
+			off.comments = [
+				{
+					type: 'Block',
+					value:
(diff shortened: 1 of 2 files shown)


manuel3108 (changes requested):
Maybe I'm missing something, but in #481 you are writing about `no-undef`, and your PR seems to mix and match `no-undef` and `no-unused-vars` randomly.

More importantly, I deleted the `eslint.config.js` in the repro provided and run `pnpm dlx https://pkg.pr.new/sveltejs/cli/sv@483` which still revealed the original error.

I'm assuming your PR was meant for `no-undef`?

baseballyama:
@manuel3108 Sorry, I got confused for some reason. Fixed now!

packages/addons/eslint/index.ts
```diff
@@ -85,11 +85,25 @@ export default defineAddon({
 			const globalsNode = common.createSpreadElement(common.expressionFromString('globals.node'));
 			const globalsObjLiteral = object.createEmpty();
 			globalsObjLiteral.properties = [globalsBrowser, globalsNode];
+			const off = common.createLiteral('off');
+			off.comments = [
```
manuel3108, line 89: Without running prettier, this generates the following code:
image

The comment probably needs to be added to the `rules` variable or something like this

manuel3108 (changes requested):
Apart from the one thing, everything seems to be working as expected now.

baseballyama:
@manuel3108 

I think the format is much better now. The only remaining issue is the indentation, but I this we can split the PR to fix it.

manuel3108 (approved):
Thats perfect now, thanks!

Indentation should be fixed by #380 in theory, so I don't think we need to bother about this now, as you said!

PR #488, opened 2025-03-15
chore: re-enable preview releases

manuel3108:
Delete the script, as it has not been used since https://github.com/sveltejs/cli/pull/15, which is way over half a year ago


manuel3108:
Self merged as https://github.com/tj-actions/changed-files is not accessible anymore

Issue #368, opened 2024-12-20
`npx sv migrate app-state` only migrates some files


benmccann:
Run it in the `web` directory of https://github.com/immich-app/immich and it won't migrate anything (some stuff was already migrated in https://github.com/immich-app/immich/pull/14807)

manuel3108:
Here is a failing test (note the word `page` in an unrelated import). I tried fixing the bug but failed.

```js
test('Updates $app/store #4', () => {
	const result = transform_svelte_code(
		`<script>
	import { page } from '$app/stores';
	import Comp from '$lib/page/Comp.svelte'
</script>

<div>{$page.url}</div>
<button onclick={() => {
	console.log($page.state);
}}></button>
`
	);
	assert.equal(
		result,
		`<script>
	import { page } from '$app/state';
	import Comp from '$lib/page/Comp.svelte'
</script>

<div>{page.url}</div>
<button onclick={() => {
	console.log(page.state);
}}></button>
`
	);
});
```

FYI @dummdidumm

ranjan-purbey:
I think it fails due to this line:
https://github.com/sveltejs/cli/blob/e325bb2c71fe3fefce4cfcb07e3ab33870eddb45/packages/migrate/migrations/app-state/migrate.js#L70

Here we return early and skip the migration if the imported store alias is not prefixed with a `$`, unless a few conditions are met. We probably need to add another condition here (while checking for false positives) to check if the alias is part of a string or a comment.

For e.g., if a component has the following line, it won't be migrated according to the current logic:
```javascript
const str = "before page after";
```
or
```javascript
// this is a page
```

garikAsplund:
One thing I noticed when working on #494 was that if I ran the migration on a newly created project it would work great. However, when I changed an `$app/state` back to `$app/stores` after the initial migration, subsequent runs wouldn't fix it. 

I don't know if that helps pinpoint any issues here or just muddies the waters.

manuel3108:
That's because the word `page` will appear already, and that messes with the script somehow. This is basically also the testcase i mentioned above.

PR #494, opened 2025-03-20
feat: added `app-state` migration into `svelte-5` migration

### LINKED ISSUE #485: Automatically run app-state migration as part of svelte-5 migration

garikAsplund:
Fixes #485 

This is the simplest approach I could think of--I know it can get a little hairy with all the versions of Svelte and SvelteKit out there. Still, I believe most of the logic is already in place with the CLI to make sure the user has the right versioning heading into the migration.

This only runs the app-state migration script if the user has SvelteKit. It doesn't really log anything or add any error handling--I'll leave those up to you if desired.

Also, I know you guys have a contributing guide on the docket, but I was wondering how exactly you test this CLI on your own projects. I ask because I ended up using `npm link` which is pretty standard, but I had to edit [these lines of code](https://github.com/sveltejs/cli/blob/d1f4432c0e31677ae754d50394547eeff820ddfe/packages/cli/commands/migrate.ts#L28C3-L34C79) to make sure that when I tested my fixes locally that it linked to my scripts instead of pulling `@latest` from `npx` 🙏

### DIFF
--- a/.changeset/strange-islands-float.md
+++ b/.changeset/strange-islands-float.md
@@ -1,0 +1,5 @@
+---
+"svelte-migrate": minor
+---
+
+feat: added `app-state` migration into `svelte-5` migration
--- a/packages/migrate/migrations/svelte-5/index.js
+++ b/packages/migrate/migrations/svelte-5/index.js
@@ -17,7 +17,7 @@ import {
 } from '../../utils.js';
 import { migrate as migrate_svelte_4 } from '../svelte-4/index.js';
 import { migrate as migrate_sveltekit_2 } from '../sveltekit-2/index.js';
-import { migrate as migrate_app_state } from '../app-state/index.js';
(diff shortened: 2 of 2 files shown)


manuel3108:
You can actually just run `node ./packages/migrate/bin.js` to just execute the migrate cli. The script you mentioned is basically doing the same thing. Really need to get that contributing guide out.

Thanks for the PR btw!
I think this is the correct way to go forward, and I don't think a prompt or log is personally necessary. But i would love to get @benmccann feedback on this.

Bu I don't think we should merge this as long as #368 is still open, looks like it's only finding some of the occurrences and not all.

benmccann:
I think this is a good PR. I don't know if we need to block on that issue. It would be great to get it fixed, but migrating some stuff is better than migrating none.

manuel3108 (approved):
In that case, let's go!

Issue #187, opened 2024-10-20
Check various issues with `package-manager-detector`


benmccann:
A few things I want to remember to check:
- use its version of `getUserAgent`
- handling of monorepo vs not
- review how/where packageManager field is used and the user agent vs file system lock concepts

manuel3108:
IIRC, this has already been implemented.

benmccann:
I hadn't gotten around to this yet, but could have titled the issue better

manuel3108:
Maybe im still misunderstanding the issue, but shouldn't this be an issue in https://github.com/antfu-collective/package-manager-detector in that case?

manuel3108:
Still not sure what this issue is about

PR #495, opened 2025-03-20
fix: dynamically suggest package manager installation command in `migrate` scripts

garikAsplund:
Relates #187

Had to add package-manager-detector to the migrate package as a dependency. 

I was a little confused on using `detect` vs. `getUserAgent`--maybe you have a better understanding of when and where to use each? For this I used the barebones `detect` and it worked on my machine.

Screenshot 2025-03-20 at 9 25 12 AM
Screenshot 2025-03-20 at 9 11 29 AM


packages/migrate/migrations/app-state/index.js
```diff
@@ -94,8 +95,10 @@ export async function migrate() {
 	/** @type {(s: string) => string} */
 	const cyan = (s) => pc.bold(pc.cyan(s));
 
-	// TODO: use package-manager-detector here
-	const tasks = ["install the updated dependencies ('npm i' / 'pnpm i' / etc)"];
+	const detected = await detect({ cwd: process.cwd() });
+	const pm = detected?.name ?? 'npm';
+
+	const tasks = [`Install the updated dependencies by running ${cyan(`${pm} install`)}`];
```
manuel3108, line 101:
```suggestion
	const detected = await detect({ cwd: process.cwd() });
	const pm = detected?.name ?? 'npm';
	const cmd = /** @type {import('package-manager-detector').ResolvedCommand} */ (
		resolveCommand(pm, 'install', [])
	);

	const tasks = [
		`Install the updated dependencies by running ${cyan(`${cmd.command} ${cmd.args.join(' ')}`)}`
	];
```

Hardcoding the `install` command might not be correct in each case. I think for the install command it should work basically everywhere, but let's stay agnostic if that abstraction already exists.

Same would go for the other places.

manuel3108 (changes requested):
Thanks for your pr!

`detect` is determining the package manager based on existing or non existing lockfiles like `pnpm-lock.yml` etc. `getUserAgent` determines with which package manager the current command is being executed.

Since `svelte-migrate` is expected to be run on an existing project, it's totally valid to use `detect` here in my opinion.

manuel3108 (approved):

Issue #350, opened 2024-12-05
support detection of non .prettierrc config files


exectx:
When integrating Tailwind into an existing project that already has a prettier config file that isn't a `.prettierrc` , an incomplete `.prettierrc` file is generated.

The current implementation in [`prettier/index.ts`](https://github.com/sveltejs/cli/blob/main/packages/addons/prettier/index.ts) and [`tailwindcss/index.ts`](https://github.com/sveltejs/cli/blob/main/packages/addons/tailwindcss/index.ts) should detect the existing prettier config file and use `parseJson` or `parseScript` in case the prettier config is for instance `prettier.config.js`. 

repro: `npx sv add tailwind` on existing project with a `prettier.config.js`

Issue #285 is related but focuses mainly on monorepo setups.

manuel3108:
Based on all the different allowed configuration files (19 different ones) this is not going to be fun to implement. Only implementing them partially does not make sense at all, as users will follow-up and ask if we can support configuration file xyz.

Let's do a quick github code search to find out the relevance of the most common configs
* [`.prettierrc`](https://github.com/search?q=path%3A**%2F.prettierrc&type=code&ref=advsearch) 163k
* [`.prettierrc.json`](https://github.com/search?q=path%3A**%2F.prettierrc.json&type=code&ref=advsearch) 52k
* [`prettier.config.js`](https://github.com/search?q=path%3A**%2Fprettier.config.js&type=code&ref=advsearch) 22k
* [`prettier.config.ts`](https://github.com/search?q=path%3A**%2Fprettier.config.ts&type=code&ref=advsearch) 284

`.prettierrc` is by far the most popular one, and therefore i don't see any need to implement this.

PR #513, opened 2025-03-29
fix(prettier): always use JSON config

43081j:
If the project already has a `.prettierrc` which is YAML, we will currently throw an uncaught exception trying to parse it as JSON.

This switches to always using `.prettierrc.json`, so we can safely assume it should be valid JSON if it exists.

Alternatively, we could try/catch the parsing and overwrite if it fails. Or, more expensive, try JSON and then try YAML...

### DIFF
--- a/packages/addons/prettier/index.ts
+++ b/packages/addons/prettier/index.ts
@@ -23,7 +23,7 @@ export default defineAddon({
 			`;
 		});
 
-		sv.file('.prettierrc', (content) => {
+		sv.file('.prettierrc.json', (content) => {
 			const { data, generateCode } = parseJson(content);
 			if (Object.keys(data).length === 0) {
 				// we'll only set these defaults if there is no pre-existing config
--- a/packages/addons/tailwindcss/index.ts
+++ b/packages/addons/tailwindcss/index.ts
@@ -132,7 +132,7 @@ export default defineAddon({
 		}
 
 		if (dependencyVersion('prettier')) {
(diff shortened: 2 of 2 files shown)


manuel3108:
Thanks for your PR!

Referring to this comment https://github.com/sveltejs/cli/issues/350#issuecomment-2674845755 I don't think this is actually a problem. Searching github for `.prettierrc` files that contain yml I only get [~ 900 results](https://github.com/search?q=path%3A**%2F.prettierrc+%22-+%22&type=code&ref=advsearch)

And based on the [prettier docs](https://prettier.io/docs/configuration) the files are ordered based on preference, and `.prettierrc` is ordered above `.prettierrc.json`

43081j:
>  I don't think this is actually a problem

it's a problem because the CLI will crash if i have a `prettierrc` containing YAML. it doesn't really matter how many OSS projects you can find which have such a file

we should at the very least handle the exception instead of leaving it uncaught.

this isn't really a closed issue. you don't need to switch to `prettierrc.json` but you do need to fix the bug, otherwise we're leaving an easy-to-fix uncaught exception just lying around.

benmccann:
Thanks James! Dealing with the exception gracefully sounds like a good plan to me. I'd welcome such a PR

manuel3108:
Missed that part in the PR description, totally agree. If you need any help with getting started, hit me up!

43081j:
No worries. Sorry for the confusion

I'll open another pr 👍

PR #522, opened 2025-04-03
security: upgrade vite to avoid CVE-2025-31486

manuel3108:


hyunbinseo:
If this was merged, shouldn't [CVE-2025-32395](https://github.com/advisories/GHSA-356w-63v5-8wf4) be patched as well?

PR #523, opened 2025-04-06
chore: update dependencies

manuel3108:


benmccann:
Personally I would just always set the minor and patch version to 0 to reduce churn and make these PRs easier to review. Bumping the minor and patch version has no effect on the user's project

manuel3108:
I was hoping to fix the ~7 Dependabot warnings in this repo. But you are right, the only affected runtime dependency is for `tar-fs`, which is only used for community adders, which are unsupported at the moment, so the relevant code will never be executed. 

Removed the changset

AdrianGonz97:
it seems like `vitest@3.1.X` isn't working well with out test suite in windows. I've updated it to the latest working patch version, `3.0.9`. we can resolve this later, but will merge this PR for now

AdrianGonz97 (approved):

PR #524, opened 2025-04-06
chore: update addon dependencies

manuel3108:


benmccann:
Personally I would just always set the minor and patch version to 0 to reduce churn and make these PRs easier to review. Bumping the minor and patch version has no effect on the user's project

manuel3108:
Always forgot about this, thanks for the heads up!

PR #526, opened 2025-04-06
chore: unfork clack

manuel3108:
Relates: https://github.com/bombshell-dev/clack/pull/276

Thanks to @43081j for making this possible by submitting many PRs to the clack library.

This is WIP because
* [x] we currently rely on PR-build
* [x] `taskLog` is failing
* [x] a new `format` option will be provided to `note` so that we can get our current behavior back


43081j:
We also need an equivalent for `cancel`

I think I'll implement that as an option to `outro` in clack instead, i.e so you can set the level (success, warn, error)

manuel3108:
Yeah that sounds logical. Although at least typing-wise vscode is not complaining

manuel3108:
Looks like both relevant PR for clack have been merged, checking this

43081j:
keep in mind we should still merge this first:

bombshell-dev/clack#251

so you have a real release instead of a pkg.pr.new one

ill see if i can sort that out this weekend

manuel3108:
Thats just what i was wondering about :D was searching for a non existing release. It's not that this has any time constraints, just wanted to clean up this PR because i nearly forgot about it

AdrianGonz97 (approved):
This is awesome! Thank you @43081j for all your work in getting this through!!

manuel3108:
Uhhh, nice trick!

43081j:
awesome work :D 

if we push any other useful stuff upstream ill let you know too

PR #527, opened 2025-04-07
fix(prettier): skip non-JSON prettierrc config

43081j:
If the project already has a `.prettierrc` which is YAML, we will currently throw an uncaught exception trying to parse it as JSON.

This change means we will skip such files and only try to mutate JSON configs.

### DIFF
--- a/packages/addons/prettier/index.ts
+++ b/packages/addons/prettier/index.ts
@@ -29,7 +29,7 @@ export default defineAddon({
 				({ data, generateCode } = parseJson(content));
 			} catch {
 				log.warn(
-					'A `.prettierrc` config already exists and cannot be parsed. Skipping initialization.'
+					'A `.prettierrc` config already exists and cannot be parsed as JSON. Skipping initialization.'
 				);
 				return content;
 			}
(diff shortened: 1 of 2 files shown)


43081j:
cc @benmccann @manuel3108

if you can word the warning message any better, let me know or feel free to push to this branch

basically, if you have an existing `.prettierrc` which is _not_ JSON, we will warn and skip updating it.

packages/addons/prettier/index.ts
```diff
@@ -24,7 +24,15 @@ export default defineAddon({
 		});
 
 		sv.file('.prettierrc', (content) => {
-			const { data, generateCode } = parseJson(content);
+			let data, generateCode;
+			try {
+				({ data, generateCode } = parseJson(content));
+			} catch {
+				log.warn(
+					'A `.prettierrc` config already exists and cannot be parsed. Skipping initialization.'
```
benmccann, line 32: People might be confused why it can't be parsed as valid yaml, so we should clarify we're attempting it as json
```suggestion
					'A `.prettierrc` config already exists and cannot be parsed as JSON. Skipping initialization.'
```

AdrianGonz97 (approved):
thank you!

PR #531, opened 2025-04-10
feat: adds `--install <package-manager>` flag to `create` and `add`

### LINKED ISSUE #529: Add `--install` option to `sv create` and `sv add`
If it were possible to specify an `--install` option, it would mean that `sv create` and `sv add` could be fully non-interactive provided all other options were specified:

```bash
# use the default package manager
npx sv create --install

# use the specified package manager
npx sv create --install=pnpm
```

AdrianGonz97:
closes #529

### DIFF
--- a/.changeset/happy-moles-glow.md
+++ b/.changeset/happy-moles-glow.md
@@ -1,0 +1,5 @@
+---
+'sv': patch
+---
+
+feat: adds `--install <package-manager>` flag to `create` and `add`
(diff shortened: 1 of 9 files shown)


manuel3108:
Looks good and works as expected. We should probably add this flag to the docs though

manuel3108:
Thats great. For `--types` we display two commands though:
image

We should at least unify the usage. I would prefer the way we currently have it for `--types` as i think it's more intuitive to understand, at least for people who have not that much experience with cli's. Happy to do the changes myself, but would love to hear your opinion

AdrianGonz97:
ah, nice catch

> Happy to do the changes myself, but would love to hear your opinion

Feel free to update and merge after!

manuel3108 (approved):
Actually, I do wonder why we display options differently on the `create` and `add` pages altogether, but that's something we can fix at a later point

PR #534, opened 2025-04-11
chore: get rid of `@sveltejs/ast-tooling`

manuel3108:
This package was initially created to avoid long bundle times caused by `recast` and `rollup` by extracting the dependencies and their usages into an separate packages. 

Since we are no longer using `recast` and `rollup` having this as a separate package is no longer necessary.

Together with #526 this is also an attempt at reducing the number of packages that we have, so that in the future we can stop bundling everything together.


benmccann (approved):
makes sense to me

AdrianGonz97 (approved):

PR #535, opened 2025-04-13
feat: added style to login form

brettearle:
Not sure if this is needed but it adds a consistency to the demo app. Happy to do other pages of lucia demo if styling is needed. 

Screenshot from 2025-04-14 06-55-41

### DIFF
--- a/.changeset/happy-moles-glow.md
+++ b/.changeset/happy-moles-glow.md
@@ -1,5 +1,0 @@
----
-'sv': patch
----
-
-feat: adds `--install <package-manager>` flag to `create` and `add`
--- a/.changeset/pretty-emus-breathe.md
+++ b/.changeset/pretty-emus-breathe.md
@@ -1,5 +1,0 @@
----
-'sv': patch
----
-
-fix: warn on an unparsable `.prettierrc` config file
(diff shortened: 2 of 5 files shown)


manuel3108:
Hmmm, not too sure about this one. Generally I like the idea, since especially since when you add tailwindcss and lucia at the same time, you are greeted with a horrible looking page:
image

If you don't add tailwind, I think the demo works as expected and does not need any adjustment. That's why I would suggest just adding a few tailwindcss classes to make it work with tailwind as well. Here is an example of what i mean:

```html
<h1>Login/Register</h1>
<form method="post" action="?/login" use:enhance>
    <label>
        Username
        <input class="border" name="username" />
    </label>
    <label>
        Password
        <input class="border" type="password" name="password" />
    </label>
    <button
        class="bg-blue-500 hover:bg-blue-700 text-white font-bold py-2 px-4 rounded"
        >Login</button
    >
    <button
        class="bg-blue-500 hover:bg-blue-700 text-white font-bold py-2 px-4 rounded"
        formaction="?/register">Register</button
    >
</form>
<p style="color: red">{form?.message ?? ''}</p>
```

image

What do you think about this?

brettearle:
@manuel3108 yeah I can do that, I don't use the tailwind add-on and thought the login page was pragmatic but not refined like rest of the demo. 
I'll add the tailwind classes instead of base styles.

Issue #509, opened 2025-03-24
add `@types/node` to `npx sv create` templates


manuel3108:
Appeared in #484 and #497 
Still unsure when exactly this happens, but since it was reported by at least two different persons, it seems to be a real problem. 

Based on both issue, it looks like this problem is unrelated to any specific add-on and should probably be set-up for all three create templates that we currently have.

Issue #497, opened 2025-03-21
sv create: Svelte library set `moduleResolution` to `NodeNext`, not `bundler`


fel1x-developer:
```console
(project directory)/vite.config.ts:8:76
Error: The 'import.meta' meta-property is not allowed in files which will build into CommonJS output. 
const dirname =
        typeof __dirname !== 'undefined' ? __dirname : path.dirname(fileURLToPath(import.meta.url));
```

Current `tsconfig.json`
```json
{
	"extends": "./.svelte-kit/tsconfig.json",
	"compilerOptions": {
		"allowJs": true,
		"checkJs": true,
		"esModuleInterop": true,
		"forceConsistentCasingInFileNames": true,
		"resolveJsonModule": true,
		"skipLibCheck": true,
		"sourceMap": true,
		"strict": true,
		"module": "NodeNext",
		"moduleResolution": "NodeNext"
	}
}
```

Updated `tsconfig.json`. Error is resolved.
```json
{
	"extends": "./.svelte-kit/tsconfig.json",
	"compilerOptions": {
		"allowJs": true,
		"checkJs": true,
		"esModuleInterop": true,
		"forceConsistentCasingInFileNames": true,
		"resolveJsonModule": true,
		"skipLibCheck": true,
		"sourceMap": true,
		"strict": true,
		"module": "ESNext",
		"moduleResolution": "bundler"
	}
}
```

hyunbinseo:
I believe `NodeNext` is only used in the Svelte library template, since it is meant to be packaged and distributed as a package.

benmccann:
> not allowed in files which will build into CommonJS output. 

You generally shouldn't be compiling to CommonJS. The whole Svelte ecosystem is ESM-only

fel1x-developer:
Sorry my `tsconfig.json` was modified by some script. After copying new `tsconfig.json` to the working directory, everything works fine. I apologize for causing confusion.

fel1x-developer:
Reopen as reproduction is found. This is caused when `sv create` with Svelte library option.
```console
~
➜ bunx sv create d
┌  Welcome to the Svelte CLI! (v0.7.0)
│
◇  Which template would you like?
│  Svelte library
│
◇  Add type checking with TypeScript?
│  Yes, using TypeScript syntax
│
◆  Project created
│
◇  What would you like to add to your project? (use arrow keys / space bar)
│  none
│
◇  Which package manager do you want to install dependencies with?
│  bun
│
[...]

Issue #484, opened 2025-03-12
Improve drizzle-orm SQLite default template


hyunbinseo:
As of Svelte CLI v0.6.26

- [x] `db:generate` is not included in the scripts:

```jsonc
  "db:push": "drizzle-kit push",
  "db:migrate": "drizzle-kit migrate",
  "db:studio": "drizzle-kit studio"
  "db:migrate": "drizzle-kit migrate", // missing
```

- [x] Unused variable is included in the `schema.ts` file:

```js
// 'text' is declared but its value is never read.ts(6133)
import { sqliteTable, text, integer } from 'drizzle-orm/sqlite-core';

export const user = sqliteTable('user', {
  id: integer('id').primaryKey(),
  age: integer('age')
});
```

- [x] Line-breaks are missing in the `drizzle.config.ts` file:

```js
import { defineConfig } from 'drizzle-kit';
if (!process.env.DATABASE_URL) throw new Error('DATABASE_URL is not set');
```

- [ ] `loadEnvFile` should probably be called in the `drizzle.config.ts` file:

```js
import { defineConfig } from 'drizzle-kit';
import { loadEnvFile } from 'node:process';

loadEnvFile();

if (!process.env.DATABASE_URL) throw new Error('DATABASE_URL is not set');
```

- [x] `@types/node` should probably be installed out-of-the box.

```js
import { defineConfig } from 'drizzle-kit';

// Cannot find name 'process'. Do you need to install type definitions for node? Try `npm i --save-dev @types/node`.ts(2580)
if (!process.env.DATABASE_URL) throw new Error('DATABASE_URL is not set');
```

- [x] Line-breaks are missing in the `src\lib\server\db\index.ts` file:

```js
import { drizzle } from 'drizzle-orm/libsql';
import { createClient } from '@libsql/client';
import * as schema from './schema';
import { env } from '$env/dynamic/private';
if (!env.DATABASE_URL) throw new Error('DATABASE_URL is not set');
const client = createClient({ url: env.DATABASE_URL });
```

Are there reasons behind these decisions? If not, I would love to create a PR.

manuel3108:
> 1\. `db:generate` is not included in the scripts:

You are right. Either we should remove `migrate` or add `generate`, as per the [docs](https://orm.drizzle.team/docs/drizzle-kit-migrate).

> 2\. Unused variable is included in the `schema.ts` file:

You are right!

> 3\. Line-breaks are missing in the `drizzle.config.ts` file:

[...]

PR #536, opened 2025-04-14
fix: Create templates added types/node

brettearle:
Closes: #509

### DIFF
--- a/packages/create/index.ts
+++ b/packages/create/index.ts
@@ -1,6 +1,6 @@
 import fs from 'node:fs';
 import path from 'node:path';
-import { mkdirp, copy, dist } from './utils.ts';
+import { mkdirp, copy, dist, usrNodeVer } from './utils.ts';
 
 export type TemplateType = (typeof templateTypes)[number];
 export type LanguageType = (typeof languageTypes)[number];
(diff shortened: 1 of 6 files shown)


hyunbinseo:
SvelteKit users could be using older versions of Node.js. v22 should not be hardcoded:

```json
{ "engines": { "node": ">=18.13" } }
```

brettearle:
Yeah that is fair, a post install script is bloat for created project. Is there a simple way that you can think of? I'm thinking of adding a node detection function that can be used during the copy over of package.json from template to build out version of types/node.

manuel3108:
Thanks for the PR! I'm not actually sure this is a safe way. Going to the [node downloads](https://nodejs.org/en/download/current) it currently suggest downloading v23, but as far as I can see, there is no `@types/node` with version `23`. I'm assuming similar things can happen for minor and patch versions as well.

Just having `^18.13` sounds like a safer way but relies on us adjusting that version to SK from time to time. But I would say that should be fine.

AdrianGonz97:
I'm not particularly convinced that this is something that we need add to the base `create` templates. 

These type errors really only occur when native node modules are being used in the project.

In the case of [the `drizzle` add-on issue](https://github.com/sveltejs/cli/issues/484), the fix would be to just add the `@types/node` package as a `devDependency` during its installation as we're using the `process` module to access env vars. 

For [the `storybook` issue](https://github.com/sveltejs/cli/issues/497), `@types/node` should really be added on storybook's end as we're simply running their initialization process via their CLI (or, if absolutely necessary, we could do it on our end too).

brettearle:
Alright, I'll adjust the PR to just 18.13 on templates for now. Is that the plan? 

Then we can update if needed to add-on by add-on basis?

PR #537, opened 2025-04-18
fix: mysql2 closed state (drizzle addon)

netbrum:
Changed `await mysql.createConnection()` to `mysql.createPool()`

--

When using `mysql.createConnection()`, the connection will end up getting closed if it's not used for longer than [wait_timeout](https://dev.mysql.com/doc/refman/8.4/en/server-system-variables.html#sysvar_wait_timeout). Error message from mysql2 after a period of inactivity can be seen below:

```
Error: Can't add new command when connection is in closed state
```

Since drizzle doesn't handle any of the connection, you would need to juggle the database client and drizzle around before querying to make sure the connection to the database wasn't closed.

I haven't experienced this with postgres.js, as it's seemingly using [pools by default](https://github.com/porsager/postgres?tab=readme-ov-file#the-connection-pool).

If not using pools with mysql2 is intentional, please let me know, thanks.

### DIFF
--- a/.changeset/poor-ducks-own.md
+++ b/.changeset/poor-ducks-own.md
@@ -2,4 +2,4 @@
 'sv': patch
 ---
 
-fix: mysql2 closed state (drizzle addon)
+fix: use connection pool when using mysql2 with Drizzle
--- a/.changeset/poor-ducks-own.md
+++ b/.changeset/poor-ducks-own.md
@@ -1,0 +1,5 @@
+---
+'sv': patch
+---
+
+fix: mysql2 closed state (drizzle addon)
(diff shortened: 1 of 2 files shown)


.changeset/poor-ducks-own.md
```diff
@@ -0,0 +1,5 @@
+---
+'sv': patch
+---
+
+fix: mysql2 closed state (drizzle addon)
```
benmccann, line 5:
```suggestion
fix: use connection pool when using mysql2 with Drizzle
```

netbrum: Thanks, I wasn't sure if I should use the pr title or explain the changes directly

manuel3108 (approved):
Thanks for your help!

PR #539, opened 2025-04-19
fix: add `@types/node` when adding `drizzle`

brettearle:
closes: #509 

Supersedes #536

### DIFF
--- a/.changeset/soft-candies-guess.md
+++ b/.changeset/soft-candies-guess.md
@@ -1,0 +1,5 @@
+---
+'sv': patch
+---
+
+fix: add `@types/node` when adding `drizzle`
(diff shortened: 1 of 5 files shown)


brettearle:
Hey @manuel3108 here is the PR to just add compatible node/types to the create. Did it on a clean branch and closed #536, can delete as well if needed

manuel3108:
@AdrianGonz97 did a much better job at analyzing the actual problem [here](https://github.com/sveltejs/cli/pull/536#issuecomment-2816034833) then I did. It's not necessary to add this package to everyone's installation, when it's only necessary in certain circumstances. I have updated this PR to reflect this, thanks for your help!

As for the storybook integration, I contacted one of the maintainers in #497, let's see where we can get from there on.

PR #540, opened 2025-04-19
fix:Tailwind style to lucia login

brettearle:
Tailwind styles added to make the login legible with tailwind selected


brettearle:
@manuel3108 have added tailwind to this standard

Screenshot from 2025-04-19 10-59-41

benmccann:
This seems rather unexpected to me if you setup Lucia and not Tailwind

brettearle:
Yeah it is, did have it as base styling. Happy to push it further and have
it conditional. At the moment the inputs are invisible with tailwind and
Lucia in the demo

brettearle:
or just make the styling inline normal css?

brettearle:
@benmccann would this be the way to do it? I was trying to use what already existed without adding new parameters etc.

@manuel3108 does this meet what you were expecting?

manuel3108:
Generally I think the proposed changes seem to be reasonable, to make the lucia page not destroyed in the `minimal` template by tailwind.

This code makes the assumption that the lucia integration is run after the `tailwind` integration, even if they are selected in the same prompt. If I have our add-on ordering code in mind correctly, this should currently always apply. But I will need to check this at a later stage.

Currently, this PR seems to introduce build errors while running the templates.

brettearle:
Awesome, I'll fix the build errors. Seems to just be the quotes causing issues. Will sort. Let me know if I can be a hand on the ordering system is needed.

brettearle:
In draft whilst I get pipeline working locally

manuel3108:
Merged #554 which allows us to set `runsAfter` in lucia, which explicitly runs the lucia after tailwind. This makes the example work, when you are adding tailwind and lucia in the same command.

Thanks for your help!

PR #541, opened 2025-04-20
fix: add `@types/node` to `drizzle` and `storybook` add-ons

### LINKED ISSUE #509: add `@types/node` to `npx sv create` templates

AdrianGonz97:
closes #509

This is an alternative to #536 and #539 where `@types/node` is added to `devDependencies` by the `drizzle` and `storybook` add-ons. 

The version of `@types/node` is derived from the version of node from the current running process. Since `@types/node` only publishes types for `lts` versions of node, we'll need to install the nearest `lts` version for `current` releases. To do this, we'll just decrement the major number by one (and chop off the minor and patch numbers) to target the previous `lts` release version (note that we can detect whether a release is `lts` or `current` based on the if the major is an even or odd number, respectively)

For example: if the installed node version is `22.14.0` (`lts`), then `^22` will be added to the `package.json`. If the node version is `23.11.0` (`current`), then `^22` will be added instead.

### DIFF
--- a/.changeset/odd-bikes-like.md
+++ b/.changeset/odd-bikes-like.md
@@ -1,0 +1,5 @@
+---
+'sv': patch
+---
+
+fix: add `@types/node` as a dev dependency to the `drizzle` and `storybook` add-ons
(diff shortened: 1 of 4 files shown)


brettearle:
Nice work. I didn't think of the odd even check.

manuel3108 (approved):
Seems to be working perfectly

PR #547, opened 2025-04-24
fix: Always add Storybook last

JReinhold:
Ensure that if Storybook is selected during creation, it is always added after the other addon.

See _Side note_ in https://github.com/sveltejs/cli/issues/497#issuecomment-2823449201 for context

### DIFF
--- a/packages/cli/lib/install.ts
+++ b/packages/cli/lib/install.ts
@@ -185,6 +185,11 @@ async function runAddon({ addon, multiple, workspace }: RunAddon) {
 // but works for now in contrary to the previouse implementation
 function orderAddons(addons: Array<Addon<any>>, setupResults: Record<string, AddonSetupResult>) {
 	return addons.sort(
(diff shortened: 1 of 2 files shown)


packages/cli/lib/install.ts
```diff
@@ -185,6 +185,11 @@ async function runAddon({ addon, multiple, workspace }: RunAddon) {
 // but works for now in contrary to the previouse implementation
```
manuel3108, line 185:
```suggestion
// but works for now in contrary to the previous implementation
```

JReinhold:
I see, makes sense to me. I played around with `dependsOn` at first, but it wasn't the right fit here.

Could we temporarily merge this in? As even today with the existing versions, users will get a broken out of the box, which is not a nice experience.

manuel3108:
Sounds good, opened https://github.com/sveltejs/cli/issues/550 for the full implementation of `runsAfter`

PR #542, opened 2025-04-22
chore: use the latest adapter-auto

hyunbinseo:
Despite https://github.com/sveltejs/cli/pull/520, an outdated `adapter-auto` is added:

```json
{
  "devDependencies": {
    "@sveltejs/adapter-auto": "^4.0.0",
```

```
┌  Welcome to the Svelte CLI! (v0.8.1)
│
◇  Which template would you like?
│  SvelteKit minimal
│
◇  Add type checking with TypeScript?
│  Yes, using TypeScript syntax
│
◆  Project created
│
◇  What would you like to add to your project? (use arrow keys / space bar)
│  none
```

Probably because the adapter is not installed as an add-on?

PR #548, opened 2025-04-24
security: upgrade vite to avoid CVE-2025-32395

hyunbinseo:
PR for https://github.com/sveltejs/cli/pull/522#issuecomment-2800227299

https://github.com/advisories/GHSA-356w-63v5-8wf4

### DIFF
--- a/.changeset/yummy-things-learn.md
+++ b/.changeset/yummy-things-learn.md
@@ -1,0 +1,5 @@
+---
+'sv': patch
+---
+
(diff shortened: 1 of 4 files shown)


manuel3108:
Thanks!
We will need to a find a better way to handle that stuff in the future, potentially by only specifying `^6` in the future, but I will merge this for now.

hyunbinseo:
Agreed. Would be nice if the `package.json` files' versions can be easily bumped. Similar to https://github.com/sveltejs/cli/pull/542

PR #552, opened 2025-04-25
docs: remove duplicate list of add-ons

benmccann:


manuel3108:
Personally, I think the list is very needed, especially on mobile as you don't see the all the available on a first glance and you first need to have the idea of opening the navigation to get that navigation. Instead, I think we should link each adder to the details page

benmccann:
Good point about mobile. I've updated it as you've suggested

PR #553, opened 2025-04-27
debug: argument parsing

manuel3108:
Aims to provide some debug information for some weirdness encountered while testing #543. This PR is only here to get a preview release.

Commands to run:
```
pnpx https://pkg.pr.new/sveltejs/cli/sv@553 add paraglide=languageTags:en,de,demo:no
```

```
pnpx https://pkg.pr.new/sveltejs/cli/sv@553 add paraglide="languageTags:en,de,demo:no"
```

issue currently present on 1 of 2 windows machines. This issue is potentially related to the region the PC was setup or something. Not present on linux as far as we can tell


manuel3108:
Ahh, based of the incorrect branch, but does not matter for this.

manuel3108:
Root cause identified, powershell is the problem. Does not happen while using CMD. We will just quote all args in the docs, and we should be good to go

PR #555, opened 2025-05-03
chore: remove unused drizzle driver code

benmccann:
I'm not sure that this line does anything. It looks like at some point there was a `driver` property defined for the non-sqlite DBs, but I don't see anything setting it, so don't know that this line is needed anymore


benmccann:
Closing in favor of https://github.com/sveltejs/cli/pull/556

PR #556, opened 2025-05-03
chore: leverage expressionFromString more in drizzle adder

benmccann:
A bit of the code here was unused. See https://github.com/sveltejs/cli/pull/555

~~I haven't gotten a chance to test this yet, but thought I'd share my progress before I have to run for the day. Manuel thought that we might need to tweak `expressionFromString` to delete the `Literal.raw property` so that it can pick the proper quote type. I've not checked that yet and assume we'll need to update it~~

### DIFF
--- a/packages/addons/drizzle/index.ts
+++ b/packages/addons/drizzle/index.ts
@@ -189,12 +189,12 @@ export default defineAddon({
 				ast,
 				common.expressionFromString(`
 					defineConfig({
-						schema: "./src/lib/server/db/schema.${typescript ? 'ts' : 'js'}".
+						schema: "./src/lib/server/db/schema.${typescript ? 'ts' : 'js'}",
 						dialect: ${options.sqlite === 'turso' ? 'turso' : options.database}
 						dbCredentials: {
 							url: process.env.DATABASE_URL,
 							${options.sqlite === 'turso' ? 'authToken: process.env.DATABASE_AUTH_TOKEN' : undefined}
-						}
+						},
 						verbose: true,
 						strict: true
 					})`)
--- a/packages/addons/drizzle/index.ts
+++ b/packages/addons/drizzle/index.ts
@@ -190,7 +190,7 @@ export default defineAddon({
 				common.expressionFromString(`
 					defineConfig({
 						schema: "./src/lib/server/db/schema.${typescript ? 'ts' : 'js'}",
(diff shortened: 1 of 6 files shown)


manuel3108:
Made the following changes:
* fixed the drizzle output so that the tests can run
* for some reason, we format all test results with prettier (valid for js / html / css). This is probably a leftover from when the js block often were poorly formatted. 
* now that prettier is not run anymore, i added a failing test that Ben already mentioned in the PR description. We might be able to fix this by calling `stripAst` from within `expressionFromString` but that might be to aggressive as this will be done on all child nodes, instead as only on Literals

packages/core/tests/js/common/expression-from-string/output.ts
```diff
@@ -0,0 +1 @@
+export default defineConfig({ path: 'some/string/as/path.js', valid: true });
```
manuel3108, line 1: To be exact, this would be my expectation because [`guessQuoteStyle`](https://github.com/sveltejs/cli/blob/e9485a2c8e6df490291490d91059a2540ab9653a/packages/core/tooling/index.ts#L211) should not be able to find any quotes and thus return `undefined`. Therefore the `esrap` default quote style 'single' should apply: https://github.com/sveltejs/esrap?tab=readme-ov-file#options

packages/addons/drizzle/index.ts
```diff
@@ -69,8 +71,21 @@ export default defineAddon({
...
+		const paths = {
+			'drizzle config': path.relative(cwd, path.resolve(cwd, `drizzle.config.${ext}`)),
+			'database schema': path.relative(cwd, path.resolve(baseDBPath, `schema.${ext}`)),
+			database: path.relative(cwd, path.resolve(baseDBPath, `index.${ext}`))
+		};
+		for (const [fileType, filePath] of Object.entries(paths)) {
+			if (fs.existsSync(filePath)) {
+				unsupported(`Preexisting ${fileType} file at '${filePath}'`);
+			}
+		}
```
AdrianGonz97, line 88: Added this check so the add-on can't be re-executed on a already drizzle-initialized project

AdrianGonz97 (approved):

manuel3108 (approved):

manuel3108:
No changeset, but should not be necessary here.

Issue #493, opened 2025-03-18
Update lucia integration to use the new `getRequestEvent` functionality


ollema:
I think it would make sense if this check:
https://github.com/sveltejs/cli/blob/d1f4432c0e31677ae754d50394547eeff820ddfe/packages/addons/lucia/index.ts#L539-L544
was rewritten into a shared function using the new [`getRequestEvent`](https://svelte.dev/docs/kit/load#Using-getRequestEvent)

dummdidumm:
Why? Is this repeated in multiple places? (I'm not well-versed with this code, but it looks like this wouldn't give us anything)

ollema:
> Why? Is this repeated in multiple places? (I'm not well-versed with this code, but it looks like this wouldn't give us anything)

oh no it would not improve anything with this code since it's not repeated as you pointed out.

however I think it's a really good pattern to adhere to when it comes to auth guards, so to me it makes sense that you would include such a function in the code generated by the official svelte CLI if you understand what I mean.

as Rich wrote in the PR description:

> It's also useful if you need to do auth stuff. For example we can make a utility function that redirects the user to /login if they're not logged in already

so I figured why not encourage just this use case?

---

in a slightly similar vein, the lucia integration also adds a `validatePassword` which is not repeated and yet it is included because I assume it is good practice and it makes the good easier to read and maintain.

so maybe the only criteria is not "is this repeated in multiple places?", is what I am trying to get across, not sure if I succeeded.

PR #558, opened 2025-05-08
fix: lucia require user redirect moved to auth guard function

brettearle:
closes: #493 

Updated to use pattern similar to login utilities. This pattern matches the SvelteKit docs and allows people to see it in use straight away when adding the lucia addon.


manuel3108:
Added a changeset for the changelog. Thanks for your help Brett!

PR #565, opened 2025-05-17
fix: account for Node `current` releases with even majors when resolving `@types/node` version

### LINKED ISSUE #560: No matching version found for `@types/node@^24`
Probably caused by https://github.com/sveltejs/cli/pull/541

Node.js v24 has been released, and using it to initiate a SvelteKit project fails:

```
│  pnpm
│
■  Failed to install dependencies

Progress: resolved 1, reused 0, downloaded 0, added 0
 ERR_PNPM_NO_MATCHING_VERSION  No matching version found for @types/node@^24 while fetching it from https://registry.npmjs.org/

This error happened while installing a direct dependency of C:\Users\Hyunbin\Development\@temp

The latest release of @types/node is "22.15.18".
```

AdrianGonz97:
Closes #560

### DIFF
--- a/.changeset/dull-pens-wash.md
+++ b/.changeset/dull-pens-wash.md
@@ -1,0 +1,5 @@
+---
+'sv': patch
+---
+
+fix: account for Node `current` releases with even majors when resolving `@types/node` version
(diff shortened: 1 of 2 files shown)


manuel3108 (approved):
There is still a minor possibilty for an issue here. Let's assume that 24 becomes an LTS but the `@types/node` package is only published a few days later. But let's hope that this will never be the case for now

Issue #570, opened 2025-05-19
npm run lint fails on most basic setup with drizzle and eslint


BridgerB:
src/lib/server/db/schema.js contains:
```
import { pgTable, serial, text, integer } from 'drizzle-orm/pg-core';

export const user = pgTable('user', {
	id: serial('id').primaryKey(),
	age: integer('age')
});
```

but that fails this basic lint because of the text import:
```
~/temp/svtest2 $ npm run lint

> svtest2@0.0.1 lint
> eslint .


/home/bridger/temp/svtest2/src/lib/server/db/schema.js
  1:27  error  'text' is defined but never used  no-unused-vars

✖ 1 problem (1 error, 0 warnings)
```

Setup:
```
bridger@nixos ~/temp/svtest2 $ npx sv create .
Need to install the following packages:
sv@0.8.6
Ok to proceed? (y) y

┌  Welcome to the Svelte CLI! (v0.8.6)
│
◇  Which template would you like?
│  SvelteKit minimal
│
◇  Add type checking with TypeScript?
│  Yes, using JavaScript with JSDoc comments
│
◆  Project created
│
◇  What would you like to add to your project? (use arrow keys / space bar)
│  eslint, drizzle
│
◇  drizzle: Which database would you like to use?
│  PostgreSQL
│
◇  drizzle: Which PostgreSQL client would you like to use?
│  Postgres.JS
│
◇  drizzle: Do you want to run the database locally with docker-compose?
│  Yes
│
◆  Successfully setup add-ons
│
◇  Which package manager do you want to install dependencies with?
│  npm
│
◆  Successfully installed dependencies
│
◇  Project next steps ─────────────────────────────────────────────────────╮
│                                                                          │
│  1: git init && git add -A && git commit -m "Initial commit" (optional)  │
│  2: npm run dev -- --open                                                │
│                                                                          │
│  To close the dev server, hit Ctrl-C                                     │
│                                                                          │
│  Stuck? Visit us at https://svelte.dev/chat                              │
│                                                                          │
├──────────────────────────────────────────────────────────────────────────╯
│
◇  Add-on next steps ──────────────────────────────────────────────────╮
│                                                                      │
│  drizzle:                                                            │
│  - You will need to set DATABASE_URL in your production environment  │
│  - Run npm run db:start to start the docker container                │
│  - Run npm run db:push to update your database schema                │
│                                                                      │
├──────────────────────────────────────────────────────────────────────╯
│
└  You're all set!
```

[...]

PR #571, opened 2025-05-19
fix: removed unused import in `drizzle` schema to fix lint

BridgerB:
Issue: https://github.com/sveltejs/cli/issues/570


manuel3108 (approved):
Nice catch, and thanks for the fix!

Did the same for `sqlite` and `mysql` as the same error was present in those cases.

Issue #578, opened 2025-06-05
Add-on for vite-plugin-devtools-json


Rich-Harris:
Recently Chrome started requesting `/.well-known/appspecific/com.chrome.devtools.json` whenever you have devtools open on a `localhost` URL. As of https://github.com/sveltejs/kit/pull/13830, SvelteKit squelches the resulting 404 error, but on the first occurrence prints a message to the console suggesting that people install [vite-plugin-devtools-json](https://github.com/ChromeDevTools/vite-plugin-devtools-json). 

Maybe it's worth having an addon for this, so that people can just do `sv add vite-plugin-devtools-json`? Alternatively, or additionally, it could be installed by default in a new project?

Rich-Harris:
(If we _do_ do this, we should update the message in Kit to tell people to do `sv add` instead of manually installing it and configuring Vite)

jycouet:
The PR limiting logs is nice, thank you!
When I saw this issue, I told myself that it's a good way to navigate into `sv` by a simple thing.

> (If we _do_ do this, we should update the message in Kit to tell people to do `sv add` instead of manually installing it and configuring Vite)

Not sure where ?

PR #581, opened 2025-06-07
feat: add `vite-plugin-devtools-json` addon

jycouet:
Closes: https://github.com/sveltejs/cli/issues/578

- [x] feature: adding `vite-plugin-devtools-json` plugin to vite config
- [x] tests (js & ts)
- [x] doc

---

- ~[ ] https://github.com/sveltejs/cli/pull/582 needs to be merged before, then tests will pass here.~
- [x] waiting for vitest regression fix


jycouet:
Okay, I think that it's an issue with `svelte@5.33.17`!

jycouet:
Ho WoW, it's linked with https://github.com/sveltejs/svelte/issues/11394#issuecomment-2085747668
It's `vitest` that is failing... and it's failing also in `main` branch! (and with the current released `sv`).

I will do a PR to fix this. (I will do a separate one, to track things better)

.changeset/twelve-shirts-tell.md
```diff
@@ -0,0 +1,5 @@
+---
+'sv': patch
+---
+
+add vite-plugin-devtools-json addon
```
manuel3108:
```suggestion
feat: add `vite-plugin-devtools-json` addon
```

documentation/docs/30-add-ons/60-vite-plugin-devtools-json .md
```diff
@@ -0,0 +1,15 @@
...
+
+## Usage
+
+```bash
+npx sv add vite-plugin-devtools-json
+```
+
+## What you get
+
+- the plugin added to your vite plugin options.
```
manuel3108:
```suggestion
- the `vite` plugin added to your vite plugin options.
```

documentation/docs/30-add-ons/60-vite-plugin-devtools-json .md
```diff
@@ -0,0 +1,15 @@
+---
+title: vite-plugin-devtools-json
```
manuel3108: We should also add the plugin to this list:
https://github.com/sveltejs/cli/blob/17799ccefc86a0348b4cc38535ed215a107071c7/documentation/docs/20-commands/20-sv-add.md?plain=1#L26

Also the webapp deployment is failing according to vercel, here are the build logs. (I have no clue what's happening there)
```
Error: 404 /docs/cli/vite-plugin-devtools-json (linked from /docs/cli/overview)
To suppress or handle this error, implement `handleHttpError` in https://svelte.dev/docs/kit/configuration#prerender
    at file:///vercel/path0/node_modules/.pnpm/@sveltejs+kit@2.20.0_@sveltejs+vite-plugin-svelte@4.0.3_svelte@5.33.0_vite@5.4.11_@types+node_6r6su57xyiuzz4mkviaa4bhwte/node_modules/@sveltejs/kit/src/core/config/options.js:204:13
    at file:///vercel/path0/node_modules/.pnpm/@sveltejs+kit@2.20.0_@sveltejs+vite-plugin-svelte@4.0.3_svelte@5.33.0_vite@5.4.11_@types+node_6r6su57xyiuzz4mkviaa4bhwte/node_modules/@sveltejs/kit/src/core/postbuild/prerender.js:73:25
    at save (file:///vercel/path0/node_modules/.pnpm/@sveltejs+kit@2.20.0_@sveltejs+vite-plugin-svelte@4.0.3_svelte@5.33.0_vite@5.4.11_@types+node_6r6su57xyiuzz4mkviaa4bhwte/node_modules/@sveltejs/kit/src/core/postbuild/prerender.js:439:4)
    at visit (file:///vercel/path0/node_modules/.pnpm/@sveltejs+kit@2.20.0_@sveltejs+vite-plugin-svelte@4.0.3_svelte@5.33.0_vite@5.4.11_@types+node_6r6su57xyiuzz4mkviaa4bhwte/node_modules/@sveltejs/kit/src/core/postbuild/prerender.js:261:3)
Emitted 'error' event on Worker instance at:
    at [kOnErrorMessage] (node:internal/worker:333:10)
    at [kOnMessage] (node:internal/worker:344:37)
    at MessagePort.<anonymous> (node:internal/worker:233:57)
    at [nodejs.internal.kHybridDispatch] (node:internal/event_target:831:20)
    at MessagePort.<anonymous> (node:internal/per_context/messageport:23:28)
```

packages/addons/vite-plugin-devtools-json/index.ts
```diff
@@ -0,0 +1,34 @@
+import { defineAddon } from '@sveltejs/cli-core';
+import { array, functions, imports, object, exports } from '@sveltejs/cli-core/js';
+import { parseScript } from '@sveltejs/cli-core/parsers';
+
+export default defineAddon({
+	id: 'vite-plugin-devtools-json',
+	alias: 'devtools-json',
+	shortDescription: 'Generate DevTools project settings in the devserver',
```
manuel3108: Most of the other add-ons have a much shorter description. Maybe something like:
```suggestion
	shortDescription: 'chormium devtools',
```

packages/addons/vite-plugin-devtools-json/index.ts
```diff
@@ -0,0 +1,34 @@
...
+	homepage: 'https://github.com/ChromeDevTools/vite-plugin-devtools-json',
+	options: {},
+
+	run: ({ sv, typescript }) => {
+		const ext = typescript ? 'ts' : 'js';
+
+		sv.dependency('vite-plugin-devtools-json', '^0.1.1');
```
manuel3108:
```suggestion
		sv.devDependency('vite-plugin-devtools-json', '^0.1.1');
```

manuel3108 (commented):
Thanks for your awesome PR. The only thing that I'm unsure about is the case for newly created projects. 

Normally I would expect that a newly created project will not throw any errors in the console as this will be very strange for newcomers. On the other hand i don't necessarily want to add code that only helps users with specific browsers. Given that the problem also arises in chromium based browsers like MS Edge we are talking about [~70%](https://radar.cloudflare.com/reports/browser-market-share-2025-q1) market share. I would say that this is not enough to bother everyone with this new tool, so making it an adder sounds like the correct option. We only need to rise awareness about this add-on.

The other thing I'm concerned about is the add-on name `vite-plugin-devtools-json` feels clunky and is not telling what it's actually for. We could consider something like `chromium-devtools` but I'm sure people will have better ideas than myself.

Edit: As you said, don't bother about the `vitest` tests, that's a seperate issue

documentation/docs/30-add-ons/60-vite-plugin-devtools-json .md
jycouet: I have an idea why... I think I left a space somewhere. I'll try to fix this.

jycouet:
> 70% market share. I would say that this is not enough to bother everyone

I'm curious, what's your threshold ? _Seems pretty high already._
But yes, if you feel that it's more appropriate, we can remove it from create.
How should I do it, I add a flag in `defineAddon` like: `skipCreate: boolean` ? Or Something in `setup` ?

---

> I'm concerned about is the add-on name vite-plugin-devtools-json

Yes, you are right, it's not that cool! Most of _(or at least a lot of)_ vite plugins starts like this `vite-plugin-*` and that's a nice convention (pck json organisation for example!)

I like your suggestion `chromium-devtools`, but it's adding a new term "chromium" and removing one "json", so it would be a "svelte thing". If we plan to add more thing to chromium-devtools, yes, let's go for it. If we don't plan more things we could maybe just call it `devtools-json` (keeping the author vibe?)

Let me know, I'll update the PR 👍

documentation/docs/30-add-ons/60-vite-plugin-devtools-json .md
jycouet: I guess it's fixed ? (there is a way to redeploy the vercel part to check ?)

manuel3108: Yeah it fixed itself after one of your pushes i think: https://svelte-dev-git-preview-cli-581-svelte.vercel.app/

manuel3108:
> I'm curious, what's your threshold ? _Seems pretty high already._

I don't know, just a feeling. 30% of people might not have that problem. But on the other hand, for us devs, the percentage is potentially totally different.

> How should I do it, I add a flag in `defineAddon` like: `skipCreate: boolean` ? Or Something in `setup` ?

I think it's fine as is. I was referring to Rich's suggestion to add the plugin directly without user prompt when creating a new project. That's the thing i was considering. That the add-on is visible in the addon selection after `sv create` is totally fine by me.

> I like your suggestion `chromium-devtools`, but it's adding a new term "chromium" and removing one "json", so it would be a "svelte thing". If we plan to add more thing to chromium-devtools, yes, let's go for it. If we don't plan more things we could maybe just call it `devtools-json` (keeping the author vibe?)

Yeah, totally get your point. I had to refresh my memories because https://github.com/sveltejs/svelte/discussions/15910 was once called devtools, but since this is no longer the case we might be good to use `devtools-json`. But i still se an edge where this could be confusing. 

Would love to get other opinions on this.

jycouet:
I would like to raise another option that is turning in my head.

Why not having:
```sh
npx sv add vite-plugin devtools-json
# by default it's:
#   1/ adding devDep: "vite-plugin-devtools-json"
#   2/ importing the plugin in list of vite plugins (with default export camel case `devtoolsJson`)
```

This would allow any* vite plugin to be added!

I could see:
```sh
npx sv add vite-plugin kit-routes -n
# -n for named plugin `import { kitRoutes }...`
```

Feel free to discard the idea :)

manuel3108:
I don't this is a good idea for two reasons:
1. our cli args are already complicated another, adding another layer for vite args does not make sense I think.
2. What would happen if a vite plugin requires more customization than just using it? In that case it would still require a custom adder, with potential for naming collisions.

---
 I think the only thing we need to do here is to decide on the naming. To summarize;
* `vite-plugin-devtools-json` feels clunky 
* `chromium-devtools` adds a new term `chromium` . Additionally, this also applies for Edge and other chromium based browsers as well
* `devtools-json` compact and short

Currently, I would prefer `devtools-json`.
@Rich-Harris Any opinion on the naming of the add-on, as you initially suggested `vite-plugin-devtools-json`?

jycouet:
Thx for the comment @manuel3108, like this it's out of my head 😉👍

My vote goes for `devtools-json` as well

Rich-Harris:
I don't have a strong opinion on naming but would be good with `devtools-json`. 

Re 70%/30%: market share among the general population isn't really relevant; what matters is the % of developers that use Chromium, and that number is 100% — even those of us who ordinarily use other browsers are still _testing_ our apps in Chromium. Or at least we should be!

jycouet:
Just got a brillant idea _I think_!
What about selecting it by default in the list ?

Like this if nothing special, it's in. And if for some reason you want to opt out, you are one space away

manuel3108:
> Re 70%/30%: market share among the general population isn't really relevant; what matters is the % of developers that use Chromium, and that number is 100% — even those of us who ordinarily use other browsers are still _testing_ our apps in Chromium. Or at least we should be!

Yeah, that totally makes sense. In that case, we will need to figure out a way to run the `devtools-json` add-on right after creating a new project. Or is it potentially enough to pre-select that option in the adder selection prompt after creating the project?

manuel3108:
> What about selecting it by default in the list ?

Yeah just had the same thinking. Feel free to work on this if you want, otherwise I can take a look at it. I think the important thing would be to only make it preselected in the create prompt, not in the add prompt. Also we should probably make this configurable for the add-ons, potentially in the `setup` function

jycouet:
I'll play with this tonight 👍

jycouet:
_Youpladoum_ done! Was fun :)

```ts
// devtools-json is configured like this
setup: ({ defaultSelection }) => {
  defaultSelection({
    create: true,
    add: false
  });
},
```
Of course, default values are `false` & `false`.

---

Let me know

manuel3108 (approved):
This is awesome, thank you. Made one minor tweak but that's it.

PR #580, opened 2025-06-06
fix(`vitest`): add `resolve.conditions` to `client` test config

AdrianGonz97:
closes #579

It looks like our default config broke in `vitest@3.2.0`. Adding `resolve.conditions` to the client test config seemingly fixes it (as mentioned in the related issue), though I'm unsure if it's the _right_ fix. 

cc @dominikg @manuel3108 

```diff
export default defineConfig({
	// ...
	test: {
		projects: [
			{
				extends: './vite.config.ts',
				plugins: [svelteTesting()],
+				resolve: { conditions: ['browser'] },
				test: {
					name: 'client',
					environment: 'jsdom',
					clearMocks: true,
					include: ['src/**/*.svelte.{test,spec}.{js,ts}'],
					exclude: ['src/lib/server/**'],
					setupFiles: ['./vitest-setup-client.ts']
				}
			},
			{
				extends: './vite.config.ts',
				test: {
					name: 'server',
					environment: 'node',
					include: ['src/**/*.{test,spec}.{js,ts}'],
					exclude: ['src/**/*.svelte.{test,spec}.{js,ts}']
				}
			}
		]
	}
});
```

PR #582, opened 2025-06-07
fix: resolve-conditions-for-client

jycouet:
Linked with https://github.com/sveltejs/svelte/issues/11394

---

Today `sv` with `vitest` addon is failing with this error: 
```
`mount(...)` is not available on the server
```

This PR is fixing the issue (and fixing main branch)

🚨 I guess that this one is kinda "urgent" 🚨

### DIFF
--- a/.changeset/breezy-clowns-brush.md
+++ b/.changeset/breezy-clowns-brush.md
@@ -1,0 +1,5 @@
+---
+'sv': patch
+---
+
+fix:addon vitest remove the deprecated `workspace` property in favor of `projects`.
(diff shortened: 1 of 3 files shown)


manuel3108:
Duplicates #580, this is regression introduced by `vitest` https://github.com/vitest-dev/vitest/issues/8120

jycouet:
> Duplicates #580, this is regression introduced by `vitest` https://github.com/vitest-dev/vitest/issues/8120

Thank you @manuel3108 for links and keeping everyone in the loop 👍

Maybe vitest could be pin before vitest regression so that sv still work fully ?

PR #586, opened 2025-06-09
chore: update lucia template

hyunbinseo:
1. Removed trailing `.js` in import to match `module: esnext`.
2. Removed unused `e` in `try-catch` which caused ESLint error.


manuel3108 (approved):
Perfect, thank you so much!

Issue #557, opened 2025-05-04
Build code from strings more often


benmccann:
Just as is being done in https://github.com/sveltejs/cli/pull/556, I think we could simplify a lot of the addon code and make it more readable by building the expression from a string or potentially just returning a string directly instead of building an AST.

`eslint.config.js`, `paraglide/+page.svelte`, and the Vite config modification in the vitest addon are a few that looked like they could really be simplified - though I wouldn't be surprised if there were others as well.

PR #588, opened 2025-06-09
feat(vitest): vitest browser mode

manuel3108:
Supports `vitest` browser mode. Makes the setup more flexible by asking the user to add component / unit testing.

Adds a new `object.createFromPrimitives` to core that should help us find the right balance for https://github.com/sveltejs/cli/issues/557. We should all `object.create` calls in the future and try to use this wherever possible.

todo:
- [x] test in non kit project
- [x] check if error / warning for `The following Vite config options will be overridden by SvelteKit:  - root` is accurate and how to avoid it => accurate, not something we control

### DIFF
--- a/packages/addons/vitest-addon/index.ts
+++ b/packages/addons/vitest-addon/index.ts
@@ -1,14 +1,29 @@
-import { dedent, defineAddon, log } from '@sveltejs/cli-core';
+import { dedent, defineAddon, defineAddonOptions, log } from '@sveltejs/cli-core';
 import { array, common, exports, functions, imports, object } from '@sveltejs/cli-core/js';
 import { parseJson, parseScript } from '@sveltejs/cli-core/parsers';
+
+const options = defineAddonOptions({
+	usages: {
+		question: 'What do you want to use vitest for?',
+		type: 'multiselect',
+		default: ['unit', 'component'],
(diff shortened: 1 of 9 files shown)


packages/addons/vitest-addon/index.ts
```diff
@@ -1,19 +1,37 @@
...
 
 		sv.devDependency('vitest', '^3.2.3');
-		sv.devDependency('@testing-library/svelte', '^5.2.4');
-		sv.devDependency('@testing-library/jest-dom', '^6.6.3');
-		sv.devDependency('jsdom', '^26.0.0');
+
+		if (componentTesting) {
+			sv.devDependency('@vitest/browser', '^3.2.3');
+			sv.devDependency('vitest-browser-svelte', '^0.1.0');
+			sv.devDependency('playwright', '^1.0.0');
```
dominikg: would suggest picking the highest available playwright minor here, or at least the one that was available when vitest browser mode was introduced, i am not sure playwright 1.0.0 is actually compatible with it (playwright does breaking changes in minors unfortunately)

manuel3108: Does it matter though? Users should always get the latest minor installed, assuming they havn't used playwright in their project before

PR #594, opened 2025-06-15
style: doctype

milksense:


manuel3108:
I don't see a necessity to change this. It's been that way for ages, and [the casing does not matter](https://stackoverflow.com/questions/7020961/uppercase-or-lowercase-doctype) according to my research. If there is any technical limitation, please let me know.

PR #599, opened 2025-06-20
fix: resolve to `task` instead of `run` in addon instructions for `deno`

brunnerh:
Set up a project earlier and it output:
> Run **deno run db:push** to update your database schema

...which is wrong for deno.

`package-manager-detector` seems to be used in other spots for this, so I added it here as well.
Tried to find all places using `run`, but only found it in `lucia` & `drizzle`.

---

Regarding changesets, I found this statement in `CONTRIBUTING.md` a bit confusing:

> Only publish a change set if it is in 'sv' or 'svelte-migrate' as all other packages are bundled.

The changes are in `addons` but it _affects_ `sv`...

### DIFF
--- a/.changeset/silent-timers-carry.md
+++ b/.changeset/silent-timers-carry.md
@@ -1,0 +1,5 @@
+---
+"@sveltejs/addons": patch
+---
+
+fix: resolve to `task` instead of `run` in addon instructions for `deno`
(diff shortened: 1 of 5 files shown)


manuel3108 (approved):
Totally makes sense, thank you!

PR #604, opened 2025-06-20
chore: make `installAddon()` work

manuel3108:
We do have `create` that works flawlessly. But we do have `installAddon` that is unusable at this point in time, since it requires addons, that you can not export from anywhere. 

Since all of this is an undocumented api, im not too concerned.

### DIFF
--- a/packages/cli/lib/index.ts
+++ b/packages/cli/lib/index.ts
@@ -1,3 +1,4 @@
 export { create } from '@sveltejs/create';
 export { installAddon } from './install.ts';
 export type { AddonMap, InstallOptions, OptionMap } from './install.ts';
+export { officialAddons } from '@sveltejs/addons';


manuel3108:
It's not beautiful and not much typing support, but this makes it work without to much trouble:


```js
import { create, installAddon, officialAddons } from 'sv';

const cwd = './create-output';

await create(cwd, {
  name: 'my-project-name',
  template: 'minimal', // ["minimal", "demo", "library"]
  types: 'typescript', // ["typescript", "checkjs", "none"]
});

const tailwindAddon = officialAddons.find((a) => a.id === 'tailwindcss');
if (!tailwindAddon)
  throw new Error('Tailwind CSS addon not found in official addons');

await installAddon({
  addons: {
    tailwindcss: tailwindAddon,
  },
  cwd,
  packageManager: 'pnpm',
  options: {
    tailwindcss: {
      plugins: ['forms'],
    },
  },
});

```

Issue #608, opened 2025-06-25
Comments are missing from `svelte.config.js` after adding a package manager


eltigerchino:
After adding an adapter, the comments from the `svelte.config.js` file are completely removed. This includes the type guard for the config variable which is needed for autocomplete.

### Reproduction

```sh
pnpx sv create --template minimal --types ts --install pnpm test-sv
┌  Welcome to the Svelte CLI! (v0.8.12)
│
◆  Project created
│
◇  What would you like to add to your project? (use arrow keys / space bar)
│  sveltekit-adapter
│
◇  Which SvelteKit adapter would you like to use?
│  cloudflare
│
◆  Successfully setup add-ons
│
◆  Successfully installed dependencies
│
◇  Project next steps ─────────────────────────────────────────────────────╮
│                                                                          │
│  1: cd test-sv                                                           │
│  2: git init && git add -A && git commit -m "Initial commit" (optional)  │
│  3: pnpm run dev --open                                                  │
│                                                                          │
│  To close the dev server, hit Ctrl-C                                     │
│                                                                          │
│  Stuck? Visit us at https://svelte.dev/chat                              │
│                                                                          │
├──────────────────────────────────────────────────────────────────────────╯
│
└  You're all set!

clear
rm -rf test-sv
pnpx sv create --template minimal --types ts --no-install test-sv
┌  Welcome to the Svelte CLI! (v0.8.12)
│
◆  Project created
│
◇  What would you like to add to your project? (use arrow keys / space bar)
│  sveltekit-adapter
│
◇  Which SvelteKit adapter would you like to use?
│  cloudflare
│
◆  Successfully setup add-ons
│
◇  Project next steps ─────────────────────────────────────────────────────╮
│                                                                          │
│  1: cd test-sv                                                           │
│  2: pnpm i                                                               │
│  3: git init && git add -A && git commit -m "Initial commit" (optional)  │
│  4: pnpm run dev --open                                                  │
│                                                                          │
│  To close the dev server, hit Ctrl-C                                     │
│                                                                          │
│  Stuck? Visit us at https://svelte.dev/chat                              │
│                                                                          │
├──────────────────────────────────────────────────────────────────────────╯
│
└  You're all set!

cat test-sv/svelte.config.js
import adapter from '@sveltejs/adapter-cloudflare';
import { vitePreprocess } from '@sveltejs/vite-plugin-svelte';

const config = {
	preprocess: vitePreprocess(),
	kit: { adapter: adapter() }
};

[...]

PR #609, opened 2025-06-25
fix: preserve comments when parsing JS AST

teemingc:
fixes https://github.com/sveltejs/cli/issues/608

This PR changes the `parseScript` tooling to match the `esrap` implementation more closely. https://github.com/sveltejs/esrap/blob/4c305011b9536f60df4d0983ec50f46293b9ab8c/test/common.js#L39-L57 . It also adds a test to ensure comments are preserved.

### DIFF
--- a/.changeset/short-chicken-sell.md
+++ b/.changeset/short-chicken-sell.md
@@ -1,0 +1,5 @@
+---
+"sv": patch
+---
+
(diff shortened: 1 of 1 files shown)


manuel3108 (approved):
Perfect, thank you!

PR #618, opened 2025-06-28
fix(add): add `static` folder to `.prettierignore`

### LINKED ISSUE #607: add /static to prettierignore file
SvelteKit docs specify the static folder as "static assets that should be served as-is"
Formatting any js files is not complying with "as-is"
There's nothing stopping someone from using an external formatter for static files however I frequently encounter the issue where prettier formats my minified js file and increases files size (defeating the point of minification).
This fix is as simple as adding `/static` at the end of the prettier ignore file

manuel3108:
Closes #607 

This is probably debatable, but based on the presentation inside the docs, this does make sense in my opinion. I was currently unable to imagine any valid use-case where you would want to format your `static` folder.

### DIFF
--- a/.changeset/twenty-suits-listen.md
+++ b/.changeset/twenty-suits-listen.md
@@ -1,0 +1,5 @@
+---
+'sv': patch
+---
+
+fix(add): add `static` folder to `.prettierignore`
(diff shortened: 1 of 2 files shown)


bensgilbert:
@manuel3108, the `static/` pattern will ignore any directory named static anywhere in the project, alternatively the `/static/` pattern will ignore just the static folder at the root of the project (the intended behaviour). This will avoid confusion when (for whatever reason), the programmer has a folder named `static` somewhere else in their project. Would you mind if I made a pr changing it to `/static/`?

manuel3108:
Ahhh shit, you are obviously right. Happy to merge a pr

PR #620, opened 2025-06-28
chore: update `esrap`

manuel3108:
Figure out how to handle adding new comments during ast processing.

### DIFF
--- a/packages/core/tooling/index.ts
+++ b/packages/core/tooling/index.ts
@@ -14,6 +14,7 @@ import {
 } from 'postcss';
 import * as fleece from 'silver-fleece';
 import { print as esrapPrint } from 'esrap';
+import ts from 'esrap/languages/ts';
 import * as acorn from 'acorn';
 import { tsPlugin } from '@sveltejs/acorn-typescript';
 
@@ -47,19 +48,21 @@ export type {
 
 /**
  * Parses as string to an AST. Code below is taken from `esrap` to ensure compatibilty.
- * https://github.com/sveltejs/esrap/blob/9daf5dd43b31f17f596aa7da91678f2650666dd0/test/common.js#L12
+ * https://github.com/sveltejs/esrap/blob/920491535d31484ac5fae2327c7826839d851aed/test/common.js#L14
  */
-export function parseScript(content: string): TsEstree.Program {
+export function parseScript(content: string): {
+	ast: TsEstree.Program;
+	comments: TsEstree.Comment[];
+} {
 	const comments: TsEstree.Comment[] = [];
 
 	const acornTs = acorn.Parser.extend(tsPlugin());
 
-	// Acorn doesn't add comments to the AST by itself. This factory returns the capabilities to add them after the fact.
 	const ast = acornTs.parse(content, {
 		ecmaVersion: 'latest',
 		sourceType: 'module',
 		locations: true,
-		onComment: (block, value, start, end) => {
+		onComment: (block, value, start, end, startLoc, endLoc) => {
 			if (block && /\n/.test(value)) {
 				let a = start;
 				while (a > 0 && content[a - 1] !== '\n') a -= 1;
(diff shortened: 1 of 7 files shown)


manuel3108:
This is ready as soon as we merge https://github.com/sveltejs/esrap/pull/79 and update the referenced package here away from pkg.pr.new

sacrosanctic:
https://github.com/sveltejs/esrap/pull/90 was merged in lieu of https://github.com/sveltejs/esrap/pull/79 and released as [esrap@2.2.0](https://github.com/sveltejs/esrap/releases/tag/v2.2.0)

But I'm guessing it's not a simple package change anymore.

packages/core/tooling/index.ts
```diff
@@ -207,43 +213,55 @@ export function guessIndentString(str: string | undefined): string {
 	}
 }
 
-export function guessQuoteStyle(ast: TsEstree.Node): 'single' | 'double' | undefined {
```
jycouet, line 210: Any reason to remove `guessQuoteStyle` ?

packages/core/tooling/index.ts
```diff
@@ -207,43 +213,55 @@ export function guessIndentString(str: string | undefined): string {
...
+		this.#leading.set(node, list);
+	}
+
+	/** Add a comment that will appear after the given node */
+	addTrailing(node: TsEstree.Node, comment: CommentType): void {
+		const list = this.#trailing.get(node) ?? [];
+		list.push(comment);
+		this.#trailing.set(node, list);
+	}
+}
```
jycouet: Do we need all this "management"? Can't we use more the way esrap is expecting comments directly?

packages/core/tests/js/vite/with-satisfies/output.ts
```diff
@@ -18,23 +18,24 @@ const config = defineConfig({
 	plugins: [
 		// all plugins
 		examples,
+
```
jycouet, line 21: These new empty lines are a bit annoying to me!
I'm not sure who is the cupid here... I'll check :D

jycouet (commented):
I think I need to pull it locally to see a bit more :)
Will do in coming days.

Thx 🎉

packages/core/tooling/index.ts
manuel3108: I was under the impression that it got removed, but it isn't. Will need to check.

packages/core/tooling/index.ts
manuel3108: Yeah, this kinda bothers me as well. I wasn't sure which api i would like to expose to calees of `parseScript()` that actually makes sense and is similar to our current api. 

Similar goes for the `CommentState` class btw. This is necessary to make an internal abstraction for internal use. Otherwise we would need to publicly expose `#leading` and `#trailing` which i personally found a horrible idea. 

But happy to any improved api if you can find something!

packages/core/tests/js/vite/with-satisfies/output.ts
manuel3108: This is an `esrap` issue for sure. I think this only happens directly after comments, so this might be intentional. Without the new line you could technically read it as the command was supposed to be to be for all eleemnts. Which it might actually be in that case. In that case the comment should potentially be on `plugins` directly, but that's nothing we can change here.

packages/core/tooling/index.ts
manuel3108: The arg was just moved to a different place. Fixed.

packages/core/tooling/index.ts
jycouet: I think it's better now, let me know what do you think.

Also, I notice that we don't use `addTrailing` today. Should we expose it since today there is no usage?
No big deal.
Also, we could have a `remove()` to avoid non `@sveltejs/adapter-auto` limbo. But we can do this later

manuel3108:
Thanks! That's really an elegant change to the snapshots, expect maybe for `packages/cli/tests/snapshots/create-with-all-addons/vite.config.ts`. But that needs to be fixed inside `esrap` as you said.

packages/core/tooling/index.ts
manuel3108: I don't really mind. This now exposes the `leading` and `trailing` properties to the addons as well. We could argue that we don't need the two methods in that case and we could just remove them. 

But somehow we would need to make it clear to the user that he is not expected to push anything into `original` as that won't have any effect, unless the new comment contains `loc` properties, which it probably does not. 

Actually I'm also wondering if we could get rid of that weird unshit thing now. Will need to investigate on friday or something

jycouet: I'm fully with you... Not a big fan of the state.
I just pushed some updates and I'm WAY happier! What do you think?

manuel3108: This is perfect, thank you!

manuel3108:
This got way bigger than I initially expected. Glad we got it done.

Issue #561, opened 2025-05-17
Add `/drizzle` to the `.prettierignore` file


hyunbinseo:
`drizzle-kit generate` creates JSON with double space indentation.

Current Prettier configuration formats these files to use tabs.

Since they are generated files, they should not be formatted.

hyunbinseo:
@manuel3108 If `static/` is added in Prettier add-on, shouldn't `drizzle/` be added as well?

- https://github.com/sveltejs/cli/pull/618

PR #623, opened 2025-06-29
fix(add): add `/drizzle/` folder to `.prettierignore`

jycouet:
Closes: https://github.com/sveltejs/cli/issues/561


manuel3108 (approved):
Perfect, thank you!
I added the `runsAfter` call so that adding prettier and drizzle at the same time (`pnpx sv add drizzle prettier` yields the expected results.

Issue #199, opened 2023-12-18
Add minimum Node.js version warning to create-svelte


theetrain:
### Describe the problem

I noticed someone running into cryptic errors when running `npm create svelte@latest`; it didn't recognize the `await` keyword due to Node.js 12 being used.

### Describe the proposed solution

Have `create-svelte` check the installed Node.js version and fail on unsupported versions. Maybe this is as straightforward as adding an [`engines`](https://docs.npmjs.com/cli/v10/configuring-npm/package-json#engines) key to [package.json](https://github.com/sveltejs/kit/blob/466f445c616f6ef66d03e631515c75f49e20e126/packages/create-svelte/package.json).

### Importance

would make my life easier

PR #625, opened 2025-06-29
feat(add): Add minimum Node.js version warning if below 16.

jycouet:
Closes: https://github.com/sveltejs/cli/issues/199


packages/cli/utils/common.ts
```diff
@@ -4,6 +4,7 @@ import * as p from '@clack/prompts';
 import type { Argument, HelpConfiguration, Option } from 'commander';
 import { UnsupportedError } from './errors.ts';
 import process from 'node:process';
+import semver from 'semver';
```
manuel3108: There is already this code in the repo, that seems to be more elaborate. Also it does not require an additional dependency. We could probably extract that into a separate function.

https://github.com/sveltejs/cli/blob/58315358878a2c3a6bf48ce2ec68e75bc41f1a62/packages/addons/common.ts#L87-L89

packages/cli/utils/common.ts
```diff
@@ -75,6 +76,15 @@ type MaybePromise = () => Promise<void> | void;
 export async function runCommand(action: MaybePromise): Promise<void> {
 	try {
 		p.intro(`Welcome to the Svelte CLI! ${pc.gray(`(v${pkg.version})`)}`);
+
+		const isUnSupportedNodeVersion =
+			process.versions.node && semver.lt(process.versions.node, '16.0.0');
```
manuel3108: Any reason for choosing version 16? According to https://nodejs.org/en/about/previous-releases the latest non end of life version would be 20. Alternatively, we could align with kit and go with version 18.

benmccann: I think any version it runs on should be allowed. I'd at least support Node 18 if the CLI works there since kit supports it

jycouet: I looked into docs, found [v4-migration-guide - Node 16 or higher](https://svelte.dev/docs/svelte/v4-migration-guide#Minimum-version-requirements) but now, looking a second time, on kit side I found [migrating-to-sveltekit-2 - SvelteKit 2 requires Node 18.13 or higher](https://svelte.dev/docs/kit/migrating-to-sveltekit-2#Updated-dependency-requirements)

So maybe it should be `18.13` or higher ?

packages/cli/utils/common.ts
jycouet: I saw that `semver` was in the monorepo, so was thinking it's probably ok, but I agree that we should be carefull in adding deps.

Reading more about `process.versions.node`, it looks like it's always a `x.y.z`, no _"latest"_ or _"^18"_ or ... So, I'll look into the existing and extract into a separate function.

I'll let you know

packages/cli/utils/common.ts
manuel3108: Yeah, i think that sounds logical

packages/cli/utils/common.ts
jycouet: I extracted some stuff and put in `core` to be use everywhere later.

I kinda like what I did with:
```ts
const unsupported = minimumRequirement('18.3').for(process.versions.node);
```

Let's see if you like it as well :p

AdrianGonz97:
im not seeing any discussions for it, but couldn't we just add an `engines` field to `sv`'s `package.json` instead so node can do the version check for us? @manuel3108 @benmccann

manuel3108:
I think I did some research last week and I think I remembered that npx and similar are not checking the node version for whatever reason. I haven't tested it though.

.changeset/great-bags-smoke.md
```diff
@@ -0,0 +1,5 @@
+---
+'sv': patch
+---
+
+feat(add): Add minimum Node.js version warning if below 16.
```
benmccann: we don't put `(add)`, we start with lowercase letters, and code says 18.3 whereas this says 16
```suggestion
feat: print warning if using Node.js version below 18.3
```

manuel3108 (approved):
Thanks for your help! I have slightly adjusted the code to feel more like the rest of the codebase. But apart from that this was perfect!

jycouet:
> Thanks for your help! I have slightly adjusted the code to feel more like the rest of the codebase. But apart from that this was perfect!

I liked the style I did. But yes, it's ok to keep the repo style 🎉
Thx for improvements

PR #629, opened 2025-07-10
feat: update templates to vite 7 and vite-plugin-svelte 6

dominikg:
vite-plugin-svelte 6 has been  released. It is already present in the current peerDependency range of sveltekit

### DIFF
--- a/.changeset/giant-drinks-worry.md
+++ b/.changeset/giant-drinks-worry.md
@@ -1,5 +1,5 @@
 ---
-'sv': minor
+'sv': patch
 ---
 
 feat: update templates to vite 7 and vite-plugin-svelte 6


.changeset/giant-drinks-worry.md
```diff
@@ -0,0 +1,5 @@
+---
+'sv': minor
```
AdrianGonz97: should be a `patch` since we're in prerelease
```suggestion
'sv': patch
```

dominikg:
> Kit 2.9.0 doesn't include Vite 7 in its peerdeps, so that should be bumped up to wherever that peerdep was updated in Kit.

good catch, updated to 2.22.0

AdrianGonz97 (approved):
thank you!

Romanrazo (approved):

PR #633, opened 2025-07-15
feat: add `vite.addPlugin`

### LINKED ISSUE #632: Unable to process `vite.config.ts` while installing tailwindcss
Here's my error:
Error: Unable to process 'vite.config.ts'. Reason: Cannot read properties of undefined (reading 'length')
│      at Object.file (file:///private/tmp/bunx-503-sv@latest/node_modules/sv/dist/addons-Br9tic5t.js:1066:35)
│      at Object.run (file:///private/tmp/bunx-503-sv@latest/node_modules/sv/dist/addons-Br9tic5t.js:5437:6)
│      at runAddon (file:///private/tmp/bunx-503-sv@latest/node_modules/sv/dist/addons-Br9tic5t.js:1107:14)
│      at applyAddons (file:///private/tmp/bunx-503-sv@latest/node_modules/sv/dist/addons-Br9tic5t.js:1017:50)
│      at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
│      at async runAddCommand (file:///private/tmp/bunx-503-sv@latest/node_modules/sv/dist/bin.js:1071:79)
│      at async file:///private/tmp/bunx-503-sv@latest/node_modules/sv/dist/bin.js:836:25
│      at async runCommand (file:///private/tmp/bunx-503-sv@latest/node_modules/sv/dist/bin.js:565:3)
│

* This happens no matter which package manager I use.
* Based on the error, I think the issue is somewhere around here in the `applyAddons()` function https://github.com/sveltejs/cli/blob/ca3c8b57e46048e9eb4c02e7f807098e55a3ee41/packages/cli/lib/install.ts#L70

<details>
<summary>
Here's the config:
</summary>

```ts
// vite.config.ts
import { enhancedImages } from "@sveltejs/enhanced-img";
import { sveltekit } from "@sveltejs/kit/vite";
import path from "node:path";
import type { UserConfig } from "vite";
import { kitRoutes } from "vite-plugin-kit-routes";
import browserslist from "browserslist";
import { browserslistToTargets } from "lightningcss";
import devtoolsJson from "vite-plugin-devtools-json";

const $src = path.resolve(__dirname, "./src");
const $db = path.resolve($src, "./types/supabase.ts");
const $lib = path.resolve($src, "./lib");
const $routes = path.resolve($src, "./routes");
const $scripts = path.resolve($lib, "./scripts");
const $stores = path.resolve($scripts, "./stores");
const $images = path.resolve($lib, "./images");
const $components = path.resolve($lib, "./components");
const $containers = path.resolve($components, "./containers");
const $navigation = path.resolve($components, "./navigation");
const $classes = path.resolve($scripts, "./classes");

const config: UserConfig = {
	plugins: [
		devtoolsJson(),
		enhancedImages(),
		sveltekit(),
		kitRoutes({
			format: "variables",
		}),
	],
	server: {
		open: true,
	},
	resolve: {
		alias: {
			$db,
			$scripts,
			$routes,
			$stores,
			$images,
			$classes,
			$components,
			$containers,
			$navigation,
		},
	},

	optimizeDeps: {
		exclude: ["svelte-inspect-value"],
	},
	build: {
		sourcemap: true,
		target: "esnext",
		cssMinify: "lightningcss",
	},
	css: {
		transformer: "lightningcss",
		lightningcss: {
			targets: browserslistToTargets(browserslist("defaults, not ie 11")),
		},
	},
};

export default config;
```

</details>

jycouet:
fix #632

---

I just created a `helper.ts` & some tests, but it should probably move to another location ?
Maybe `addInArrayOfObject` & `exportDefaultConfig` could go in `common` ?
And `addPluginToViteConfig` is maybe a bit higher level ? Or it could go to common as well ?

I did the demo on `devtool` plugin, and the API looks nice! no ? @benmccann @manuel3108 let me know what do you think.


manuel3108:
Ignore my last comment, just found the issue. let me check then

jycouet:
haha, perfect 👍
I'll proceed, clean up & co and will go out of "draft" when ready to review.

jycouet:
Kinda beautiful 🫠

image

.changeset/tough-carrots-eat.md
```diff
@@ -0,0 +1,5 @@
+---
+'sv': patch
+---
+
+add `vite.addPlugin` to simplify adding a plugin on various vite config style.
```
manuel3108:
```suggestion
feat: add `vite.addPlugin` to simplify adding a plugin on various vite config styles
```

packages/core/tests/js/imports/named-import-simple/run.ts
```diff
@@ -0,0 +1,9 @@
+import { imports, type AstTypes } from '@sveltejs/cli-core/js';
```
manuel3108: Are the two `named-import-simple` files a leftover? Whats the difference to the non simple use-case?

packages/core/tests/js/vite/many-plugins/run.ts
```diff
@@ -0,0 +1,17 @@
...
+
+	const pLast = 'lastPlugin';
+	imports.addDefault(ast, { as: pLast, from: 'last-plugin' });
+	[...Array(10)].forEach((_, i) => {
+		vite.addPlugin(ast, { code: `${pLast}(${i})` });
+	});
+
+	const pFirst = 'firstPlugin';
+	imports.addNamed(ast, { imports: { [pFirst]: pFirst }, from: 'first-plugin' });
+	vite.addPlugin(ast, { code: `${pFirst}()`, mode: 'prepend' });
```
manuel3108: isn't this test actually testing the `prepend` `mode`? The test dosn't really have anything to do with many plugins, or am i missing something? Would suggest to only add 3-4 plugins overall and rename this test to be something like `prepend-mode`.

LMK if im missing something

packages/core/tests/js/imports/named-import-simple/run.ts
manuel3108: Ohhh, now i do get it. It's the array. We should probably extract into it's own PR, since there is lot's of places we can and should use this within the codebase.

Great idea, but I don't think it has anything to do with this PR

packages/core/tooling/js/imports.ts
```diff
@@ -59,11 +59,23 @@ export function addDefault(node: AstTypes.Program, options: { from: string; as:
 export function addNamed(
 	node: AstTypes.Program,
 	options: {
+		imports: Record<string, string> | string[];
 		from: string;
-		imports: Record<string, string>;
 		isType?: boolean;
 	}
 ): void {
+	if (Array.isArray(options.imports)) {
```
manuel3108: Same as above, would suggest extracting this into it's own PR

packages/core/tooling/js/vite.ts
```diff
@@ -0,0 +1,132 @@
...
+
+	// Handle wrapper functions (e.g., defineConfig({})) if ignoreWrapper is specified
+	let configObject: AstTypes.ObjectExpression;
+	if (ignoreWrapper && 'arguments' in rootObject && Array.isArray(rootObject.arguments)) {
+		// Check if this is the wrapper we want to ignore
+		if (
+			rootObject.type === 'CallExpression' &&
+			rootObject.callee.type === 'Identifier' &&
+			rootObject.callee.name === ignoreWrapper
+		) {
```
manuel3108: Would suggest inverting those two conditions and bailing out early to improve readability. Maybe even more, but those two should be pretty easy i think.

Apart from that the method looks really good!

manuel3108 (changes requested):
I really like this! This is way better than the previous implementation, imo.  Some minor stuff, but i think we are nearly good to go

packages/core/tests/js/imports/named-import-simple/run.ts
jycouet: 100% right. I will do another PR for this 👍
And revert here ;)

packages/core/tests/js/vite/many-plugins/run.ts
jycouet: The idea was to have many plugins & prepend yes.
I'll do less plugins 👍

packages/core/tests/js/imports/named-import-simple/run.ts
jycouet: Here is it: https://github.com/sveltejs/cli/pull/639

I will base this one on top

manuel3108:
Gonna review this after #639 is merged to reduce the diff.

jycouet:
Should now be in a better stage with the other PR merged. Now this PR is focusing on vite. 
- [x] feature
- [x] tests
- [x] update of 3 plugins (sooooo nice!)

---

I like a lot this repo :)

manuel3108 (approved):
Thank you!

PR #634, opened 2025-07-15
Improve (ba)sh usage, mainly in docs

fkobi:
Every time that I have read the default readme I did not like the fact that generic shell commands were labeled as bash and I have changed it. After the third time I have decided to fix it upstream.

Also there are some minor improvements that are bash-related.
Everything is explained in the commit messages.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -16,7 +16,7 @@ Please keep your pull requests focused to feature or issue. Focused smaller chan
 ## Preparing
 This is a monorepo, meaning the repo holds multiple packages. It requires the use of [pnpm](https://pnpm.io/). You can [install pnpm](https://pnpm.io/installation) with:
 
-```bash
+```sh
 npm i -g pnpm
 ```
 
@@ -25,7 +25,7 @@ Linux users, you will have to ensure 'sudo' is not required. See [docker post in
 
 `pnpm` commands run in the project's root directory will run on all sub-projects. You can checkout the code and install the dependencies with:
 
-```bash
+```sh
 cd cli
 pnpm install
 ```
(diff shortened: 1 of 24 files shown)


benmccann:
Is there any difference between the way `bash` and `sh` is rendered? Could you share a screenshot of before and after?

I would guess https://github.com/sveltejs/svelte.dev might need to be updated as well if one is really preferable to the other

fkobi:
> Is there any difference between the way bash and sh is rendered?

No, `bash` is (nearly) a superset of `sh`.
`sh` a general term for shell while `bash` is a specific shell with MANY extensions.
It makes sense to use the general one since no bash-specific features are used.

> I would guess...

I can do that once this gets merged!

manuel3108:
Just to understand: Does this make a difference how tools display the syntax highligting? I just compared the current docs page, with this one here and im unable to find any difference. Apart from that, I think we should be good to go here.

fkobi:
> Does this make a difference how tools display the syntax highligting?

It should not since `bash` is used on posix `sh` level.

manuel3108 (approved):
Personally I don't see any need to change this, but since it's not doing any harm, let's go for it.

fkobi:
> I would guess https://github.com/sveltejs/svelte.dev might need to be updated as well if one is really preferable to the other

I was going to open the PR but the PR template says
"""
If this is a documentation PR (i.e. changing content within `apps/svelte.dev/content/docs`), then this is the wrong repository to make those changes. The content in this folder is synced from other repositories. Therefore, these changes should be made in their respective repositories (at https://github.com/sveltejs/svelte or https://github.com/sveltejs/kit, or example).
"""

I guess I'll go to https://github.com/sveltejs/svelte and https://github.com/sveltejs/kit to edit them.

What should be done about the blog posts though?

benmccann:
The blog posts and templates live in the svelte.dev repo

Issue #167, opened 2024-10-17
combine next steps prompt for `create`


AdrianGonz97:
maybe we can have something like this:
```shell
◇  Next steps ──────────────────────────────────────────────────────────────╮
│                                                                            │
│  Getting started:                                                          │
│    1: cd foo-bar                                                           │
│    2: git init && git add -A && git commit -m "Initial commit" (optional)  │
│    3: pnpm run dev -- --open                                               │
│                                                                            │
│  To close the dev server, hit Ctrl-C                                       │
│                                                                            │
│  Integrations:                                                             │
│    drizzle:                                                                │
│      - You will need to set DATABASE_URL in your production environment    │
│      - Run npm run db:push to update your database schema                  │
│                                                                            │
│    lucia:                                                                  │
│      - Run npm run db:push to update your database schema                  │
│      - Visit /demo/lucia route to view the demo                            │
│                                                                            │
│  Stuck? Visit us at https://svelte.dev/chat                                │
│                                                                            │
├────────────────────────────────────────────────────────────────────────────╯
│
└  You're all set!
```

jycouet:
Today we have the left part.
What do you think about the right one ?

<img width="1797" height="656" alt="Image" src="https://github.com/user-attachments/assets/71f28ee5-b735-4496-b252-debb6e128429" />

Or

<img width="619" height="371" alt="Image" src="https://github.com/user-attachments/assets/364a0cce-3eb2-4e75-9912-730c99718dfa" />

AdrianGonz97:
looks great! I would probably rename `Integrations` to `Add-ons` since (i believe) this issue was created before we settled on the name 'add-ons'. 

as for it being underlined or italicized, i would be in favor of it being italicized (or maybe we can just keep it normal too. the indentation could be sufficient enough to indicate it's a title). underlining text tends to make stuff feel more cluttered/busy imho

jycouet:
<img width="624" height="358" alt="Image" src="https://github.com/user-attachments/assets/8877a254-89b9-48d9-bf42-bf72d0ff939d" />

<img width="624" height="309" alt="Image" src="https://github.com/user-attachments/assets/56a9402c-e765-4f0b-9d54-c4b1b1166d70" />

After a few tests, I think I like `gray`

And managing `Add-ons` && `Add-on` 😉

benmccann:
This looks nice overall

[...]

PR #637, opened 2025-07-18
feat: combine next steps prompt for `create`

jycouet:
Fixes: #167

## On create

image

## On add

image

### DIFF
--- a/.changeset/blue-islands-eat.md
+++ b/.changeset/blue-islands-eat.md
@@ -1,0 +1,5 @@
+---
+'sv': patch
+---
+
+feat: combine next steps prompt for `create`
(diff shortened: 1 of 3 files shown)


AdrianGonz97:
made a few tweaks to give it a bit more breathing room and have it be more symmetrical in `create`:

img

(italics TBD. not sure if i really like them yet, but it does provide that _slight_ amount of needed contrast)

benmccann:
What about a colon instead of italics?

AdrianGonz97:
> What about a colon instead of italics?

gave that a whirl earlier and i wasn't really a fan of how many colons there were now that the titles have them:
img

AdrianGonz97:
small alternative if we want to spice it up a bit

img

jycouet:
🧡 Emojis

PR #639, opened 2025-07-20
chore: allow passing an array of import names to imports.addNamed

jycouet:


packages/addons/paraglide/index.ts
```diff
@@ -96,7 +96,7 @@ export default defineAddon({
 			const vitePluginName = 'paraglideVitePlugin';
 			imports.addNamed(ast, {
 				from: '@inlang/paraglide-js',
-				imports: { paraglideVitePlugin: vitePluginName }
+				imports: [vitePluginName]
```
manuel3108, line 99: I would be careful with this one, as changing the value of `vitePluginName` to something else might cause unexpected results.

manuel3108 (changes requested):
This is basically perfect, thank you!

packages/addons/paraglide/index.ts
jycouet: It's actually this exact line that triggered me to this this PR :D

```ts
// this today:
const vitePluginName = 'paraglideVitePlugin';
imports: { paraglideVitePlugin: vitePluginName }

// is doing actually: 
imports: { paraglideVitePlugin: 'paraglideVitePlugin' }

// And with the new syntax it's: 
imports: ['paraglideVitePlugin']

// That can be written like: 
const vitePluginName = 'paraglideVitePlugin';
imports: [vitePluginName]
```

All in all, if you change `vitePluginName`, yes it's not working... But everything is not working ^^
A few lines under: `const pluginFunctionCall = functions.createCall({ name: vitePluginName, args: [] });` will also not work.

You would suggest to do ?
```ts
const vitePluginName = 'paraglideVitePlugin';
  imports.addNamed(ast, {
    from: '@inlang/paraglide-js',
    imports: ['paraglideVitePlugin']
    // ...
```

I see what you mean, but I'm not sure about the suggestion.

Let me know :)

manuel3108: Oh yeah, didn't think it through. Everything is going to break as you said, so let's go with this!

Issue #218, opened 2023-12-15
Skeleton template improvements


jycouet:
### Describe the problem

It's not really a problem, it's more an improvement (QoL improvement!).
I would like to tweak a few things in the `skeleton` default template.

### Describe the proposed solution

Here is the list I have in mind:

1/ Having a `+layout.svelte` (with the h1)
It's showing the fonctionnality,

2/ moving `fav ico` to /src/lib/assets
To show immutable assets

3/ Add a `robot.txt` to keep a static folder

4/ Move playwright files beside routes like `page.e2e.(j|t)s`

### Alternatives considered

My current alternative is to do all these tweaks each time I create a skeleton. _(I do it a lot of repro in different context)_

### Importance

would make my life easier

### Additional Information

I will be happy to do it. I would like the blessing before sending the PR :)

jycouet:
Hello here 👋

I would be happy to do some of the points (or all 😊). Let me know.

manuel3108:
Relates #203 
2. and 3. sound like good ideas to me. 
for 1. i wouldn't know what content we should put in there in the skeleton template that would make sense, therefore I think we shouldn't bother about that one.
4. comes down to personal preference.  Although you could make the point that we already put the vitest test files besides the routes / components. But still unsure about this one.

jycouet:
Time to do 2 and 3 then 👍

For 1 - We could have `<h1>Welcome to SvelteKit</h1>` in Layout
and `<p>Visit <a href="https://svelte.dev/docs/kit">svelte.dev/docs/kit</a> to read the documentation</p>` in Page.
Maybe even have some link in Layout if we have demos. (Lucia?!)
We can defer to another PR or to never :p (I'll come back to it later ;))

For 4 - Yes, it's all about preference. (I don't like a `test` folder with the "same" structure, we already have the structure!)
We can defer to another PR or to never :p (I'll come back to it later ;))

jycouet:
**For 2/**
In my "work work" app, I get these green `_app/immutable/assets/xxx`
![Image](https://github.com/user-attachments/assets/1ac7ecfb-ac4b-4173-8d9b-b3057e9d2380)

So, I started with:
```html
<script>
  import favicon from '$lib/assets/favicon.png'
</script>

<svelte:head>
	<link rel="icon" href={favicon} />
</svelte:head>
```
It's not producing a `_app/immutable/assets` but directly this: 
```ts
[...]

PR #643, opened 2025-07-25
Skeleton template improvements

jycouet:
Link with https://github.com/sveltejs/cli/issues/218
- [ ] Having a +layout.svelte (with the h1)
- [x] fav in `/src/lib/assets` (inline or immutable with `assetsInlineLimit` flag in vite.config)
- [x] Add a `robots.txt` to keep a static folder
- [ ] Move playwright files beside routes like page.e2e.(j|t)s


jycouet:
## Layout

I would still like to add a `+layout.svelte` with: 
```html
<script>
	import favicon from '$lib/assets/favicon.svg';

	let { children } = $props();
</script>

<svelte:head>
	<link rel="icon" href={favicon} />
</svelte:head>

<h1>Welcome to SvelteKit</h1>
{@render children?.()}
```
And update `page.svelte.test.ts` to `layout.svelte.test.ts`

## Tests
Today when we have `vitest` & `playwright`, we have: 
- `src/routes/page.svelte.test.ts`
- `src/demo.spec.ts`
- `e2e/demo.test.ts`

So, some `spec.ts`, some `test.ts`, some in `routes`, some not. I think it's good to guide more here.

We could do:
- ~~`src/routes/page.svelte.test.ts`~~ `src/routes/page.svelte.spec.ts`
- `src/demo.spec.ts`
- `e2e/demo.test.ts`

=> To have `.spec.ts` in all vitest files
=> To have `.test.ts` in all playwright files
=> e2e tests are not in the routes folder as they can describe a flow involving multiple pages.

Let me know your feelings :p

packages/create/templates/minimal/static/robots.txt
```diff
@@ -0,0 +1,3 @@
+# https://www.robotstxt.org/robotstxt.html
```
benmccann: People can find docs, ask an LLM, etc. as necessary so I don't really feel like we need to include this here. I do think it'd be helpful to explain what this does though as it's not immediately clear from the syntax if it bans or allows all bots if you're not familiar with robots.txt
```suggestion
# allow crawling everything by default
```

packages/create/templates/minimal/src/routes/+page.svelte
```diff
@@ -1,2 +1,10 @@
+<script lang="ts">
+	import favicon from '$lib/assets/favicon.svg';
+</script>
+
+<svelte:head>
+	<link rel="icon" href={favicon} />
+</svelte:head>
```
manuel3108: Shouldn't this be part of the layout? As soon as you create a second page, this will behave differently than it does now.

manuel3108:
I'm not totally sure I agree with moving the h1 in the layout, but I definitely agree for the favicon. So the layout file would still be present, just without and filled with the favicon. I think that's a good compromise.

I do agree with renaming that one vitest `.test.ts` file, it's totally inconsistent. Thanks for catching that. Also agree with your other conclusions!

manuel3108:
Ohh, and any reason you are only "linking" the issue in the pr description and not "closing" it?

packages/create/templates/minimal/src/routes/+page.svelte
jycouet: Yes, even a better reason to have a layout. 👍

jycouet:
> I'm not totally sure I agree with moving the h1 in the layout, but I definitely agree for the favicon. So the layout file would still be present, just without and filled with the favicon. I think that's a good compromise.

I like this good compromise 😁

> I do agree with renaming that one vitest .test.ts file, it's totally inconsistent. Thanks for catching that. Also agree with your other conclusions!

I updated only the test file name to match `spec`. Let's still keep the glob `src/**/*.{test,spec}.{js,ts}`. RIght ?

> Ohh, and any reason you are only "linking" the issue in the pr description and not "closing" it?

Not really... It's just it was not addressing all points. Now that it's the case, I updated it ^^

manuel3108:
> I updated only the test file name to match spec. Let's still keep the glob src/**/*.{test,spec}.{js,ts}. RIght ?

Yes, exactly.

Thinking out loud, any reason for not applying those changes to the other templates?
- `robots.txt` is missing in `library` and has a different content in `demo`. I think it would make sense everywhere
- `favicon.svg`. Moving into `lib` in `library` probably does not make sense, as it would probably be published as part of the library (although, maybe tree-shaking detects this?). Making the same changes we made here to `demo` makes sense I think.

Didn't make those changes myself, as I'm not 100% sure that's the correct way to do it. Let me know what you think!

jycouet:
> * `robots.txt` is missing in `library` 

I think it's not really needed there. The first goal is to public a `lib`, not really deploying the website. No ? (Or maybe for the doc?!) Hummm, I don't know. I would not do it I think.

> `robots.txt` has a different content in `demo`. 

Yes, we could align with what we set here ?

> * `favicon.svg`. Moving into `lib` in `library` probably does not make sense, as it would probably be published as part of the library (although, maybe tree-shaking detects this?). 

Yeah, let's not move `favicon.svg` in `library`.

> `favicon.svg` Making the same changes we made here to `demo` makes sense I think.

Yeah, it's probably good to do this change

**EDIT**: Humm I'm not sure!
It already has `import github from '$lib/images/github.svg';` showing immutability.
Also it's a folder `images` and not `assets`. What do you suggest ?

---

I'll update with your good thinking & my first opinions. Let's see after ;)

manuel3108 (approved):
I agree with your last comment, let's leave it as is.

Thank you!

hyunbinseo:
The updated favicon code should be reverted. This issue is still open:

- https://github.com/sveltejs/svelte/issues/15337

jycouet:
Oh wow, that's wild ! Didn't know about this :/

Is it better to revert the full PR or do another PR to bring back the old favicon only ? 👀

hyunbinseo:
If the `favicon.svg` stays in the `static` directory, is there a need to keep a generic `robots.txt` file?

In my personal template `svelte-kitty`, I am excluding certain predefined routes, making it meaningful.

```txt
User-agent: *
Disallow: /login
Disallow: /unsupported
````

https://github.com/hyunbinseo/svelte-kitty/blob/main/packages/create-svelte-kitty/template/static/robots.txt

jycouet:
If `favicon.svg` is going back, yes, there is no reason to keep this "empty" `robots.txt`.
But `favicon.svg` is temporary going back, right ? So, not sure about the `robots.txt`.

(The original idea was to have something in this folder so that people know that it exist)

Let me know, and I'll do.

Nice kitty!

hyunbinseo:
Yes, `/static/favicon.svg` is temporary.

It should be reverted to Vite inline asset import in the long term.

> (The original idea was to have something in this folder so that people know that it exist)

This was my whole point with `svelte-kitty` and its structure. Then I stepped on the `<svelte:head>` issue.

jycouet:
https://github.com/sveltejs/cli/pull/656

PR #644, opened 2025-07-25
:bug: FIX: Project steps alignment

jycouet:
Oups! This PR is fixing this:

image

### DIFF
--- a/.changeset/common-signs-grin.md
+++ b/.changeset/common-signs-grin.md
@@ -1,0 +1,5 @@
+---
+'sv': patch
+---
+
+fix: Project steps alignment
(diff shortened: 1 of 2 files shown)


.changeset/common-signs-grin.md
```diff
@@ -0,0 +1,5 @@
+---
+'sv': patch
+---
+
+fix: Project steps alignment
```
benmccann: At least in SvelteKit we usually start changelog entries with a lowercase. Not sure if Manuel has been doing the same here or not
```suggestion
fix: align project steps
```

manuel3108: I do

PR #647, opened 2025-07-26
feat: `vitest` require assertions

manuel3108:
This came up in the maintainers chat. It's regarding this config: https://vitest.dev/config/#expect-requireassertions

If there is a test without an assertion, it will currently pass unnoticed. But then it's not a real test.

Paolo contacted the `vitest` team, and they said this was set as default a long time ago, to ease the migration process from Jest and so on. Since `sv` is used to create new projects and not to migrate old Jest projects, we are safe to set such useful constraints to the user.

This was mainly expected to change the `vitest` add-on, but I also change our internal `vitest` configs to avoid this in this repo as well.

### DIFF
--- a/packages/create/vitest.config.ts
+++ b/packages/create/vitest.config.ts
@@ -5,9 +5,6 @@ export default defineProject({
 	test: {
 		name: 'create',
 		include: ['test/*.ts'],
-		retry: env.CI ? 3 : 0,
-		expect: {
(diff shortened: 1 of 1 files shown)


manuel3108:
Ok, let's not enforce this on `migrate` it causes to many errors and is not something we are currently actively working on.
Edit: and the same for create

PR #648, opened 2025-07-26
feat: add `db:generate` command to `drizzle` add-on

### LINKED ISSUE #484: Improve drizzle-orm SQLite default template

manuel3108:
Closes #484

### DIFF
--- a/.changeset/silly-waves-sit.md
+++ b/.changeset/silly-waves-sit.md
@@ -2,4 +2,4 @@
 'sv': patch
 ---
 
-feat: add `db:generate` command to `drizzle` add-on
+feat(drizzle): add `db:generate` script to `package.json`


.changeset/silly-waves-sit.md
```diff
@@ -0,0 +1,5 @@
+---
+'sv': patch
+---
+
+feat: add `db:generate` command to `drizzle` add-on
```
AdrianGonz97:
```suggestion
feat(drizzle): add `db:generate` script to `package.json`
```

AdrianGonz97 (approved):

PR #650, opened 2025-07-26
feat(cli): sv add --no-git-check

### LINKED ISSUE #240: Explain 'preconditions' in docs
The `sv-add` docs list the `--no-preconditions` option but with no clue as to what preconditions are and why I'd want to disable them (and when it would be safe to do so)

Better still would be to get rid of the option altogether in my view — either they're preconditions (in which case you can't disable them without breaking stuff) or they're not

jycouet:
Closes #240

I've set changeset to `minor` as a flag is removed. _(Not sure if it's really minor, could be patch also)_


manuel3108:
Will need to check this in details another time, but makes sense at a first glance.
Also minor totally makes sense here, as I know a few people actually using the cli args to generate one off projects.

packages/cli/commands/add/preconditions.ts
```diff
@@ -3,54 +3,60 @@ import { promisify } from 'node:util';
...
-			{
-				name: 'unsupported add-ons',
-				run: () => {
-					const reasons = addons.flatMap((a) =>
-						addonSetupResult[a.id].unsupported.map((reason) => ({ id: a.id, reason }))
-					);
-
-					if (reasons.length === 0) {
-						return { success: true, message: undefined };
+export function getCleanPreconditions(cwd: string, gitCheck: boolean) {
```
manuel3108:
```suggestion
export function getCleanCWDPreconditions(cwd: string, gitCheck: boolean) {
```
Clean without anything else sounds strange. But adding `cwd` doesn't look better in my opinion. May I don't have good ideas today.

Since we are only returning one precondition here, should we also change the signature? I cannot see a way when we would return to multiple preconditions here. Same goes for the other method in this file

jycouet: Yeah, I tried different stuff... I'm not sure also what's the best (name & signature).

Returning array, is kinda cool for evolution & handling higher level. But I agree, we don't use it today! :)

jycouet:
> I do think this broke something. `pnpm create vite` && `pnpx https://pkg.pr.new/sveltejs/cli/sv@650 add drizzle` runs successfully for some reason, even it shouldn't.

I just tried and got:
image

manuel3108:
You are right, I was in a nested svelte-kit application, in which case what I saw even makes sense. Can confirm it's working as expected.

packages/cli/commands/add/preconditions.ts
manuel3108: According to some internet statistics (AI) I think naming those methods `verifyCleanWorkingDirectory` and `verifyUnsupportedAddons` would make a lot of sense. Also, it would help us to get away from the term `preconditions`.

What do you think?

jycouet: Make sense yes 👍

.changeset/tangy-deer-fetch.md
```diff
@@ -4,3 +4,4 @@
 
 - feat(cli): remove `--no-preconditions` option from `sv add`
 - feat(cli): add `--no-git-check` option to `sv add`. With this flag, even if some files are dirty, no prompt will be shown
+- feat(cli): type `Precondition` was replaced with `Verification`
```
manuel3108: Do we need to add a changeset for that? I don't think this type was publicly exposed, even if it couldn't be used since community adders are not exposed.

jycouet: It's exported from core, but yes let's not add it, it's not for users at this point.

manuel3108 (approved):
Thank you!

If you want, I think there is even further potential for improvement. Like why is the unsupported check even part of the verifications, if it always throws on fails and the catch is located somewhere else?

As you said in the issue, the unsupported check is not optional, it's a must that it passes. The git check is a can / should. 

Should we totally get rid of the concept of verifications in the next step? And check both things independently?

hyunbinseo:
I've been using `--no-preconditions` and would like to migrate:

https://github.com/hyunbinseo/svelte-kit-templates/blob/v0.8.21/index.js#L24

Is it a simple rename? If so, specifying this in the change-log would be helpful

> The current preconditions is a check to see if you have a dirty git repo https://github.com/sveltejs/cli/issues/240#issuecomment-2442407079

manuel3108:
It's a bit more, but in your case you can treat it as a rename. 

The other check we are currently doing is that you cannot add an addon that does not support pure svelte apps (like drizzle) to a pure svelte app (without kit).

But I don't think that matters in your example.

PR #656, opened 2025-07-31
revert: favicon location

jycouet:
While [Failing to set `<svelte:head>` element with `getAttribute is not a function` svelte#15337](https://github.com/sveltejs/svelte/issues/15337) is open, let's revert the location of the favicon.

### DIFF
--- a/.changeset/metal-insects-smile.md
+++ b/.changeset/metal-insects-smile.md
@@ -2,4 +2,4 @@
 'sv': patch
 ---
 
-revert: favicon location to `/static/favicon.svg`
+revert: favicon location to `/static/favicon.svg` for minimal template


.changeset/metal-insects-smile.md
```diff
@@ -0,0 +1,5 @@
+---
+'sv': patch
+---
+
+revert: favicon location to `/static/favicon.svg`
```
hyunbinseo: Shouldn't this mention the minimal template as the previous PR did?

manuel3108:
This is not happening for me. I did mention in the maintainers chat that we should merge a fix for the original issue inside svelte, instead of reverting this, as this is not happening for everyone and seems to be related to browser extensions or something.

Let's hold off from this for a few days.

hyunbinseo:
I believe a starter template should be conservative.

It is a bug in the wild, and it could happen to the customer - not to the developer.

If the sole purpose is to demonstrate Vite's asset import, maybe an image loading comparison could be added?

jycouet:
As it's the minimal template, it should probably not be too bloated. I would not add a separate image loading, as this belongs more to a demo.

I and my customers didn't experience this bug (sentry setup), I can imagine that it's fairly limited _(but I don't know, just feeling)_.
At the same time, I agree that it should be conservative, but showing best practices!

All in all, I don't know! :)

The best would of course be to fix the underlying bug.

PR #660, opened 2025-08-02
chore: improved storybook prefetching

manuel3108:
Relates https://github.com/sveltejs/cli/pull/622
Relates https://github.com/sveltejs/svelte-ecosystem-ci/pull/32

In direct comparison to #622 this has the disadvantage, that this also run's locally. We could check via the `CI` env variable, but I don't think that would help to reproduce pipeline failures locally.  On the other hand, this is an improvement in comparison to #622 because it allows our tests to be run everywhere, without requiring explicit setup. This is in example the case in `svelte-ecosystem-ci` (https://github.com/sveltejs/svelte-ecosystem-ci/pull/32)

If this works the way I think it should (I will check in the pipeline), I would personally consider this as a slight improvement.

### DIFF
--- a/packages/addons/_tests/storybook/test.ts
+++ b/packages/addons/_tests/storybook/test.ts
@@ -12,8 +12,10 @@ const { test, variants, prepareServer } = setupTest({ storybook, eslint });
 let port = 6006;
 
 beforeAll(() => {
-	// prefetch the storybook cli to reduce fetching errors in tests
-	execSync('pnpx create-storybook@latest --version');
+	if (process.env.CI) {
+		// prefetch the storybook cli during ci to reduce fetching errors in tests
+		execSync('pnpx create-storybook@latest --version');
+	}
 });
 
 test.for(variants)(
--- a/packages/addons/_tests/storybook/test.ts
+++ b/packages/addons/_tests/storybook/test.ts
@@ -1,5 +1,7 @@
 import process from 'node:process';
+import { execSync } from 'node:child_process';
 import { expect } from '@playwright/test';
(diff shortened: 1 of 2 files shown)


packages/addons/_tests/storybook/test.ts
```diff
@@ -9,6 +11,11 @@ const { test, variants, prepareServer } = setupTest({ storybook, eslint });
 
 let port = 6006;
 
+beforeAll(() => {
+	// prefetch the storybook cli to reduce fetching errors in tests
+	execSync('pnpx create-storybook@latest --version');
+});
```
AdrianGonz97: adding the prefetch to the workflow file instead of directly in the test code was intentional in #622 so that we're not prefetching storybook _every time we run `pnpm test` locally_. if we're going to continue with this setup, we should at least wrap it in a `process.env.CI` check so that it only prefetches in CI

manuel3108: I did understand that. Background here is https://github.com/sveltejs/svelte-ecosystem-ci/pull/32, which requires us to also add that prefetching in ecosystem-ci. And remembering that / keeping that in sync is probably going to be hard.

I'm fine by adding a CI check. 

> every time we run pnpm test locally
Only true if you always run the addons project. I rarely run them locally, as they are super computationally intensive and my laptop just doesn't like them. But that might be a me problem.

AdrianGonz97: > I'm fine by adding a CI check.

should be good to merge then once that's added!

AdrianGonz97 (approved):

PR #662, opened 2025-08-03
feat(cli): `sv create --from-playground`

### LINKED ISSUE #602: `sv create` allow download from playground
Idea `sv create --from-playground=id-or-hash`

Instead of downloading & saving, extracting, installing deps, we could do all of that directly

manuel3108:
Closes #602 

This is a draft. As of now, the cli understands the new arg, but does nothing with it.
The only interesting part at this point is the tests, but they are obviously still a work in progress.

Once this is done `sv create --from-playground https://svelte.dev/playground/hello-world` should just work and create a new app locally with the minimal template. It will assum App.svelte is the main component, and instantiate that in `routes/+page.svelte` of the created project. 

Users are free to add any addons if they want to.

### DIFF
--- a/packages/create/playground.ts
+++ b/packages/create/playground.ts
@@ -1,0 +1,78 @@
+export function validatePlaygroundUrl(link: string): boolean {
+	try {
+		const url = new URL(link);
+		if (url.hostname !== 'svelte.dev' || !url.pathname.startsWith('/playground/')) {
+			return false;
+		}
+
+		const { playgroundId, hash } = extractPartsFromPlaygroundUrl(link);
+		return playgroundId !== undefined || hash !== undefined;
+
+		// eslint-disable-next-line @typescript-eslint/no-unused-vars
+	} catch (_) {
+		// new Url() will throw if the URL is invalid
+		return false;
+	}
+}
+
+export function extractPartsFromPlaygroundUrl(link: string): {
+	playgroundId: string | undefined;
+	hash: string | undefined;
+} {
+	const url = new URL(link);
+	const [, playgroundId] = url.pathname.match(/\/playground\/([^/]+)/) || [];
+	const hash = url.hash !== '' ? url.hash.slice(1) : undefined;
+
+	return { playgroundId, hash };
+}
+
+type PlaygroundData = {
+	name: string;
+	files: Array<{
+		name: string;
+		content: string;
+	}>;
+};
+export async function downloadFilesFromPlayground({
+	playgroundId,
+	hash
+}: {
+	playgroundId?: string;
+	hash?: string;
+}): Promise<PlaygroundData> {
+	let data = [];
+	// forked playgrounds have a playground_id and an optional hash.
+	// usually the hash is more up to date so take the hash if present.
+	if (hash) {
+		data = JSON.parse(await decode_and_decompress_text(hash));
+	} else {
+		const response = await fetch(`https://svelte.dev/playground/api/${playgroundId}.json`);
+		data = await response.json();
+	}
+
+	// saved playgrounds and playground hashes have a different structure
+	// therefore we need to handle both cases.
+	const files = data.components !== undefined ? data.components : data.files;
+	return {
+		name: data.name,
+		files: files.map((file: { name: string; contents: string; source: string }) => {
+			return {
+				name: file.name,
+				content: file.source || file.contents
+			};
+		})
+	};
+}
+
+// Taken from https://github.com/sveltejs/svelte.dev/blob/ba7ad256f786aa5bc67eac3a58608f3f50b59e91/apps/svelte.dev/src/routes/(authed)/playground/%5Bid%5D/gzip.js#L19-L29
+/** @param {string} input */
+async function decode_and_decompress_text(input: string) {
+	const decoded = atob(input.replaceAll('-', '+').replaceAll('_', '/'));
+	// putting it directly into the blob gives a corrupted file
+	const u8 = new Uint8Array(decoded.length);
+	for (let i = 0; i < decoded.length; i++) {
+		u8[i] = decoded.charCodeAt(i);
+	}
+	const stream = new Blob([u8]).stream().pipeThrough(new DecompressionStream('gzip'));
+	return new Response(stream).text();
+}
--- a/.changeset/fancy-towns-argue.md
+++ b/.changeset/fancy-towns-argue.md
@@ -1,0 +1,5 @@
+---
+'sv': patch
+---
+
+feat(cli): `sv create --from-playground`
--- a/packages/cli/commands/create.ts
+++ b/packages/cli/commands/create.ts
@@ -43,7 +43,8 @@ const OptionsSchema = v.strictObject({
 	),
 	addOns: v.boolean(),
 	install: v.union([v.boolean(), v.picklist(AGENT_NAMES)]),
-	template: v.optional(v.picklist(templateChoices))
+	template: v.optional(v.picklist(templateChoices)),
+	fromPlayground: v.optional(v.string())
 });
 type Options = v.InferOutput<typeof OptionsSchema>;
 type ProjectPath = v.InferOutput<typeof ProjectPathSchema>;
@@ -56,6 +57,7 @@ export const create = new Command('create')
 	.option('--no-types')
 	.option('--no-add-ons', 'skips interactive add-on installer')
 	.option('--no-install', 'skip installing dependencies')
+	.option('--from-playground <string>', 'create a project from the svelte playground')
 	.addOption(installOption)
 	.configureHelp(common.helpConfig)
 	.action((projectPath, opts) => {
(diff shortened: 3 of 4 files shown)


jycouet:
This is exciting 🤩


I have this one that could show a few things: https://svelte.dev/playground/770bbef086034b9f8e337bab57efe8d8
- `App.svelte` => `/src/routes/+page.svelte`
- `Footer.svelte` =>  `/src/routes/Footer.svelte` easy like this, all imports are still working.
- `import * as changeCase from "change-case";` => Need to add this in `package.json`
- Maybe the main layout can be tuned to have a "Playground" style with a link to the playground.
localhost_5174_

<details><summary>+layout.svelte example</summary>
<pre>
<script lang="ts">
	import favicon from '$lib/assets/favicon.svg';

	let { children } = $props();
</script>

<svelte:head>
	<link rel="icon" href={favicon} />
</svelte:head>

<div class="layout">
	<nav class="navbar">
		<div class="nav-left">
			<a href="/" class="svelte-icon">
				Svelte
			</a>
			<p class="title">Playground</p>
		</div>
		<div class="nav-right">
			<a
				href="https://svelte.dev/playground/770bbef086034b9f8e337bab57efe8d8"
				class="playground-link"
				target="_blank"
				rel="noopener noreferrer"
			>
				Cases <span aria-hidden="true" style="margin-left:0.25em;">↗</span>
			</a>
		</div>
	</nav>

	<main class="content">
		{@render children?.()}
	</main>
</div>

<style>
	:global(body) {
		margin: 0;
	}

	.layout {
		min-height: 100vh;
		background-color: #1c1e22;
	}

	.navbar {
		display: flex;
		justify-content: space-between;
		align-items: center;
		padding: 1rem 2rem;
		background-color: #23272f;
		border-bottom: 1px solid #2d3139;
	}

	.nav-left {
		display: flex;
		align-items: center;
		gap: 1rem;
	}

	.svelte-icon {
		display: flex;
		align-items: center;
		text-decoration: none;
		transition: opacity 0.2s ease;
	}

	.svelte-icon:hover {
		opacity: 0.8;
	}

	.title {
		color: #ffffff;
		font-size: 1.5rem;
		font-weight: 600;
		margin: 0;
	}

	.nav-right {
		display: flex;
		align-items: center;
	}

	.playground-link {
		color: #ffffff;
		text-decoration: none;
		font-weight: 500;
		padding: 0.5rem 1rem;
		border-radius: 0.375rem;
		transition: background-color 0.2s ease;
	}

	.playground-link:hover {
		background-color: #2d3139;
	}

	.content {
		padding: 2rem;
		color: #ffffff;
	}
</style>
</pre>
</details>

manuel3108:
> I have this one that could show a few things: https://svelte.dev/playground/770bbef086034b9f8e337bab57efe8d8

This is perfect, and i was able to get it fully working from the tests.
The only thing remaining here is code optimization and making it work from the core cli, instead from the `crete` tests. 



> Maybe the main layout can be tuned to have a "Playground" style with a link to the playground.

Not sure, I'm a fan of this. The current "download app" button does nothing like this. Also, I'm not sure if this provides any additional value to the user, or just causes styling problems because of the added css. We can always add this later if users want this.

jycouet:
Great progress 🚀🚀🚀 Nice

>> Maybe the main layout can be tuned to have a "Playground" style with a link to the playground.

> Not sure, I'm a fan of this. The current "download app" button does nothing like this. Also, I'm not sure if this provides any additional value to the user, or just causes styling problems because of the added css. We can always add this later if users want this.

The cool thing about a layout like this is that: 
- You can have a link back to the playground. so you know where it's coming from.
- with css per component, it should not affect anything
- you can escape from it by just deleting one file
- we can feel that we are "home" in our playgound... locally (with similar style)

Yes, I love this idea 😜
But it can come later too with a flag in or out ;)

Anyhow, very nice job

manuel3108:
Updated pr description, this is working now. Let me know what you think.

packages/create/playground.ts
```diff
@@ -0,0 +1,151 @@
+import path from 'node:path';
+import * as fs from 'node:fs';
+import { parseJson, parseScript, parseSvelte } from '@sveltejs/cli-core/parsers';
+import * as js from '@sveltejs/cli-core/js';
+
+export async function write_playground_files(url: string, cwd: string): Promise<void> {
+	if (!validatePlaygroundUrl(url)) throw new Error(`Invalid playground URL: ${url}`);
```
jycouet: I think that this will be a "common" mistake in the url formating.

Today I got: 

Image

Maybe we can help more with a nice message like: 
`You typed: "toto", it should be a valid playground url like ...`

jycouet: Maybe even not exiting the `cli` when this happen to have the chance to paste a valid `url` ?

jycouet (commented):
The result is like this: 
Image

Comments in random order:
- I find it strange to have the "Template content" coming from a playground. Maybe `+page.svelte` should have ONLY `<App />`, nothing else ?
- I still like the playground layout idea :p
- Maybe a default folder name could be proposed "svelte-playground-slug-name-of-playgound" ?
- Maybe the `readme.md` could mention the origin of the playground ?
- If some extra deps are needed, prompt it in CLI and add it to readme.md ?

It's probably nitpicking, because it's working well already. 
Great work 🎉🎉🎉

packages/create/playground.ts
manuel3108: This exact one should only trigger when you call `create` programmatically. But you are absolute right, this should be fixed.

manuel3108:
> I find it strange to have the "Template content" coming from a playground. Maybe `+page.svelte` should have ONLY `<App />`, nothing else ?

Agree, modified.

> Maybe a default folder name could be proposed "svelte-playground-slug-name-of-playgound" ?

That would leave us to handle all that characters that cannot be directly translated into a directory name. Also, most of the repls I would personally download would have the name "Hello world (edited)" so I don't think this is super important. Can always be added later without any problems. 


> Maybe the `readme.md` could mention the origin of the playground ?

I understand, but I'm hesitant about this. Adding it near the top / somewhere in the middle is going to be hard, adding it at the end is easy, but probably nobody would notice. Would leave this open and wait for user feedback.

> * If some extra deps are needed, prompt it in CLI and add it to readme.md ?

Totally agree and implemented

--- 

The only major thing left is detecting and using the svelte version parameter from the playground url. And more testing. Let me know what you think about the code. Also if you have a 3-5 more real playground urls to test this again, that would be great.

jycouet:
I did a bit of cleanup in my playgrounds: 
- https://svelte.dev/playground/853e2d01359d455798e13c1fdaddff30?version=5.38.3
- https://svelte.dev/playground/99c2cd640b1445dca98ab740781118a0?version=5.38.3
- https://svelte.dev/playground/65a404127d0b4c728515e2de5033fa39?version=5.38.3

Nothing really special.

> > Maybe the readme.md could mention the origin of the playground ?

> I understand, but I'm hesitant about this. Adding it near the top / somewhere in the middle is going to be hard, adding it at the end is easy, but probably nobody would notice. Would leave this open and wait for user feedback.

On top top, it's not that easy ?

manuel3108:
I just added a final change to support the `version` query parameter from the playground url. Example: https://svelte.dev/playground/hello-world?version=pr-16711

If everybody agrees, this should be good to go now.

jycouet (approved):
Looks great 🚀

You want to silently launch without docs? Or you want to add the new option directly?

_BTW, how to run docs locally with new changes? 👀_

manuel3108:
> You want to silently launch without docs? Or you want to add the new option directly?

Uupps, good catch. Is the description I added to detailed? That reminded me that we were considering adding that command to playground itself. Going to start a thread in discord in a second.

> BTW, how to run docs locally with new changes? 👀

Don't! If you really want to here are the docs: https://github.com/sveltejs/svelte.dev/?tab=readme-ov-file#setup
Just push the docs changes to a pr, and the docs-bot will add a comment where you can see your changes. In this case, it's here: https://svelte-dev-git-preview-cli-662-svelte.vercel.app/docs/cli/overview

documentation/docs/20-commands/10-sv-create.md
```diff
@@ -12,6 +12,16 @@ npx sv create [options] [path]
 
 ## Options
 
+### `--from-playground <url>`
+
+Create a SvelteKit project from a Svelte playground URL. This downloads the playground code, automatically detects and installs external dependencies, and sets up a complete SvelteKit project structure. External dependencies found in import statements are automatically added to `package.json` and can be installed during project setup.
```
jycouet: - I feel that you explain 2 times `external dependencies`.
- Maybe emphasize the ability to download multiple files

I would propose:
```suggestion
Create a SvelteKit project from a Svelte Playground URL. This downloads all playground files, detects external dependencies, and sets up a complete SvelteKit project structure with everything ready to go.
```

Feel free to ignore :)

manuel3108:
Let's go! If you feel like adding this to the playground, feel free

Issue #672, opened 2025-08-15
`sv add` creates a `vite.config.ts` even if `vite.config.js` exists


Rich-Harris:
It looks like `sv add` assumes it should write to a `vite.config.ts` file if it detects that it's in a TypeScript project, even if `vite.config.js` already exists. To reproduce:

```bash
npx sv create vite-config-repro --template=minimal --types=ts --no-add-ons --install=pnpm
cd vite-config-repro
mv vite.config.ts vite.config.js
npx sv add devtools-json --install=pnpm
ls vite.config.*
```

PR #673, opened 2025-08-15
Typescript project doesn't mean `ts` files

jycouet:
closes: #672

### DIFF
--- a/.changeset/sour-wings-taste.md
+++ b/.changeset/sour-wings-taste.md
@@ -2,4 +2,4 @@
 'sv': patch
 ---
 
-fix: `vite.config.*` file detection works for both .js and .ts variants
+fix(cli): `vite.config.*` file detection works for both .js and .ts variants
--- a/.changeset/sour-wings-taste.md
+++ b/.changeset/sour-wings-taste.md
@@ -1,0 +1,5 @@
+---
+'sv': patch
+---
+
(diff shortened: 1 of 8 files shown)


.changeset/sour-wings-taste.md
```diff
@@ -0,0 +1,5 @@
+---
+'sv': patch
+---
+
+fix: `vite.config.*` file detection works for both .js and .ts variants
```
manuel3108:
```suggestion
fix(cli): `vite.config.*` file detection works for both .js and .ts variants
```

packages/cli/commands/add/workspace.ts
```diff
@@ -23,6 +23,8 @@ export async function createWorkspace({
 	const viteConfigPath = path.join(resolvedCwd, commonFilePaths.viteConfigTS);
 	let usesTypescript = fs.existsSync(viteConfigPath);
```
manuel3108, line 24: I do wonder why we chose to detect typescript based on the vite config file in the past. But since it's working i'm not going to touch it.

manuel3108 (approved):
This works perfectly, thanks!

packages/cli/commands/add/workspace.ts
jycouet: Told myself the same thing 😅

PR #679, opened 2025-08-30
docs: add quotes to avoid error executing command with add-on options

totemcaf:
Example source code in documentation show command without quotes in add-on commands.

When running that example code in a Linux terminal, the command fails. Adding the quotes it runs ok.

To avoid developers trying to find the problem, this PR improves documentation.


jycouet:
Let's close this PR as https://github.com/sveltejs/cli/pull/681 was merged.
Thx for the report.

PR #680, opened 2025-08-31
feat(cli): `sv add remote-functions`

### LINKED ISSUE #641: `create`: update template to use remote functions and async

jycouet:
Closes #641

## Topics
- [ ] new deps `scule` (for casing). Ok for it ?
- [ ] I did just demo with `query` to avoid another lib for `standard-schema`. Let me know if you think we should do more
- [ ] I'm not sure about `withDemo` and `demo`. Let's see what you think :)
- [x] #685 to be merged first. Initial note: added `object.ensureNestedProperty`. Should I do it in another PR ? _(Maybe there is also a bit to rethink... `overrideProperty`, `overrideProperties`, `addProperties` and now `ensureNestedProperty`... The 2 last could be managed by the 2 firsts with some small adjustements)_ 
- [ ] Maybe it should not be an "official" add-on ? One day it will not be experimental, so only the demo will be left inside ? do we have a story for "deprecated add-on" ? We will keep the code, but not list them as official ? _(let's speak)_

### DIFF
--- a/packages/addons/remote-functions/index.ts
+++ b/packages/addons/remote-functions/index.ts
@@ -4,7 +4,7 @@ import {
 	defineAddonOptions,
 	isVersionUnsupportedBelow
 } from '@sveltejs/cli-core';
-import { common, exports, object, type AstTypes } from '@sveltejs/cli-core/js';
+import { common, exports, object } from '@sveltejs/cli-core/js';
 import { parseScript } from '@sveltejs/cli-core/parsers';
 import { addToDemoPage } from '../common.ts';
 import { titleCase, kebabCase, pascalCase } from 'scule';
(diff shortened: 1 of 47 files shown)


packages/addons/remote-functions/index.ts
```diff
@@ -0,0 +1,210 @@
+import {
+	colors,
+	defineAddon,
+	defineAddonOptions,
+	isVersionUnsupportedBelow
+} from '@sveltejs/cli-core';
+import { common, exports, object, type AstTypes } from '@sveltejs/cli-core/js';
+import { parseScript } from '@sveltejs/cli-core/parsers';
+import { addToDemoPage } from '../common.ts';
+import { capitalCase, kebabCase, pascalCase } from 'change-case';
```
43081j: i would use [scule](https://github.com/unjs/scule) for this - its slightly smaller and has stronger types/inference (though we may not make use of that here)

packages/addons/remote-functions/index.ts
```diff
@@ -0,0 +1,210 @@
...
+
+		// experimental flags
+		sv.file('svelte.config.js', (content) => {
+			const { ast, generateCode } = parseScript(content);
+
+			const { value: config } = exports.createDefault(ast, { fallback: object.create({}) });
+
+			// Handle kit.experimental.remoteFunctions
+			const kitConfig = config.properties.find(
+				(p) => p.type === 'Property' && p.key.type === 'Identifier' && p.key.name === 'kit'
```
43081j: for this you can probably use `object.property`:

```ts
const kitConfig = object.property(config, {
  name: 'kit',
  fallback: object.create({
    experimental: {
      remoteFunctions: true
    }
  }
});
```

you might be able to do this for each of the nested objects to simplify away the `if`/`else` chains

though you'll still want to check if the `kitConfig.type === 'ObjectExpression'` and probably overwrite it if its not

jycouet: I did a step more and added a util that will help a lot the readability.
What do you think like this ?

packages/addons/remote-functions/index.ts
jycouet: I don't have much preference, I'll follow your advice. 
Thank you for looking 👍

packages/addons/remote-functions/index.ts
43081j: Neat solution! That'll come in useful elsewhere I'm sure

packages/addons/sveltekit-adapter/index.ts
```diff
@@ -80,34 +80,23 @@ export default defineAddon({
...
+			// remove leading comments from the `adapter` property
+			const kitConfig = object.property(config, {
+				name: 'kit',
+				fallback: object.create({}) // we know that is was created before
+			});
+			const adapterProp = object.property(kitConfig, {
+				name: 'adapter',
+				fallback: object.create({}) // we know that is was created before
+			});
+			adapterProp.leadingComments = [];
```
43081j: FYI this isn't quite the same, as `object.property` returns the _value_ node

before, `adapterProp` will have been the property node itself

i do wonder if we should have two methods here:
- `object.propertyValue` (what `object.property` currently is)
- `object.property` (does what `object.property` does now, but returns the `property`)

internally, `property` is always `AstTypes.Property`, so we could have something like this:

```ts
export function property(
	node: AstTypes.ObjectExpression,
	options: {
		name: string;
		fallback: AstTypes.Expression | AstTypes.Identifier;
	}
): AstTypes.Property {
  // ...
  return property;
}

export function propertyValue<T extends AstTypes.Expression | AstTypes.Identifier>(
	node: AstTypes.ObjectExpression,
	options: {
		name: string;
		fallback: AstTypes.Expression | AstTypes.Identifier;
	}
): AstTypes.Property {
  return property(node, options).value;
}
```

edit: aha you noticed! nevermind :D

jycouet: I'll resolve this one and speak `object` here #685

manuel3108:
> * Maybe it should not be an "official" add-on ? One day it will not be experimental, so only the demo will be left inside ? do we have a story for "deprecated add-on" ? We will keep the code, but not list them as official ? _(let's speak)_

That's the reason when creating https://github.com/sveltejs/cli/issues/641 i was mainly thinking about updating the `demo` template to use some of those features. I think it would be great to have an `experimental` addon in the future to easily provide more easily accessible demos for features in development. As sson as those features become "mainstream" I think we should have enough docs that the demo should become irrelevant. That would also allow us to further promote stuff like the recently merged [observability stuff](https://svelte.dev/blog/sveltekit-integrated-observability)

As of now, I'm not sure if i think creating an addon is the right way to do it. What do the others think?

jycouet:
Once Remote Functions is official, the demo project will need to be updated to replace `load` (https://github.com/sveltejs/kit/discussions/13897#discussioncomment-14326112)

Actually, we should probably just prepare this! (and pkg will be also available in the meantime (with experimental))
=> I think it's the best! ATM

---

I think that more than `experimental` addon, community addon could have helped here.
`npx sv add @sveltejs/sv-add-on-remote-functions` (like this, it's not impacting official add-on)

manuel3108:
> I think that more than `experimental` addon, community addon could have helped here.

You are absolutely right. I think this is the way. Closing this for now, I think we can do the other stuff in a seperate PR

Issue #651, opened 2025-07-27
Review `dependencies` vs `devDependencies`


benmccann:
Most things are in `devDependencies`, but there are a handful in `dependencies` and I think a few of them could probably live in `devDependencies` as well. A good way to test might be to try to put them in `dependencies` and then build an app with `adapter-node` and see if it still runs.

jycouet:
After a quick test I think that red ones could move to `devDependencies` yes
```diff
"dependencies": {
--  "@inlang/paraglide-js": "^2.0.0",
--  "@libsql/client": "^0.14.0",
    "@node-rs/argon2": "^2.0.2",
--  "@oslojs/crypto": "^1.0.1",
--  "@oslojs/encoding": "^1.1.0",
--  "drizzle-orm": "^0.40.0"
}
```

manuel3108:
I think this was just something that happend by following the respective docs.

For drizzle: https://orm.drizzle.team/docs/get-started/sqlite-new#step-1---install-required-packages
For inlang: https://inlang.com/m/dxnzrydw/paraglide-sveltekit-i18n/getting-started#getting-started

But if everything still works as devDep's, sure let's move them

PR #682, opened 2025-08-31
chore: change dependencies to devDependencies in addons

jycouet:
Closes: #651

### DIFF
--- a/.changeset/fancy-cars-stand.md
+++ b/.changeset/fancy-cars-stand.md
@@ -1,0 +1,5 @@
+---
+'sv': patch
+---
+
+chore: change some dependencies to devDependencies in addons
--- a/.changeset/fancy-cars-stand.md
+++ b/.changeset/fancy-cars-stand.md
@@ -2,4 +2,4 @@
 'sv': patch
 ---
 
-chore: change some dependencies to devDependencies in addons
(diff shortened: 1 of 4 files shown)


.changeset/fancy-cars-stand.md
```diff
@@ -0,0 +1,5 @@
+---
+'sv': patch
+---
+
+chore: change some dependencies to devDependencies in addons
```
manuel3108:
```suggestion
chore(addons): change some dependencies to devDependencies
```

jycouet:
> Any reason for not moving `@node-rs/argon2` to dev in `lucia`? Seems to be working for me locally.
> 
> Otherwise, this looks perfect.

I think it's working in development (`pnpm dev`), but you should not be able to build (`pnpm build`) if it's in `devDependencies` therefore it has to stay in `dependencies`

manuel3108:
`build` and `preview` work for me in this case

jycouet:
Ha, found it!

You need `@node-rs/argon2` in `dependencies` if you have `@sveltejs/adapter-node`. If not, the build will fail.

manuel3108 (approved):
Can confirm!
Thank you!

PR #686, opened 2025-09-02
Builder pattern for addon options

jycouet:


packages/core/addon/options.ts
```diff
@@ -20,31 +20,30 @@ export type NumberQuestion = {
...
 };
 
 export type MultiSelectQuestion<Value = any> = {
 	type: 'multiselect';
-	default: Value[];
-	options: Array<{ value: Value; label?: string; hint?: string }>;
+	default: Value;
+	options: Array<{ value: string; label?: string; hint?: string }>;
 	required: boolean;
 };
```
jycouet: `Value[]` and `Value` was confusing here I think. And the default `Value` should already be an array type.

I wish it would be possible to build up the Array from (small) `value`, but I think that it's not possible.

packages/addons/drizzle/index.ts
```diff
@@ -12,40 +12,40 @@ const PORTS = {
...
-		default: 'postgres.js',
+		default: 'postgres.js' as 'postgres.js' | 'neon',
 		options: [
 			{ value: 'postgres.js', label: 'Postgres.JS', hint: 'recommended for most users' },
 			{ value: 'neon', label: 'Neon', hint: 'popular hosted platform' }
 		],
 		condition: ({ database }) => database === 'postgresql'
-	},
-	mysql: {
+	})
```
jycouet, line 37:
```ts
condition: ({ database })
          //  ^ is not a string! 
          //    but really 'postgresql' | 'mysql' | 'sqlite' 🥳
```

benmccann:
I'm not sure I'm in love with the name `prepareAddonOptions`. How about `addonOptions().option().build()` or `newAddonOptions().add().build()`?

jycouet:
> I'm not sure I'm in love with the name `prepareAddonOptions`. How about `addonOptions().option().build()` or `newAddonOptions().add().build()`?

What about:
`defineAddonOptions().add().build()` ?

benmccann:
Works for me

packages/core/addon/options.ts
43081j: won't this have been because the default value of a multi-select can be multiple values?

e.g. `default: ["foo", "bar"]` when my `options` are `[{value: "foo"}, {value: "bar"}]`

the `Value` should be the type of an item in the array, rather than the type of the array itself, i would've thought

43081j:
this works so much better than the object-style definitions before

i think we'll do the same in clack eventually 😅

packages/core/addon/options.ts
jycouet: Good catch ! I'll look later 👀

.changeset/early-trees-go.md
```diff
@@ -0,0 +1,5 @@
+---
+'sv': patch
+---
+
+internal(core): change `defineAddonOptions({ /*config */ })` to `defineAddonOptions().add('key', { /*config */ }).build()` in order to provide better type safety.
```
manuel3108:
```suggestion
chore(core): change `defineAddonOptions({ /*config */ })` to `defineAddonOptions().add('key', { /*config */ }).build()` in order to provide better type safety.
```

packages/addons/drizzle/index.ts
manuel3108: WOW, this is great!

manuel3108 (commented):
I.... don't understand Typescript. Why do we need to make it more complicated to make it actually work... (especially referring to the internal `createOptionBuilder`). 

If you ignore the internals though, this is sick. It fixes the issue without changing anything major. Apart from the open comment, I think this should be good to go.

packages/core/addon/options.ts
jycouet: Exactly, the default value is of type `string[]` or `('foo' | 'bar')[]`, so it's already an array. Value is alreay an array.

Doing `Value[]` would do `string[][]`

I added a comment in the code for my future me also ^^

jycouet:
> If you ignore the internals though, this is sick. It fixes the issue without changing anything major. Apart from the open comment, I think this should be good to go.

Spent a few fun hours on this! héhé.
Last comment done now IMO

Issue #690, opened 2025-09-05
`sv add lucia`: Cannot read properties of undefined (reading 'dependsOn')


manuel3108:
```
PS D:\cli> pnpx sv add lucia --cwd ../cli-temp
┌  Welcome to the Svelte CLI! (v0.9.2)
│
◇  The lucia add-on requires drizzle to also be setup. Include it?
│  Yes
│
■  TypeError: Cannot read properties of undefined (reading 'dependsOn')
│      at runAddCommand (file:///C:/Users/manue/AppData/Local/pnpm-cache/dlx/4606e059aa0298da97dee039e2d80711d61420ed6c469b7c1bd1e9da56666de4/19919cbeeb6-13d54/node_modules/.pnpm/sv@0.9.2/node_modules/sv/dist/bin.js:986:43)
│      at process.processTicksAndRejections (node:internal/process/task_queues:105:5)
│      at async file:///C:/Users/manue/AppData/Local/pnpm-cache/dlx/4606e059aa0298da97dee039e2d80711d61420ed6c469b7c1bd1e9da56666de4/19919cbeeb6-13d54/node_modules/.pnpm/sv@0.9.2/node_modules/sv/dist/bin.js:843:25
│      at async runCommand (file:///C:/Users/manue/AppData/Local/pnpm-cache/dlx/4606e059aa0298da97dee039e2d80711d61420ed6c469b7c1bd1e9da56666de4/19919cbeeb6-13d54/node_modules/.pnpm/sv@0.9.2/node_modules/sv/dist/bin.js:568:3)
│
└  Operation failed.
```

Manually selecting lucia from `sv add` works just fine. No clue when / how this got introduced.

PR #691, opened 2025-09-05
fix(add): allow passing add-on as argument that dependsOn another

jycouet:
closes: #690

### DIFF
--- a/packages/cli/commands/add/index.ts
+++ b/packages/cli/commands/add/index.ts
@@ -411,11 +411,6 @@ export async function runAddCommand(
 	}
 
 	// add inter-addon dependencies
-	addonSetupResults = setupAddons(
-		selectedAddons.map(({ addon }) => addon),
-		workspace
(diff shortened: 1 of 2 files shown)


packages/cli/commands/add/index.ts
```diff
@@ -411,6 +411,11 @@ export async function runAddCommand(
 	}
 
 	// add inter-addon dependencies
+	addonSetupResults = setupAddons(
+		selectedAddons.map(({ addon }) => addon),
+		workspace
+	);
```
manuel3108: I think this is redundant? At least locally, it works without this. The place below also makes much more sense I think

jycouet: Yes, you are right. Thank you for spotting it.

AdrianGonz97 (approved):
looks good! just made a small tweak to the ordering so that it flows a tad better

PR #692, opened 2025-09-05
fix: adjust add-on option types

AdrianGonz97:
Noticed some spots that could be improved from #686. Mainly lessening the need to use `as const` at the builder's usage site, prettifying the resulting option type by flattening it, and other minor tweaks.

### DIFF
--- a/packages/cli/lib/install.ts
+++ b/packages/cli/lib/install.ts
@@ -173,7 +173,7 @@ async function runAddon({ addon, multiple, workspace }: RunAddon) {
 		devDependency: (pkg, version) => {
 			dependencies.push({ pkg, version, dev: true });
 		},
-		pnpmBuildDependendency: (pkg) => {
+		pnpmBuildDependency: (pkg) => {
 			pnpmBuildDependencies.push(pkg);
 		}
 	};
--- a/packages/core/addon/config.ts
+++ b/packages/core/addon/config.ts
@@ -20,7 +20,7 @@ export type Scripts<Args extends OptionDefinition> = {
 };
 
 export type SvApi = {
-	pnpmBuildDependendency: (pkg: string) => void;
+	pnpmBuildDependency: (pkg: string) => void;
 	dependency: (pkg: string, version: string) => void;
 	devDependency: (pkg: string, version: string) => void;
 	execute: (args: string[], stdio: 'inherit' | 'pipe') => Promise<void>;
--- a/packages/core/addon/config.ts
+++ b/packages/core/addon/config.ts
@@ -85,29 +85,32 @@ export type Verification = {
 	run: () => MaybePromise<{ success: boolean; message: string | undefined }>;
 };
 
+type Prettify<T> = {
+	[K in keyof T]: T[K];
+} & unknown;
+
 // Builder pattern for addon options
-export type OptionBuilder<T extends OptionDefinition = Record<string, any>> = {
-	add<K extends string, Q extends Question<T & Record<K, Q>>>(
+export type OptionBuilder<T extends OptionDefinition> = {
+	add<K extends string, const Q extends Question<T & Record<K, Q>>>(
 		key: K,
 		question: Q
 	): OptionBuilder<T & Record<K, Q>>;
-	build(): T;
+	build(): Prettify<T>;
 };
 
-export function defineAddonOptions(): OptionBuilder<Record<string, any>> {
-	return createOptionBuilder({} as Record<string, any>);
+// Initializing with an empty object is intended given that the starting state _is_ empty.
+// eslint-disable-next-line @typescript-eslint/no-empty-object-type
+export function defineAddonOptions(): OptionBuilder<{}> {
+	return createOptionBuilder({});
 }
 
-function createOptionBuilder<T extends OptionDefinition>(options: T = {} as T): OptionBuilder<T> {
+function createOptionBuilder<const T extends OptionDefinition>(options: T): OptionBuilder<T> {
 	return {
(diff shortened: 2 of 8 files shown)


packages/core/addon/options.ts
```diff
@@ -17,20 +17,20 @@ export type NumberQuestion = {
...
+	options: Array<{ value: Value; label?: string; hint?: string }>;
 };
 
-export type MultiSelectQuestion<Value extends any[] = string[]> = {
+export type MultiSelectQuestion<Value> = {
 	type: 'multiselect';
-	default: Value;
-	options: Array<{ value: string; label?: string; hint?: string }>;
+	default: NoInfer<Value[]>;
+	options: Array<{ value: Value; label?: string; hint?: string }>;
```
AdrianGonz97, line 29: this was some-what reverted to improve inference of the array type since we want to infer the values from the `options` rather than from the `default`, given that all possible values should be defined in the options

packages/core/addon/options.ts
```diff
@@ -54,6 +61,6 @@ export type OptionValues<Args extends OptionDefinition> = {
 				: Args[K] extends SelectQuestion<infer Value>
 					? Value
 					: Args[K] extends MultiSelectQuestion<infer Value>
-						? Value // as the type of Value should already be an array (default: [] or default: ['foo', 'bar'])
-						: never;
+						? Value[]
+						: 'ERROR: The value for this type is invalid. Ensure that the `default` value exists in `options`.';
```
AdrianGonz97, line 65: added this as `never` wasn't causing any type errors in the conditionals. not exactly sure why that was happening, but adding this error message instead gives a better type error. 

ideally though, we'd want the type error to end up in the option's definition instead [like here](https://www.typescriptlang.org/play/?#code/KYDwDg9gTgLgBDAnmYcDKwA2wDGMCKArsAM4wCWEAdgDwBqAhpsQHxwC8cA3gFACQAE2AAzBoUwwAXHAByEAJJVhwKPSasA3PwhgK1EtICCUKA0Q0ucAG7rg0xs2Aa4mBgCMsAfmlko5KgDmzgAW-jDecL7+AXAAvixasVo8woRUeJRUkVi4MDQ4+vAOrAAUAI7EZJnSGNh4RKR6tMXALACUNTn1lU1qjmy8cHBQwDCEUFkVjZmJPDwFVGRwARAQAgCMHNl1MCW8fIIiYhLSAOQAtogkZZinADRw-Hw6TQZwANpPfJY2jmeQZACI2utwerg8mDOAAUIICRmh8AAZU5xO5fH62M6XEH3FzuLBnACyiARyNR6OsmLgpxB5BgwFx4IJ1NJdIZcSeAF00bE2nMFks3AwNkNOCQurt9odROIpNShQJcU8Xpk3p8Dt9KX9qQCYEDSDdGfjIdSYXDgKSUbE0RqMdqLldDWDjUSSUirTaDnbiGdafSjRCzqz-RyDtyeLy5nMAPTRuDnWXkcU7OagSCwBDIVCExO1XINKrUPqVNicfZCGUnWQKJQqYukd6chLaXSqowmMwWLU+uAtEjOJmYCJRQIhMLDmB+QJxZtJOapdJNeOJ5O5fKFXu2Egsco9apwHMSch57rTIt99rSQ8UE8EPfnrcDHhDEZjCZwKaFqizeYblZrAAmLYEyPVc8D2J4K2OOV3gdHEHlOXV9RxcMDhVfRpHVL1uzsHVYT1YEnTxQNTXw-VLXJW0cKxR1QWI5lTmJCjrQpX4expG42QDBjg3ZWIuR5Plf0WeAFSAzgQIoMDJUgo5ZUwuCiNOU5UOeVsMI+ViqUQsjCLowdoV0i13Uo7C2NwxT9JdakmJMliqPM31OP9Z0SNOXirQEiM2iAA), but i couldn't get it to work for the time being.

packages/core/addon/config.ts
```diff
@@ -85,29 +85,32 @@ export type Verification = {
 	run: () => MaybePromise<{ success: boolean; message: string | undefined }>;
 };
 
+type Prettify<T> = {
+	[K in keyof T]: T[K];
+} & unknown;
+
```
AdrianGonz97, line 91: this bit flattens the built option's type to something more readable

before:
img

after:
img

packages/core/addon/options.ts
jycouet: That's way better yes 👍
_(I even started a followup branch, but you did everything better already)_

packages/core/addon/options.ts
jycouet: Yeah, and the message will be better than the comment thx

packages/cli/lib/install.ts
```diff
@@ -173,7 +173,7 @@ async function runAddon({ addon, multiple, workspace }: RunAddon) {
 		devDependency: (pkg, version) => {
 			dependencies.push({ pkg, version, dev: true });
 		},
-		pnpmBuildDependendency: (pkg) => {
+		pnpmBuildDependency: (pkg) => {
```
jycouet, line 176: haha, well spotted!

jycouet (commented):
Thank you for this follow up. 

I think that we can also revert the `typedEntries` in `packages/addons/tailwindcss/index.ts` and bring back the array of plugin (similar to `packages/addons/sveltekit-adapter/index.ts` ?

May I do it in this branch ? or ?

AdrianGonz97:
> May I do it in this branch ? or ?

go for it!

jycouet (approved):
Looks great to me 🚀
Thx

Issue #688, opened 2025-09-04
Show args used


jycouet:
After `npx sv add` you get 👇

<img width="628" height="420" alt="Image" src="https://github.com/user-attachments/assets/43a03d04-c75c-48b5-9405-305bc625adae" />

Would be cool to have a flag `--show-args` so that I will be able to use it next time?

```bash
# here something like
npx sv add drizzle=database:SQLite+client:libSQL --no-install
```

_name is debatable :)_

Conduitry:
Having the CLI output how to run it next time to get it to do unattended whatever it is you selected in the menus this time sounds like a useful feature generally. I don't think we need to hide that behind a flag.

jycouet:
I was thinking flag because it could be long and _ugly_, but I can start without a flag first and see.

jycouet:
I think that https://github.com/sveltejs/cli/issues/689 would be better to do first.

Today, I played around and managed:
<img width="870" height="415" alt="Image" src="https://github.com/user-attachments/assets/549dd405-426a-4d51-8c6c-bf7df8515dc9" />

Maybe in this format it would be better?
```sh
npx sv create jyc \
  --template minimal \
  --types ts \
  --no-install \
  --no-add-ons && \
npx sv add -C jyc \
  prettier \
  vitest=usages:unit,component \
  eslint \
  drizzle=database:sqlite+sqlite:libsql \
  --no-git-check \
  --install pnpm
```

All in all, it's taking a lot of space. Let's see!

benmccann:
yeah, the flag might be nice to keep the UI clean. I wonder if it should actually create the project in that case or not

jycouet:
> yeah, the flag might be nice to keep the UI clean. I wonder if it should actually create the project in that case or not

Could be a flag `--dry-run` that would output the full cmd & not create anything. _(I think I like that 👍)_

Conduitry:
I was thinking about `--dry-run` earlier, but then I found myself wondering whether I would expect that to check anything or do anything besides just print out the command line options. I think dry runs typically tell you more precisely what of is that would happen, and maybe run some checks to see whether everything looks okay. I'm not sure what that would mean in this case.

That said, I don't know that the noise of printing out the options on every run would be that bad. People aren't going to be running this terribly often, I don't think. I don't know what the noise would be getting in the way of.

manuel3108:
> Maybe in this format it would be better?

[...]

PR #695, opened 2025-09-08
feat(cli): Create & add one go

### LINKED ISSUE #689: create & add in one go
Today, `create` is one command and `add` is another command. 
It's almost hidden for users when you do a `create` as the cli will proceed with the add process just after.

But, I see 2 issues: 
1/ When you do a `create` and cancel the process at the add-ons selection, the project is created. Since it's cancelled, the project should not be created.
2/ We can't, from one cli launch, replicate the full selection of args. We have to do something like: 
```
npx sv create jyc --template minimal --types ts --no-install --no-add-ons && npx sv add -C jyc prettier vitest=usages:unit,component eslint drizzle=database:sqlite+sqlite:libsql --no-git-check --install pnpm
```
Do the first cli with `--no-install --no-add-ons`, then the second cli, ... Probably not the easiest for user perspective.

kizivat:
This should resolve #689 .

1. solves an issue where canceling the CLI while selecting add-ons, the project is still crated
2. it adds `--add` variadic option to `create` enabling specifying add-ons during project creation

Point 1. is solved by separating the question prompting and applying add-ons in the `add` script. Both are reused in the `create` action handler.

I'm creating this as a draft PR to receive general direction before solving `// TODO_ONE:`s.

### DIFF
--- a/.changeset/cruel-pets-obey.md
+++ b/.changeset/cruel-pets-obey.md
@@ -1,0 +1,5 @@
+---
+'sv': patch
+---
+
+fix(tailwind): update vscode setting `files.associations` to `tailwindcss`
--- a/.changeset/quick-comics-tell.md
+++ b/.changeset/quick-comics-tell.md
@@ -1,0 +1,5 @@
+---
+'sv': patch
+---
+
+feat(mcp): include an `AGENTS.md` or similar when using the `mcp` addon
--- a/.changeset/ripe-lines-beg.md
+++ b/.changeset/ripe-lines-beg.md
@@ -1,0 +1,5 @@
+---
+'sv': patch
+---
+
+feat(vitest): when `add vitest` is used within a project that uses vitest@3, the addon will display some next steps to finalize the migration to vitest@4
--- a/.changeset/wide-animals-listen.md
+++ b/.changeset/wide-animals-listen.md
@@ -1,0 +1,5 @@
+---
+'sv': patch
+---
+
+fix(demo): rewrite relative import extensions to JavaScript equivalents _(`tsconfig.json` update)_
--- a/documentation/docs/30-add-ons/17-mcp.md
+++ b/documentation/docs/30-add-ons/17-mcp.md
@@ -12,7 +12,8 @@ npx sv add mcp
 
 ## What you get
 
-- A good mcp configuration for your project depending on your IDE
+- An MCP configuration for [local](https://svelte.dev/docs/mcp/local-setup) or [remote](https://svelte.dev/docs/mcp/remote-setup) setup
+- A [README for agents](https://agents.md/) to help you use the MCP server effectively
 
 ## Options
 
--- a/packages/addons/mcp/index.ts
+++ b/packages/addons/mcp/index.ts
@@ -1,5 +1,7 @@
-import { defineAddon, defineAddonOptions } from '@sveltejs/cli-core';
+import { defineAddon, defineAddonOptions, log } from '@sveltejs/cli-core';
 import { parseJson } from '@sveltejs/cli-core/parsers';
+import { getSharedFiles } from '../../create/utils.ts';
+import { getHighlighter } from '../../cli/commands/add/utils.ts';
 
 const options = defineAddonOptions()
 	.add('ide', {
@@ -60,7 +62,8 @@ export default defineAddon({
 			| {
 					schema?: string;
 					mcpServersKey?: string;
-					filePath: string;
+					agentPath: string;
+					mcpPath: string;
 					typeLocal?: 'stdio' | 'local';
 					typeRemote?: 'http' | 'remote';
 					env?: boolean;
@@ -70,36 +73,47 @@ export default defineAddon({
 			| { other: true }
 		> = {
 			'claude-code': {
-				filePath: '.mcp.json',
+				agentPath: 'CLAUDE.md',
+				mcpPath: '.mcp.json',
 				typeLocal: 'stdio',
 				typeRemote: 'http',
 				env: true
 			},
 			cursor: {
-				filePath: '.cursor/mcp.json'
+				agentPath: 'AGENTS.md',
+				mcpPath: '.cursor/mcp.json'
 			},
 			gemini: {
+				agentPath: 'GEMINI.md',
 				schema:
 					'https://raw.githubusercontent.com/google-gemini/gemini-cli/main/schemas/settings.schema.json',
-				filePath: '.gemini/settings.json'
+				mcpPath: '.gemini/settings.json'
 			},
 			opencode: {
+				agentPath: 'AGENTS.md',
 				schema: 'https://opencode.ai/config.json',
 				mcpServersKey: 'mcp',
-				filePath: 'opencode.json',
+				mcpPath: 'opencode.json',
 				typeLocal: 'local',
 				typeRemote: 'remote',
 				command: ['npx', '-y', '@sveltejs/mcp'],
 				args: null
 			},
 			vscode: {
+				agentPath: 'AGENTS.md',
 				mcpServersKey: 'servers',
-				filePath: '.vscode/mcp.json'
+				mcpPath: '.vscode/mcp.json'
 			},
 			other: {
 				other: true
 			}
 		};
+
+		const filesAdded: string[] = [];
+		const filesExistingAlready: string[] = [];
+
+		const sharedFiles = getSharedFiles().filter((file) => file.include.includes('mcp'));
+		const agentFile = sharedFiles.find((file) => file.name === 'AGENTS.md');
 
 		for (const ide of options.ide) {
 			const value = configurator[ide];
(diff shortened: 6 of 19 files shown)


packages/cli/commands/create.ts
```diff
@@ -54,11 +64,13 @@ export const create = new Command('create')
...
 	.addOption(langOption)
 	.option('--no-types')
-	.option('--no-add-ons', 'skips interactive add-on installer')
+	.addOption(addOption)
+	.addOption(noAddonsOption)
 	.option('--no-install', 'skip installing dependencies')
 	.addOption(installOption)
 	.configureHelp(common.helpConfig)
 	.action((projectPath, opts) => {
+		console.log(opts);
```
jycouet: As it's `draft` it's OK, but be careful to remove it later (Maybe add a `TODO David` so that you know ;) )

packages/cli/commands/create.ts
```diff
@@ -34,6 +41,8 @@ const langOption = new Option('--types <lang>', 'add type checking').choices(lan
 const templateOption = new Option('--template <type>', 'template to scaffold').choices(
 	templateChoices
 );
+const noAddonsOption = new Option('--no-add-ons', 'do not prompt to add add-ons').conflicts('add');
```
jycouet, line 54: TIL `conflicts` ! That's nice.

So we get: `error: option '--no-add-ons' cannot be used with option '--add <addon...>'` directly.
I'm wondering if the message could happen after `Welcome to the Svelte CLI! (v0.9.4)` (just nice to have, if it exist, but not a MUST)

packages/cli/commands/add/index.ts
```diff
@@ -314,79 +307,36 @@ export async function runAddCommand(
...
+	if (Array.isArray(options.community) && options.community.length > 0) {
+		selectedCommunityAddons.push(...(await resolveCommunityAddons(options.cwd, options.community)));
 	}
 
+	const selectedAddons: SelectedAddon[] = [
+		...selectedOfficialAddons.map((addon) => ({ type: 'official' as const, addon })),
+		...selectedCommunityAddons.map((addon) => ({ type: 'community' as const, addon }))
+	];
+
+	// TODO: run setup if we have access to workspace
```
jycouet: I think I understand better your question, and https://github.com/sveltejs/cli/issues/689#issuecomment-3263231728 would be still valid :)

What matters is only `unsupported(` I guess.

Thinking out loud...
- `setup` step is maybe a bit too wild: it's doing: `runsAfter`, `unsupported`, `dependsOn`
- in `create` we can probably assume stuff... eg: Lucia needs drizzle... So if you pick Lucia, you need to get dizzle as well... No need to check if you have drizzle deps as we are in create.
- today create only creates kit projects (so all unsupported kit are out)

Maybe we can create a "virtual workspace", not with files, just "we know this, we know that". No ?!

And/Or not only have `setup`, but also another level before: `requirements` ? That could built the virtual workspace?

I'm not sure what's the best way to do.

packages/cli/commands/add/index.ts
```diff
@@ -409,80 +359,83 @@ export async function runAddCommand(
...
-		const force = await p.confirm({
-			message: 'Verifications failed. Do you wish to continue?',
-			initialValue: false
-		});
-		if (p.isCancel(force) || !force) {
-			p.cancel('Operation cancelled.');
-			process.exit(1);
-		}
-	}
+	// TODO: add verifications and inter-addon deps
```
jycouet: Same as above, what do we do with workspace that is not here yet.

packages/cli/commands/add/index.ts
```diff
@@ -177,57 +162,63 @@ export const add = new Command('add')
...
-	type QuestionValues = OptionValues<any>;
-	type AddonOption = Record<AddonId, QuestionValues>;
+	const emptyAnswersReducer = (acc: Record<string, OptionValues<any>>, id: string) => {
+		acc[id] = {};
+		return acc;
+	};
 
-	const official: AddonOption = {};
-	const community: AddonOption = {};
+	const answersOfficial: Record<string, OptionValues<any>> = selectedOfficialAddons
```
jycouet, line 221: Didn't go too much into this, but renaming `official` to `answersOfficial` add some overhead in the diff. (maybe you can add a todo and keep `official` at first)

jycouet:
Hello @kizivat do you want to puch forward this PR? or should I try to bring it to the finish line ? :)
Let me know 👍

jycouet:
I'm not sure why `svelte.dev` is not having a preview. Any idea @manuel3108 ?

packages/cli/commands/add/index.ts
```diff
@@ -410,20 +367,20 @@ export async function runAddCommand(
 		}
 	}
 
+	// add verifications and inter-addon deps
+
 	// add inter-addon dependencies
```
manuel3108: duplicated comment

packages/cli/commands/add/index.ts
```diff
@@ -410,20 +367,20 @@ export async function runAddCommand(
...
 	for (const { addon } of selectedAddons) {
-		workspace = await createWorkspace(workspace);
-
-		const setups = selectedAddons.map(({ addon }) => addon);
-		const setupResult = setupAddons(setups, workspace)[addon.id];
+		workspace = virtualWorkspace || (await createWorkspace({ ...workspace }));
```
manuel3108: not sure this is the right call here. Previously we `createWorkspace` after every run, so that if tailwind and vitest run (after each other), the vitest run should already know via the workspace that `tailwind` is already installed.

Or am I missing something here?

packages/cli/commands/add/index.ts
```diff
@@ -530,13 +484,38 @@ export async function runAddCommand(
...
+	virtualWorkspace
+}: {
+	answersOfficial: Record<string, OptionValues<any>>;
+	answersCommunity: Record<string, OptionValues<any>>;
+	options: Options;
+	selectedAddons: SelectedAddon[];
+	addonSetupResults?: Record<string, AddonSetupResult>;
+	virtualWorkspace?: Workspace<any>;
+}): Promise<{ nextSteps: string[]; packageManager?: AgentName | null }> {
+	let workspace = virtualWorkspace || (await createWorkspace({ cwd: options.cwd }));
```
manuel3108: Same as above? in which case would virtualWorkspace even be undefined? Couldnt we just do this check once and make sure it's not undefined later on.

packages/cli/commands/add/index.ts
manuel3108: Actually not sure why that works, but that seems to work

packages/cli/commands/add/index.ts
jycouet: You are 100% right!
I will refactor this later 👍

jycouet: I refactored it spliting `workspace` & `options`
- workspace is pretty much static
- and you want to build up options questions after questions

Like this, we don't need to build all the time workspace. And only options are enhanced step by step.

It start to look good IMO.
I resolve this comment, if needed we can start new one in up to date code.

jycouet:
Will fix #688 as well
image

.changeset/smooth-foxes-unite.md
```diff
@@ -0,0 +1,5 @@
+---
+'sv': patch
```
manuel3108: Note to myself: one of the changsets (or both) should probably be minor, as we are changing quite a lot under the hood and this is imo a huge milestone. And I'm pretty sure we can also find a technical reason for that 😆

packages/cli/utils/common.ts
```diff
@@ -135,3 +136,20 @@ export function parseAddonOptions(optionFlags: string | undefined): string[] | u
 
 	return options;
 }
+
+export function logArgs(agent: AgentName, actionName: string, args: string[]) {
+	const agentCmd: Record<AgentName, string> = {
+		npm: 'npx sv',
```
manuel3108: Reminder for myself: didn't we had a tool for that?

manuel3108 (commented):
Consider this a WIP review, have to leave now for a bit.

Other notes:

* The printed re-usable command does print `pnpm dlx sv` if i used it locally with `pnpm sv`. This is ok and expected.
* The printed re-usable command does not include the given `--path` flag. You could consider this being a bug, but since you are probably not going to work on the same folder twice, I would actually consider this a feature. Works as expected. WAIT: It actually includes the path but without the `--path` flag, which does technically also work. Based on the reasoning above, i don't think we should ever include the path here. Actually it's even worse, it prints the wrong path. I used `--path ../cli-temp` and it just printed `pnpm dlx sv cli-temp`
* Also for the sake of cross platform users we should qoute the `add-on args.` (Windows users here). Otherwise i had trouble getting the `tailwindcss` flags to work. We also do this in the docs i think.
* Also the printed re-usable command for `pnpm dlx sv add` ommits at least the used `--install pnpm` flag. Interestingly this works for `create`

packages/cli/utils/common.ts
manuel3108: Yeah, I think we should use `resolveCommand` here.

manuel3108:
Apart from what i wrote above, i did not find anything else.

jycouet:
@manuel3108 about your `--path` comment.
I guess you refer to `npx sv create [options] [path]` <-- this `path`

I didn't realise that it's working like this! :o
For me it's more "project name" and not really "path" _(yes the project name is part of the path)_

I actually never tried `../cli-temp` and I would probably do `cli-temp --cwd ..` But this doesn't even work probably! because `--cwd` is not in the create command. 

It's ok, I learn. Anyway, I think that we should display it. I'll check it later

manuel3108:
> I guess you refer to `npx sv create [options] [path]` <-- this `path`

Yeah you are probably right. I just double-checked my command history, and i don't find anything for `--path`. But yes that would make sense. I always use that command to test how they locally build cli works, as in the past i had problems doing it in the same directory. So I always do it outside of our monorepo.

Actually, I just checked our docs, and we should also unify the `add` and `create` docs. For `add` we currently do not properly visualize all the parameters as we do in `create`. See #786

packages/cli/commands/add/index.ts
```diff
@@ -673,6 +676,8 @@ export async function runAddonsApply({
 			await formatFiles({ packageManager, cwd: options.cwd, paths: filesToFormat });
 			stop('Successfully formatted modified files');
 		} catch (e) {
+			console.log(`e`, e);
```
manuel3108: is this intentional?

Issue #687, opened 2025-09-04
install deps step, only if needed


jycouet:
Would be nice to have `packageManagerPrompt` & `installDependencies` only if needed.

- If an add-on has no deps, `sv` should skip this step
- If dependencies are already there _(with at least version x)_, `sv` should skip this step

manuel3108:
Not sure if this is easy enough / worth the effort to detect. Especially that as of now 12 out of 12 addons add dependencies according to my knowledge.

jycouet:
Imagine you have `vitest` and you want to add `component testing`, vitest is already there.
I could also imagine a few community add-on just for scaffolding stuff (wo deps)

I think that it's pretty easy to handle with `isVersionUnsupportedBelow` that I did earlier.
I can try something and we see

manuel3108:
See https://github.com/sveltejs/cli/pull/696#issuecomment-3329616995

PR #696, opened 2025-09-08
feat(add): install prompt only if needed

jycouet:
Closes: #687

---

```sh
# create an empty project
pnpm build && rm -rf jyc && node ./packages/cli/dist/bin.js create jyc --template minimal --types ts --no-install --no-add-ons

# add devtools-json (install prompt will come)
node ./packages/cli/dist/bin.js add -C jyc --no-git-check devtools-json

# add devtools-json again, (install prompt will not show)
node ./packages/cli/dist/bin.js add -C jyc --no-git-check devtools-json
```


packages/cli/commands/add/utils.ts
```diff
@@ -57,27 +57,45 @@ export function readFile(cwd: string, filePath: string): string {
 	return text;
 }
 
+const checkInstallNeeded = (atLeastVersion: string, version: string | undefined) => {
+	if (!version) return true;
+	const atLeastVersionStripped = atLeastVersion.replace('^', '');
+	const versionStripped = version.replace('^', '');
+	return !isVersionAtLeast(versionStripped, atLeastVersionStripped);
+};
```
AdrianGonz97, line 65: i think we may be doing too much with the inclusion of version checking. dependency versions can get very complicated with tons of edge-cases, so i'm not sure if this is really worth the effort. (note: we can get away with it with `@types/node` since we're targeting a very specific dependency version that we'll always know the shape/type of. we don't really have that guarantee with other 3rd party deps)

perhaps we should just show the install prompt only if there are any dependencies to be installed? although, im not sure how useful that'd be since almost every add-on tends to add some sort of new dependency

jycouet: > very complicated with tons of edge-cases

Don't tell me 😅, spent hours there before.

---

The idea is that `checkInstallNeeded` _(and `isVersionAtLeast` under)_, do it's best for simple case, and if not will fallback with prompting. So I would say that it's helping "normal" cases and fallback to todays way.

I would totally see "Community Add-On" for scaffolding code (with a dep).

---

But you are already the second one saying that it might not worth the effort. _(It's good to slow down my ideas from time to time :p)_
I still think it's a good addition as the fallback is like today. But yes, it's a bit of code, maintenance, surface, ...

manuel3108:
Looks like for now we are 2 on one regarding too much effort. So let's leave it as is for now

PR #698, opened 2025-09-09
chore(cli): speedup internal tests

jycouet:
I notice that we had a few timeout and wanted to improve a bit the speed of this.

Here are my notes:
- [x] new option to `skipBrowser` when there is no tests requiring `page` (a small ~10% speedup)
- [x] I did a lot to create all variants and just one `pnpm i`, but the speedup was not way better and the code was ulgy... I removed.
- [x] Tune some `installCommand` & `buildCommand` to run only when we have `ts`. I was thinking to gain quite some time... but it doesn't show.

Any other idea of where to look ?

### DIFF
--- a/packages/addons/_tests/storybook/test.ts
+++ b/packages/addons/_tests/storybook/test.ts
@@ -19,7 +19,7 @@ beforeAll(() => {
 	}
 });
 
-test.skip.for(variants)(
+test.for(variants)(
 	'storybook loaded - %s',
 	{ concurrent: !CI },
 	async (variant, { page, ...ctx }) => {


AdrianGonz97:
thanks for taking a look! i've spent a ton of time trying to get our tests to run faster in the past and it's certainly a difficult task 😅. to be frank though, the main hold up is really `storybook`. for some reason, running its cli and then spinning up the preview takes ages (when compared to everything else) and is truly the choke point for all of our tests. in my testing, if we were to exclude it, then our entire test suite would run in about 1/4 of the time.

> * [x]  new option to `skipBrowser` when there is no tests requiring `page` (a small ~10% speedup)

good idea! i tweaked the name to just `browser` and inverted its condition. i think it reads a bit better

> * [x]  Tune some `installCommand` & `buildCommand` to run only when we have `ts`. I was thinking to gain quite some time... but it doesn't show.

this distinction for `ts` files isn't really necessary since the project needs to be built anyway for the tests that run a preview server. although in cases where a preview server isn't needed, like `vitest`, `eslint`, and `playwright`, i just removed the use of `prepareServer` altogether

in addition to the above, i also pushed a few improvements, notably executing commands asynchronously (`execAsync` instead of `execSync`) which did cut the times in half for a few of the tests (`vitest` and `eslint`)

(edit: i think i broke things with the async stuff, sorry 😅 - seemingly only breaks in macos for some reason... i'll take another crack at it tomorrow!)

jycouet:
Héhé, nice tweaks! _I let you check the failing async stuff, if you need me to look at something let me know_

For `storybook`, do we need to do "everything" for all variants ? 
All in all we are checking if there are 2 elements in preview. We could imagine `vite-js` and `kit-ts` only for example ? We would have a pretty much "similar" coverage. No ?

manuel3108:
Due to the tests here for `storybook` in the past we were able to catch quite a few issues that we shared with the storybook team, which in turn got fixed. So it had a valid use-case. You could argue though if this should be our job at all. 

I don't have a better idea though, so just giving my two cents

packages/addons/_tests/storybook/test.ts
```diff
@@ -10,17 +10,18 @@ import eslint from '../../eslint/index.ts';
...
 		// prefetch the storybook cli during ci to reduce fetching errors in tests
 		execSync('pnpx create-storybook@latest --version');
 	}
 });
 
-test.for(variants)(
+test.skip.for(variants)(
 	'storybook loaded - %s',
-	{ concurrent: !process.env.CI },
+	{ concurrent: !CI },
```
jycouet: Shouldn't we try to have the same in `CI` and not in `CI` ?

AdrianGonz97: ideally, yes! however the `storybook` tests, [for some reason](https://github.com/sveltejs/cli/issues/293), are super flaky _only in CI_ when the tests were ran concurrently. unfortunately this was the only solution we could find that works in CI without slowing down tests locally

jycouet:
Thx for looking, working and winning this battle.
Let's see when we will do this activity again ^^

PR #699, opened 2025-09-09
refactor(release): run with node24 and remove NPM_TOKEN

dominikg:
to use oidc publish
 
 requires npm package settings update, draft until that is done


Rich-Harris (approved):
`svelte-migrate` and `sv` are configured, looks like those are the only public packages published from this repo

Issue #700, opened 2025-09-10
When `vitest`, vite config `defineConfig` should come from `vitest/config`


jycouet:
When you select `vitest`, 

```ts
// this import
import { defineConfig } from "vite";

// should switch to 
import { defineConfig } from "vitest/config";
```

If you don't do this & you have 0 import of `vitest` in your project, your `vite.config` will show an error.

cc @paoloricciuti

PR #703, opened 2025-09-11
fix(vitest): now import `defineConfig` from `vitest/config`

jycouet:
Closes: #700


manuel3108:
Can confirm this works as expected, trusting you on the code.
Im still struggling to understand which vite / ts magic is making this error occur in the first place, but this seems like a nice fix.

PR #710, opened 2025-09-13
chore(cli): ts detection & rmv NODE_ENV tweaks

jycouet:


packages/cli/utils/env.ts
AdrianGonz97, line 1: what's the reason behind removing the `TESTING` constant? one of the places we were using it was to prevent flooding the output during tests (e.g. `storybook` fills the whole terminal with text)

jycouet: It was used in 2 places, the ts detection (now changed) and in `storybook `.
In the end, it was to reduce the number of `process.env.` Maybe we could tight the `storybook` part to `process.env.CI` ?
It's not exactly the same thing, I can also bring it back (or inline it like in other places? or centralize `env` in general?)

jycouet: Now focusing on `ts detection` only.

Issue #705, opened 2025-09-12
pnpm `onlyBuiltDependencies` is added regardless of `pnpm-workspace.yaml`


hyunbinseo:
`package.json / pnpm.onlyBuiltDependencies` is added

```json
{
  "packageManager": "pnpm@10.15.1+sha512.34e538c329b5553014ca8e8f4535997f96180a1d0f614339357449935350d924e22f8614682191264ec33d1462ac21561aff97f6bb18065351c162c7e8f6de67",
  "pnpm": {
    "onlyBuiltDependencies": [
      "esbuild"
    ]
  }
}
```

even when `pnpm-workspace.yaml / onlyBuiltDependencies` exists.

```yaml
# pnpm-workspace.yaml

packages:
  - packages/*
  - sites/*

onlyBuiltDependencies:
  - '@tailwindcss/oxide'
  - esbuild
```

manuel3108:
Relates #521

jycouet:
2 issues can be done in 1 PR 🧡
I might do it over the weekend. _(I'll have to check if we have some .yaml helpers)_

jycouet:
Released in `sv@0.9.7`

Issue #521, opened 2025-04-02
Update pnpm `onlyBuiltDependencies` config


hyunbinseo:
`onlyBuiltDependencies` in `pnpm-workspace.yaml` is supported from [`pnpm@10.5`](https://github.com/pnpm/pnpm/releases/tag/v10.5.0):

> The `pnpm.*` settings from `package.json` can now be specified in the `pnpm-workspace.yaml` file instead.

```yaml
onlyBuiltDependencies:
- esbuild
- fuse-native
```

`pnpm approve-builds` adds `onlyBuiltDependencies` to `pnpm-workspace.yaml` as of `10.7.1`.

Current implementation is fixed to the `pnpm.onlyBuiltDependencies` field in the `pacakge.json`:

- https://github.com/sveltejs/cli/pull/432

Backwards compatibility is supported and is probably fixed (?) in [`10.6.1`](https://github.com/pnpm/pnpm/releases/tag/v10.6.1)

> When executing the `approve-builds` command, if package.json contains `onlyBuiltDependencies` or `ignoredBuiltDependencies`, the selected dependency package will continue to be written into `package.json`.

manuel3108:
I wonder what should happen if there is no `pnpm-workspace.yaml` because you are creating a new project. Should it just be created to only add those three lines?

Basically I think this is a good idea, because it puts pnpm related stuff in an pnpm file

jycouet:
Released in `sv@0.9.7`

PR #713, opened 2025-09-15
feat(cli): pnpm config will now be stored in `pnpm-workspace.yaml`

jycouet:
Closes: #521
Closes: #705

### DIFF
--- a/.changeset/weak-chefs-notice.md
+++ b/.changeset/weak-chefs-notice.md
@@ -1,0 +1,5 @@
+---
+'sv': patch
+---
+
+feat(cli): pnpm config will now be stored in `pnpm-workspace.yaml` (e.g. `onlyBuiltDependencies`)
--- a/package.json
+++ b/package.json
@@ -40,10 +40,5 @@
 		"unplugin-isolated-decl": "^0.8.3",
 		"vitest": "4.0.0-beta.6"
 	},
-	"packageManager": "pnpm@10.4.1",
-	"pnpm": {
-		"onlyBuiltDependencies": [
-			"esbuild"
-		]
-	}
+	"packageManager": "pnpm@10.4.1"
 }
--- a/packages/cli/utils/package-manager.ts
+++ b/packages/cli/utils/package-manager.ts
@@ -12,7 +12,7 @@ import {
 	detect,
 	type AgentName
 } from 'package-manager-detector';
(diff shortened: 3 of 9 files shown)


jycouet:
Hey @hyunbinseo would you mind testing this `sv` ? :)

packages/addons/_tests/vitest/test.ts
```diff
@@ -4,7 +4,7 @@ import vitest from '../../vitest-addon/index.ts';
 
 const { test, variants } = setupTest({ vitest }, { browser: false });
 
-test.concurrent.for(variants)('core - %s', async (variant, { expect, ...ctx }) => {
+test.sequential.for(variants)('core - %s', async (variant, { expect, ...ctx }) => {
```
manuel3108, line 7: Why add the sequential stuff here?

jycouet: I'm trying to think why "It's working on my machine" and not on CI :(.

---

On this concurrent/sequential testing... this is my thinking at the moment:
Let's say we have 4 variants. Usually, they have the same deps. And today the structure of an add-on test is like this: 
```bash
.test-output
  addons
    vitest
      package.json # vitest-workspace-root
      pnpm-workspace.yaml
      variant_0
        package.json
        # ...
      variant_1
        package.json
        # ...
      variant_2
        package.json
        # ...
      variant_3
        package.json
        # ...
```
And we say to the 4 variants: `concurrently` and they all start with `pnpm i` in parallel.
So they all do the installation process (~4 secs) at the same time and everything ends up in `.test-output/addons/vitest/node_modules` (I'm even thinking how come it's working?!)

If you do it `sequential`, then, `variant_0` will do the install (~4 sec) then build, expect, ... then it's time to `variant_1` will do an install, but here, since the workspace was already installed it should be instant then build, expect, ... etc etc...

---

Locally, with manual testing, `onlyBuiltDependencies` in `pnpm-workspace.yaml` works pretty well. So I'm trying to see what could be the difference with the CI.

On Mac: `Error: Command failed: pnpm install`
On Windows: `Serialized Error: { errno: -4082, code: 'EBUSY', syscall: 'rmdir', path: 'D:\a\cli\cli\.test-output\addons\tailwindcss\-767017923_7' }`
What could it be ?! :(

I should probably try this in my fork to not notify everyone all the time 👍

PR #717, opened 2025-09-18
chore(cli): speedup internal tests v2

jycouet:
Proposal to change how tests are organized _(to speedup tests again ^^)_

- `variant` + `kind` = `flavor`
- `setupTest` prepare the test mono repo with all flavors and only one install is needed
- remote `ps-tree` dep

### DIFF
--- a/packages/addons/_tests/_setup/suite.ts
+++ b/packages/addons/_tests/_setup/suite.ts
@@ -124,7 +124,7 @@ export function setupTest<Addons extends AddonMap>(
 		};
 	});
 
-	return { test, variants, prepareServer, flavors };
+	return { test, flavors, prepareServer };
 }
 
 type PrepareServerOptions = {
--- a/packages/addons/_tests/_setup/suite.ts
+++ b/packages/addons/_tests/_setup/suite.ts
@@ -21,12 +21,21 @@ export const execAsync = promisify(exec);
 
 type Fixtures<Addons extends AddonMap> = {
 	page: Page;
-	run(variant: ProjectVariant, options: OptionMap<Addons>): Promise<string>;
+	run(flavor: Flavor<Addons>): string;
+};
+
+type Flavor<Addons extends AddonMap> = {
+	variant: ProjectVariant;
+	kind: { type: string; options: OptionMap<Addons> };
 };
 
 export function setupTest<Addons extends AddonMap>(
 	addons: Addons,
-	options?: { browser?: boolean }
+	options?: {
+		kinds: Array<Flavor<Addons>['kind']>;
+		filter?: (flavor: Flavor<Addons>) => boolean;
+		browser?: boolean;
+	}
 ) {
 	const test = vitest.test.extend<Fixtures<Addons>>({} as any);
 
@@ -44,8 +53,18 @@ export function setupTest<Addons extends AddonMap>(
 		});
 	}
 
-	vitest.beforeAll(({ name }) => {
-		const testName = path.dirname(name).split('/').at(-1)!;
+	const flavors: Array<Flavor<Addons>> = [];
+	for (const kind of options?.kinds ?? []) {
+		for (const variant of variants) {
+			const flavor = { variant, kind };
+			if (!options?.filter || options?.filter?.(flavor)) {
+				flavors.push(flavor);
+			}
+		}
+	}
+	let testName: string;
+	vitest.beforeAll(async ({ name }) => {
+		testName = path.dirname(name).split('/').at(-1)!;
 
 		// constructs a builder for create test projects
 		create = createProject({ cwd, templatesDir, testName });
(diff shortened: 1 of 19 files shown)


jycouet:
Some info _(I thought I wrote it somewhere, but can't find it)_

### Before this PR tests structure looks like:
```
test/add-on-y
  package.json
  pnpm-workspace.yaml
  node_modules
    ...
  12345_0
    package.json
    ...
  12345_1
    package.json
    ...
  12345_2
    package.json
    ...
  12345_3
    package.json
    ...
```
And the process is: 
1/ create`pnpm-workspace.yaml` 
2/ **concurrently** 
  - `pnpm i` in each project folder `12345_0`, `12345_1`, ... adding stuff in the shared `node_modules`.
  - build & tests

### After this PR tests structure looks like:
```
test/add-on-y
  package.json
  pnpm-workspace.yaml
  node_modules
    ...
  default-kit-js
    package.json
    ...
  default-kit-ts
    package.json
    ...
  default-vite-js
    package.json
    ...
  default-vite-ts
    package.json
    ...
```
1/ create`pnpm-workspace.yaml` 
2/ `pnpm i` at `add-on-y` level... once.
3/ **concurrently** 
  - build & tests

packages/cli/lib/testing.ts
```diff
@@ -84,7 +84,7 @@ export function createProject({ cwd, testName, templatesDir }: CreateOptions): C
 type PreviewOptions = { cwd: string; command?: string };
 export async function startPreview({
 	cwd,
-	command = 'npm run preview'
+	command = 'pnpm preview'
```
AdrianGonz97: this should really remain as `npm run preview` as it's meant to be a public api for community-addon authors to use to test their add-ons. `npm` is a good default since we can be pretty certain that they have it installed since it comes with node (unless they're using deno or bun, i guess)
```suggestion
	command = 'npm run preview'
```

packages/cli/utils/package-manager.ts
```diff
@@ -2,7 +2,7 @@ import fs from 'node:fs';
 import path from 'node:path';
 import process from 'node:process';
 import * as find from 'empathic/find';
-import { exec } from 'tinyexec';
+import { x } from 'tinyexec';
```
AdrianGonz97: any reason to not use `exec`? we use it elsewhere and it's a bit more explicit rather than just being a single letter var
```suggestion
import { exec } from 'tinyexec';
```

packages/cli/utils/package-manager.ts
```diff
@@ -57,19 +58,16 @@ export async function installDependencies(agent: AgentName, cwd: string): Promis
...
-		});
-		proc.process?.stderr?.on('data', (data) => {
-			task.message(data.toString(), { raw: true });
-		});
-
-		await proc;
+		for await (const line of proc) {
+			// line will be from stderr/stdout in the order you'd see it in a term
+			task.message(line, { raw: true });
+		}
```
AdrianGonz97, line 70: ahh, so much nicer!

packages/addons/_tests/_setup/suite.ts
```diff
@@ -21,12 +21,21 @@ export const execAsync = promisify(exec);
 
 type Fixtures<Addons extends AddonMap> = {
 	page: Page;
-	run(variant: ProjectVariant, options: OptionMap<Addons>): Promise<string>;
+	run(flavor: Flavor<Addons>): string;
+};
+
+type Flavor<Addons extends AddonMap> = {
+	variant: ProjectVariant;
+	kind: { type: string; options: OptionMap<Addons> };
```
AdrianGonz97: `Flavor` seems a bit too abstract 😅

how about something like `TestCase` or `TestItem` or `AddonTest`/`AddonTestCase`?

jycouet: I have to say that I kinda llike `flavor` a lot, at least for the instance. For the type, now that you write it... It looks a bit funny.

Let's see with `AddonTestCase`

community-addon-template/tests/setup/suite.ts
```diff
@@ -21,24 +21,50 @@ const SETUP_DIR = fileURLToPath(new URL('.', import.meta.url));
 
 type Fixtures<Addons extends AddonMap> = {
 	page: Page;
-	run(variant: ProjectVariant, options: OptionMap<Addons>): Promise<string>;
+	run(addonTestCase: Flavor<Addons>): string;
```
manuel3108: I think you actually forgot to rename the types in your last commit, shouldn't this be `AddonTestCase<>` now?

packages/addons/_tests/_setup/suite.ts
```diff
@@ -44,8 +53,18 @@ export function setupTest<Addons extends AddonMap>(
...
+			if (!options?.filter || options?.filter?.(addonTestCase)) {
+				addonTestCases.push(addonTestCase);
+			}
+		}
+	}
+	let testName: string;
+	vitest.beforeAll(async ({ name }) => {
+		testName = path.dirname(name).split('/').at(-1)!;
 
 		// constructs a builder for create test projects
```
manuel3108:
```suggestion
		// constructs a builder to create test projects
```
unrelated, but while we are at it

packages/core/tooling/index.ts
```diff
@@ -238,3 +239,11 @@ export function guessQuoteStyle(ast: TsEstree.Node): 'single' | 'double' | undef
 
 	return singleCount > doubleCount ? 'single' : 'double';
 }
+
+export function parseYaml(content: string): ReturnType<typeof yaml.parseDocument> {
+	return yaml.parseDocument(content);
+}
+
+export function serializeYaml(data: unknown): string {
```
manuel3108: Not on the right computer to confirm this, but couldn't the param type here be the same as the return type of parse? `ReturnType<typeof yaml.parseDocument>`

manuel3108 (commented):
This looks great! Will leave it to Adrian to leave the final approval, he has been way more involved in the tests recently.

community-addon-template/tests/setup/suite.ts
jycouet: 🫣 oupy!

_I don't really know why, but with this branch I have a mix with remotes_

packages/core/tooling/index.ts
jycouet: Yes, correct ✌️

jycouet:
Thank you for looking and adding these tweaks & last bits.
Not 100% sure about the naming `ctx.cwd`, maybe `ctx.getCwd`? But it's rather internal for now, so let's keep it like this and see how we feel about it later.

PR #719, opened 2025-09-18
fix(cli): export types & switch to tsdown

jycouet:


tsdown.config.ts
```diff
@@ -0,0 +1,53 @@
...
+import fs from 'node:fs';
+import path from 'node:path';
+import process from 'node:process';
+import { defineConfig } from 'tsdown';
+
+export default defineConfig({
+	cwd: 'packages/cli',
+	entry: ['lib/index.ts', 'lib/testing.ts', 'bin.ts'],
+	sourcemap: !process.env.CI,
+	dts: { oxc: true },
```
AdrianGonz97: the types from `package-manager-detector` wasn't being bundled when looking at the generated `dist/lib/testing.d.ts` file, this should fix it:
```suggestion
const pkgJson = fs.readFileSync(path.resolve('packages', 'cli', 'package.json'), 'utf8');
const pkg = JSON.parse(pkgJson);

export default defineConfig({
	cwd: 'packages/cli',
	entry: ['lib/index.ts', 'lib/testing.ts', 'bin.ts'],
	sourcemap: !process.env.CI,
	dts: {
		oxc: true,
		// setting `resolve: true` seems to anger Rolldown due to our `postcss` dep
		resolve: Object.keys(pkg.devDependencies)
	},
```

tsdown.config.ts
```diff
@@ -0,0 +1,53 @@
...
+	],
+	inputOptions: {
+		experimental: {
+			resolveNewUrlToAsset: false
+		}
+	},
+	hooks: {
+		async 'build:done'() {
+			await buildCliTemplates();
+		}
```
AdrianGonz97: this was causing an infinite build loop in watch mode. changing to the `build:before` hook fixes it:
```suggestion
		async 'build:before'() {
			await buildCliTemplates();
		}
```

packages/core/package.json
```diff
@@ -16,30 +17,12 @@
...
-		},
-		"./parsers": {
-			"types": "./dist/parsers.d.ts",
-			"default": "./dist/parsers.js"
-		}
+		".": "./index.ts",
+		"./css": "./tooling/css/index.ts",
+		"./html": "./tooling/html/index.ts",
+		"./js": "./tooling/js/index.ts",
+		"./parsers": "./tooling/parsers.ts"
```
AdrianGonz97, line 25: we'll have to keep in mind that `@sveltejs/cli-core` will be a public package in the future, so this will have to be updated back to its original state when community add-ons are released

AdrianGonz97 (commented):
this looks great! for some reason, i'm unable to push some changes to this pr, so i've added some suggestions below.

mrginglymus:
Thanks, types are coming through now!

There are a few types that could be usefully reexported (TemplateType and LanguageType) to avoid `type TemplateType = Parameters<typeof create>[1]["template"];`

What's significantly missing though is typed official addons - because they're exported as a list, there's no way have type-safe configured addons:

https://github.com/sveltejs/cli/blob/262ac80d070bf8b9f132d895e2d36aeaedda57bb/packages/addons/_config/official.ts#L18

Could we change this to 

```ts
export const officialAddons = {
  prettier,
  eslint,
  ...,
} satisfies Record<string, AddonWithoutExplicitArgs> as const;
```
 or similar? That way a user will be able to define, in a type safe way, a set of preconfigured options for the addons.

jycouet:
_I should not have created the PR! that's why we can't push there. :/ (TIL!)_
Now @sxzz invited me to his repo, I'll try to apply suggestions _(today)_ and see 👍 

EDIT: it seems that you can push now @AdrianGonz97

packages/core/package.json
jycouet: Will it be `@sveltejs/cli-core` or maybe under `sv` like `sv/core` ?
What about `@sveltejs/create` & `@sveltejs/addons` ?

It might be a question for later anyway

jycouet:
> Could we change this to
> 
> ```ts
> export const officialAddons = {
>   prettier,
>   eslint,
>   ...,
> } satisfies Record<string, AddonWithoutExplicitArgs> as const;
> ```
> 
> or similar? That way a user will be able to define, in a type safe way, a set of preconfigured options for the addons.

Before exporting this, we need to fix `pnpm tsc`. (Similar I think)
I'm not sure where it's coming from as the typing was good before _(internally)_.  _The thing is that they are not all `AddonWithoutExplicitArgs`._

Who wants to crack the matrix ? :D

mrginglymus:
Well I thought I'd fixed `pnpm check`, but....rolldown doesn't support custom export conditions?

sxzz:
Of course, Rolldown supports it.

mrginglymus:
Something in my current toolchain (via `rollup-plugin-dts`) is dropping it...

manuel3108:
Converted to draft, dosn't seem ready yet according to the comments

packages/core/package.json
AdrianGonz97: > Will it be `@sveltejs/cli-core` or maybe under `sv` like `sv/core` ?

its still TBD, but our plans were to have `core` be its own standalone package so that community add-ons don't have `sv` as a direct dependency

> What about `@sveltejs/create` & `@sveltejs/addons` ?

`create` should probably stay in `sv`, but there's an argument to be made as to where `addons` should go. it may be more useful to have it be part of `core` but we can figure it out when we get there :p

AdrianGonz97:
exposing the `officialAddons` map is turning into a bit of a head scratcher with relation to `tsdown`. 

atm, i'm trying to disable `oxc` (as well as disabling `isolatedDeclarations`) and use `tsc` instead to generate the types through `tsdown`, but it seems that `rolldown-plugin-dts` is getting hung up on some of the types that originate from `@sveltejs/create`, and are reexported from the cli:
```
[plugin rolldown-plugin-dts:fake-js]
SyntaxError: Export 'LanguageType' is not defined. (57:35)
```

i have a feeling this may be a `rolldown-plugin-dts` bug, but im unsure. have any ideas @sxzz?

AdrianGonz97:
i guess we could keep using `isolatedDeclarations` with `oxc`, but it would force us to generate types that aren't really ideal:

img

with `Addon<any>`, we lose the types of the addon's options, but keep the typed keys for each entry (e.g. `officialAddons.prettier`, etc). we'd also have to maintain the `OfficialAddons` type, but that's a rather small annoyance.

ideally, using `as const` (shown on the left) would be the best, but it doesn't seem to be compatible with `isolatedDeclarations`

PR #720, opened 2025-09-19
fix(cli): `--from-playground` option & node 20

jycouet:


manuel3108:
Actually this doesnt look like a flaky test, but rather something else. However, i cannot imagine that these changes caused this. Rerunning now. If it passes, feel free to merge.

Issue #721, opened 2025-09-23
Run `pnpm update` after initialization


hyunbinseo:
Running `pnpm update` updates the SemVer inside `package.json` to match the initial `pnpm-lock.yaml` generated.

```diff
- "@sveltejs/adapter-auto": "^6.0.0",
- "@sveltejs/kit": "^2.22.0",
- "@sveltejs/vite-plugin-svelte": "^6.0.0",
- "svelte": "^5.0.0",
- "svelte-check": "^4.0.0",
- "typescript": "^5.0.0",
- "vite": "^7.0.4"
```

```diff
+ "@sveltejs/adapter-auto": "^6.1.0",
+ "@sveltejs/kit": "^2.43.1",
+ "@sveltejs/vite-plugin-svelte": "^6.2.0",
+ "svelte": "^5.39.4",
+ "svelte-check": "^4.3.2",
+ "typescript": "^5.9.2",
+ "vite": "^7.1.7"
```

manuel3108:
Are you suggesting to always run `pnpm update` after the creation of project? Im not convinced this is going to be a good idea, since that might leave you with a broken project, right after the creation. Currently we run CI on the packages we install, to make sure that everything is at least partly compatible.

If you suggest to bumping the deps from time to time then #722 is already under way.

hyunbinseo:
Svelte `^5.0.0` does install `5.39.4` in initialization, it is just not applied in the `package.json` file.

I always run `pnpm update` after initialization, so that the future `pnpm update` can precisely show what packages have been updated by simply diffing the `package.json` file.

jycouet:
I think that bumping versions from time to time from sv is the best to be "in control".
And it applies to all package managers.

Before this PR, sv was bumping only addons regularly, now also based projects, so I guess it will be closer and will avoid a gap of minor of 39 versions.

manuel3108:
Agree, let's move forward with that PR

PR #722, opened 2025-09-23
chore(cli): bump defaults versions

jycouet:
Closes: #721

---

Making the update script a bit more open to check `templates` & `shared` stuff.
Let me know if it's "too much" :)


manuel3108:
Same test issues as in #720 for whatever reason. 
In the past we had some different opinions on this script overall. I don't think that those dicussions have been public though.

One of the main thoughts was that package managers are already fetching and installing the newest version for us if it's a first install (at least for patches, not sure about minors, don't think so for majors). So why should we bother keeping these up to date? But tbh, having a look at the div, this would barely make a difference in this case.

jycouet:
With the update tests are now good ✌️

PR #730, opened 2025-10-03
fix push to `kit-template-default`

jycouet:
Looking at https://github.com/sveltejs/cli/actions/runs/18222249563/job/51884670580 there is an issue with the new build. (it's internal thing)

I addapted the script. Should work now.

But looking https://github.com/sveltejs/kit-template-default last deployed was `0.8.21` ! _(the one after is `0.9`)_ Maybe an issue with the `bot` permissions?

### DIFF
--- a/packages/create/scripts/update-template-repo-contents.js
+++ b/packages/create/scripts/update-template-repo-contents.js
@@ -1,6 +1,6 @@
 import fs from 'node:fs';
 import path from 'node:path';
-import { create } from '../dist/index.js';
+import { create } from '../../cli/dist/lib/index.js';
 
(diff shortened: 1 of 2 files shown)


manuel3108:
Yeah, it's probably a permission thing. Someone recently suggested somewhere that instead of provide a repo template we should just run sv from the command line, after stackblitz has initialized, similar to what is done with `npm install` at the moment. 
But no clue where I saw that and what the outcome was.

jycouet:
I think that we should first merge this so that it's at least building wo error when publishing a package. _(it's due to the new tsdown)_

Then we can see if we remove completely the template or the step or... 

Let me know

PR #731, opened 2025-10-03
`--from-playground` & PlaygroundLayout

jycouet:
Here is the result: 
image

## Summary
- [x] a new file `/src/lib/PlaygroundLayout.svelte` to bring the Playground look and feel.
  - [x] title: name of the playground
  - [x] link: to the online playground
  - [x] theme switch available
- [x] change files location to `src/lib/playground`

### DIFF
--- a/.changeset/tasty-friends-think.md
+++ b/.changeset/tasty-friends-think.md
@@ -1,0 +1,5 @@
+---
+'sv': patch
+---
+
+feat(cli): `--from-playground` will now bring a PlaygroundLayout to get a more consistent experience with the online playground
--- a/packages/cli/commands/create.ts
+++ b/packages/cli/commands/create.ts
@@ -244,7 +244,7 @@ async function createProjectFromPlayground(url: string, cwd: string): Promise<vo
 	const dependencies = detectPlaygroundDependencies(playground.files);
 	const installDependencies = await confirmExternalDependencies(Array.from(dependencies.keys()));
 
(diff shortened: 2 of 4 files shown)


manuel3108:
Even if I have said different things in the past, i must say that I kinda like this. The only thing i would like to prevent code wise is to have the massive string in that source code file. Maybe extract it into it's own file, or read that template in from a text file or something similar. That's the only thing i would suggest.

jycouet:
It's good to have stuff step by step _(also before, the idea wass less "designed" and too early)_.

I agree with you that this massive string is not great. I will look at moving it to a file ✌️ 
_(For some reason, I know it can be a challenge to have this file not ignoed at publush time, old nightmare... Will check :D)_

jycouet:
Hooo, I'm confused! 
It was not the case when it was not extracted into another file... I'll investigate.

manuel3108:
Probably because we did not need to parse the file? I think either the parsing or the printing is doing strang stuff, since we are currently not using the svelte compiler

jycouet:
> Probably because we did not need to parse the file? I think either the parsing or the printing is doing strang stuff, since we are currently not using the svelte compiler

Yeah, so I updated the code to still use it _(to have `lang="ts"` or not)_ + reduce the template as much as possible + doing a replace. Not that nice, but working workaround.
I added a few tests to make sure the output is ok.

manuel3108 (approved):
Works great now!

Issue #734, opened 2025-10-08
Add .mcp.json when claude/codex is detected


mquandalle:
It would be handy if cli agents like `claude` or `codex` could be detected and an option to add [svelte mcp](https://svelte.dev/docs/mcp/overview) support at the project level be proposed by the `sv create` command.

jycouet:
Yes, we should probably look at having an official `add` cmd for `svelte mcp`.

This is what we have in mind for now (cc @paoloricciuti )
<img width="719" height="269" alt="Image" src="https://github.com/user-attachments/assets/d71c30f7-faca-42e5-9211-4d5596418f35" />

<img width="495" height="73" alt="Image" src="https://github.com/user-attachments/assets/4bcc0b46-1dfd-4996-a0d4-43ae21029544" />

<img width="362" height="77" alt="Image" src="https://github.com/user-attachments/assets/384fbfbb-f827-45e7-906f-2dd0bc75eb81" />

PR #735, opened 2025-10-11
feat(cli): add new add-on `mcp` to configure your project

jycouet:
Closes: #734

### DIFF
--- a/documentation/docs/20-commands/20-sv-add.md
+++ b/documentation/docs/20-commands/20-sv-add.md
@@ -31,6 +31,7 @@ You can select multiple space-separated add-ons from [the list below](#Official-
 - [`drizzle`](drizzle)
 - [`eslint`](eslint)
 - [`lucia`](lucia)
+- [`mcp`](mcp)
 - [`mdsvex`](mdsvex)
 - [`paraglide`](paraglide)
 - [`playwright`](playwright)


packages/addons/mcp/index.ts
```diff
@@ -0,0 +1,140 @@
...
+			{ value: 'claude-code', label: 'claude code' },
+			{ value: 'cursor', label: 'Cursor' },
+			{ value: 'gemini', label: 'Gemini' },
+			{ value: 'opencode', label: 'opencode' },
+			{ value: 'vscode', label: 'VSCode' },
+			{ value: 'other', label: 'Other' }
+		],
+		required: true
+	})
+	// .add('scope', {
```
manuel3108: Should we remove that for now?

packages/addons/_tests/mcp/test.ts
```diff
@@ -0,0 +1,56 @@
...
+				type: 'default',
+				options: {
+					mcp: { ide: ['claude-code', 'cursor', 'gemini', 'opencode', 'vscode'], setup: 'local' }
+				}
+			}
+		],
+		browser: false,
+		// test only one as it's not depending on project variants
+		filter: (addonTestCase) => addonTestCase.variant === 'kit-ts',
+		preInstallAddon: ({ cwd }) => {
```
manuel3108, line 21: Did you try if you get there with `vitest`'s `before` or `beforeAll`?

jycouet: These hooks are happening "too late" since the setup of projects is happening before testing (before before ^^ haha).

For other tests, we didn't need `preInstallAddon` as we never tested things like what I wanted to test here: "What if there is a file existing before adding `mcp`"?

packages/addons/mcp/index.ts
```diff
@@ -0,0 +1,130 @@
+import { defineAddon, defineAddonOptions } from '@sveltejs/cli-core';
+import { parseJson } from '@sveltejs/cli-core/parsers';
+
+const options = defineAddonOptions()
+	.add('ide', {
+		question: 'What environment do you want to use?',
```
paoloricciuti: I would rephrase this with "which client are you using" but maybe is confusing?

packages/addons/mcp/index.ts
```diff
@@ -0,0 +1,130 @@
...
+			{ value: 'opencode', label: 'opencode' },
+			{ value: 'vscode', label: 'VSCode' },
+			{ value: 'other', label: 'Other' }
+		],
+		required: true
+	})
+	.add('setup', {
+		question: 'What setup you want to use?',
+		type: 'select',
+		default: 'local',
```
paoloricciuti: Personally i would default to remote at least until we start adding new features to the local one that aren't possible for the remote...the remote one will be more up to date, safer (it's not run on the user machine), and way less heavy in terms of space.

jycouet: I don't have strong preferences here, I've switch to `remote` now.
We could also have no selection by default?

paoloricciuti: How does that look interaction wise?

jycouet: Forget what I said! It's not a multiselect!!! :o
We have to have a default selection... ;)

packages/addons/mcp/index.ts
```diff
@@ -0,0 +1,130 @@
...
+				filePath: '.vscode/mcp.json'
+			},
+			other: {
+				other: true
+			}
+		};
+
+		for (const ide of options.ide) {
+			const value = configurator[ide];
+			if (!('other' in value)) {
```
manuel3108: We should probably invert this if to keep the code clean.

packages/addons/mcp/index.ts
```diff
@@ -0,0 +1,130 @@
...
+			if (options.setup === 'local') {
+				steps.push(
+					`For other clients: ${highlighter.website(`https://svelte.dev/docs/mcp/local-setup#Other-clients`)}`
+				);
+			}
+			if (options.setup === 'remote') {
+				steps.push(
+					`For other clients: ${highlighter.website(`https://svelte.dev/docs/mcp/remote-setup#Other-clients`)}`
+				);
+			}
```
manuel3108:
```suggestion
			steps.push(
				`For other clients: ${highlighter.website(`https://svelte.dev/docs/mcp/${options.setup}-setup#Other-clients`)}`
			);
```

manuel3108 (approved):

Issue #736, opened 2025-10-15
Generated Docker Compose file should use `compose.yaml` instead of `docker-compose.yml`


9las:
**Description:**
When creating a new Svelte project using the `sv` CLI and choosing to include Docker support, the generated file is named `docker-compose.yml`.

```bash
npx sv create my-app
# then select the Docker option
```

However, according to Docker’s current recommendations, the preferred filename is `compose.yaml`.

**Reference:**
[Docker documentation](https://docs.docker.com/compose/intro/compose-application-model/)

**Why this matters:**

- Docker Compose V2 supports `compose.yaml` as the standard name.
- Aligning with the official convention improves clarity and follows modern best practices.
- It also ensures consistency with other tools and official Docker examples.

**Suggested fix:**
Rename the generated file to `compose.yaml` in the project scaffolding template when the Docker option is selected.

jycouet:
Should be out now 🎉

PR #737, opened 2025-10-16
Docker Compose file is now stored in `compose.yaml`

jycouet:
Closes: #736

### DIFF
--- a/.changeset/great-mammals-buy.md
+++ b/.changeset/great-mammals-buy.md
@@ -1,0 +1,5 @@
+---
+'sv': patch
+---
+
+feat(cli): Docker Compose file is now stored in `compose.yaml` instead of `docker-compose.yml`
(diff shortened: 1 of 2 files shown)


Conduitry (changes requested):
We should be checking whether _either_ file exists before going ahead and creating `compose.yaml`. We don't want to create `compose.yaml` if someone already has a `docker-compose.yml`. While we're at it, we should maybe also check for `docker-compose.yaml`.

jycouet:
> We should be checking whether _either_ file exists before going ahead and creating `compose.yaml`. We don't want to create `compose.yaml` if someone already has a `docker-compose.yml`. While we're at it, we should maybe also check for `docker-compose.yaml`.

Thank you for looking and the feedback.
I could probably do quick fixes taking this into account, but maybe we should do one step back and rethink a few things. 

In random order:
`F.` Our tests should have failed with my small change.
`E.` When a `docker-compose.yml` existed, we don't modify it, and we don't tell anything to the user
`Y.` We now have a `yaml` parser, we could edit `xxx.yaml` config
`A.` Drizzle add should add a first db. But when you run it a second time should it add a second db, or cancel (and explain)? (we speak about it here: https://github.com/sveltejs/cli/pull/704#issuecomment-3393998727 would be great to have your view)

jycouet:
I'll close here in favor of this one: https://github.com/sveltejs/cli/pull/738
It's taking into account your requests @Conduitry 👍
Thx

PR #740, opened 2025-10-18
fix(add): dependency detection & workspace

jycouet:
Closes: #741

This issue was always there, but is surfacing now as we use `pnpm-workspace.yaml` by default now in `pnpm`.
Since it was ignoring the "root of the workspace", "@sveltejs/kit" was not detected and all addon that Requires SvelteKit were out!


manuel3108 (approved):
This is way better, thank you !

PR #747, opened 2025-10-23
feat(create): add $schema property to tsconfig.json and jsconfig.json

aster-void:
`"$schema"` property helps autocomplete for editors that do not have schema store support built-in (e.g. Helix).

TODO: create changeset


manuel3108:
Just checked `pnpm --package=typescript dlx tsc --init` and they do not add that `$schema` property. Thus we decided to leave everything as is.

PR #568, opened 2025-05-18
chore: use `parse` & `print` from `svelte/compiler`

manuel3108:
Closes https://github.com/sveltejs/cli/issues/94
Closes https://github.com/svelte-add/svelte-add/issues/193
Closes https://github.com/svelte-add/svelte-add/issues/507

This is a pretty rough poc to try and make `sv` work better with `.svelte` files. Under the hood, this uses [`svelte-ast-print`](https://github.com/xeho91/js-ast)

Step 1 (support for svelte files):
* [x] support for svelte specifc html syntax (`@attach`, self closing component tags, transitions, etc)
* [ ] check and adapt all other usages of `parseSvelte`
* [ ] make `svelte-ast-print` support indentation and quote styles (I think some of that already works)
* [ ] fix formatting mistakes in `svelte-ast-print` (first write up: https://github.com/xeho91/js-ast/pull/128)

Step 2 (reduce deps, different pr probably)
* [ ] we currently use many different libraries for parsing and serializing JS / CSS / HTML. All of that could in theory be handled by `svelte/compiler` and `svelte-ast-print`. Check which hoops we need to jump to make that happen. We could potentially drop the following deps:
  * dom-serializer
  * domhandler
  * domutils
  * htmlparser2
  * postcss

PR #751, opened 2025-10-26
chore: use `parse` & `print` from `svelte/compiler`

manuel3108:
This is basically #568 but with a proper pr description and re-implemented from scratch, as the other branch was horribly out of date.

Closes https://github.com/sveltejs/cli/issues/94
Closes https://github.com/svelte-add/svelte-add/issues/193
Closes https://github.com/svelte-add/svelte-add/issues/507

Relates https://github.com/sveltejs/svelte/pull/16188

This pr is based on #620, since svelte alreday uses `esrap@2`

Step 1 (support for svelte files):
* [x] support for svelte specifc html syntax (`@attach`, self closing component tags, transitions, etc) (NOTE: referenced svelte pr was last updated in june, everything that got released after that is not going to work right now)
* [x] check and adapt all other usages of `parseSvelte` (approximately 10 occurences)
* [x] make generated code from `print` more readable, probably needs to be done in https://github.com/sveltejs/svelte/pull/16188

Step 2 (reduce deps, different pr probably)
* ~~we currently use many different libraries for parsing and serializing JS / CSS / HTML. All of that could in theory be handled by `svelte/compiler`. Check which hoops we need to jump to make that happen. We could potentially drop the following deps:~~
  * dom-serializer
  * domhandler
  * domutils
  * htmlparser2
  * postcss

### DIFF
--- a/packages/addons/common.ts
+++ b/packages/addons/common.ts
@@ -76,8 +76,9 @@ export function addToDemoPage(existingContent: string, path: string, langTs: boo
 
 			if (!Array.isArray(hrefAttribute.value)) continue;
 
-			const hasDemo = hrefAttribute.value.find(
-				(x) => x.type === 'Text' && x.data === `/demo/${path}`
+			const hasDemo = hrefAttribute.value.some(
+				// we use includes as it could be "/demo/${path}" or "resolve("demo/${path}")" or "resolve('demo/${path}')"
+				(x) => x.type === 'Text' && x.data.includes(`/demo/${path}`)
 			);
 			if (hasDemo) {
 				return existingContent;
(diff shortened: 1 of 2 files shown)


manuel3108:
This is ready. 

Depending on how good tree-shaking works, this is going to temporarily increase our bundle size. Personally, I wouldn't bother because this is won't be a problem as soon as we tackle step 2 described above in a different PR

Edit: to be merged after #620 as this is the base for everything. Reviewing this after #620 was merged should also dramatically reduce the diff

packages/cli/lib/install.ts
```diff
@@ -155,8 +155,7 @@ async function runAddon({ addon, multiple, workspace, workspaceOptions }: RunAdd
 				fileContent = content(fileContent);
 				if (!fileContent) return fileContent;
 
-				// TODO: fix https://github.com/rolldown/tsdown/issues/575 to remove the `replaceAll`
-				writeFile(workspace, path, fileContent.replaceAll('<\\/script>', '</script>'));
```
jycouet, line 159: This is still needed
We can still remove the test that is skipped as new snapshot tests covers well this.

jycouet (changes requested):

packages/cli/lib/install.ts
manuel3108: Oh yeah, i had something wrong in mind. Will re-add over the next few days.

jycouet (approved):
LGTM :)

Issue #702, opened 2025-09-11
ESLint `svelte/no-navigation-without-resolve` errors in demo


hyunbinseo:
The demo template now has multiple ESLint errors.

Caused by [eslint-plugin-svelte@3.12.0](https://github.com/sveltejs/eslint-plugin-svelte/releases/tag/eslint-plugin-svelte%403.12.0) update.

- https://github.com/sveltejs/eslint-plugin-svelte/pull/1308

```svelte
<!-- Found a link with a url that isn't resolved -->
<a href="/">Home</a>
<a href="/about">About</a>
<a href="/sverdle">Sverdle</a>
```

```shell
pnpm dlx sv create .
```

```
┌  Welcome to the Svelte CLI! (v0.9.4)
│
◇  Which template would you like?
│  SvelteKit demo
│
◇  Add type checking with TypeScript?
│  Yes, using TypeScript syntax
│
◆  Project created
│
◇  What would you like to add to your project? (use arrow keys / space bar)
│  eslint
│
◆  Successfully setup add-ons
│
◇  Which package manager do you want to install dependencies with?
│  pnpm
```

```
D:\_\src\routes\Header.svelte
  20:14  error  Found a link with a url that isn't resolved  svelte/no-navigation-without-resolve
  23:14  error  Found a link with a url that isn't resolved  svelte/no-navigation-without-resolve
  26:14  error  Found a link with a url that isn't resolved  svelte/no-navigation-without-resolve

D:\_\src\routes\about\+page.svelte
  23:16  error  Found a link with a url that isn't resolved  svelte/no-navigation-without-resolve

D:\_\src\routes\sverdle\+page.svelte
  107:31  error  Found a link with a url that isn't resolved  svelte/no-navigation-without-resolve

✖ 5 problems (5 errors, 0 warnings)
```

jycouet:
Yes, you are right. You want to do it? 

_If not probably can in a few hours._

hyunbinseo:
I did leave a comment on the ESLint plugin repo. https://github.com/sveltejs/eslint-plugin-svelte/pull/1308#issuecomment-3279043029

Making `<a href="/">` an error seems too strict, and even causes problem with the tutorial.

jycouet:
If you have this rules, I think it's fair to do it eveywhere, even `<a href="/">` should be `<a href={resolve('/')}>`. No ? To strict ?
Or the rule should accept some exceptions? `/` `/about` ?

[...]

PR #754, opened 2025-10-26
feat(cli): wrap links with `resolve()`

jycouet:
Closes: #702

### DIFF
--- a/packages/addons/common.ts
+++ b/packages/addons/common.ts
@@ -67,7 +67,7 @@ export function addToDemoPage(existingContent: string, path: string, typescript:
 	const { script, template, generateCode } = parseSvelte(existingContent, { typescript });
 
 	for (const node of template.ast.childNodes) {
-		// we use inclules as it could be "/demo/${path}" or "resolve("demo/${path}")" or "resolve('demo/${path}')"
+		// we use includes as it could be "/demo/${path}" or "resolve("demo/${path}")" or "resolve('demo/${path}')"
 		if (node.type === 'tag' && node.attribs['href'].includes(`/demo/${path}`)) {
 			return existingContent;
 		}


manuel3108:
LGTM, should we also include the fixes for the two addons here to make it consistent?

```svelte
// routes/demo/+page.svelte
<a href="/demo/paraglide">paraglide</a>
<a href="/demo/lucia">lucia</a>
```

jycouet:
Should be better 👍

I was thinking that it would have been harder ! But nop 😁

packages/addons/common.ts
```diff
@@ -63,18 +63,22 @@ export function addEslintConfigPrettier(content: string): string {
...
 
-export function addToDemoPage(content: string, path: string): string {
-	const { template, generateCode } = parseSvelte(content);
+export function addToDemoPage(existingContent: string, path: string, typescript: boolean): string {
+	const { script, template, generateCode } = parseSvelte(existingContent, { typescript });
 
 	for (const node of template.ast.childNodes) {
-		if (node.type === 'tag' && node.attribs['href'] === `/demo/${path}`) {
-			return content;
+		// we use inclules as it could be "/demo/${path}" or "resolve("demo/${path}")" or "resolve('demo/${path}')"
```
sacrosanctic: typo

jycouet: Where?
I'm looking everywhere, but but find it 🫣

manuel3108: "inclules"

jycouet: I should let AI write more code & comments! For sure it will not do these jyc mistakes 😅

Should be good

manuel3108 (approved):

Issue #768, opened 2025-11-06
Ading local MCP for OpenCode produces wrong config file


mastermakrela:
Currently, adding the MCP for `opencode` using:

```zsh
❯ sv add mcp

┌ Welcome to the Svelte CLI! (v0.9.11)
│
◇ Which client would you like to use?
│ opencode
│
◇ What setup do you want to use?
│ Local
│
◆ Successfully set up add-ons
│
◇ Which package manager do you want to install dependencies with?
│ bun
│
◆ Successfully installed dependencies
│
◇ Successfully formatted modified files
│
└ You're all set!
```

creates the following configuration:

```json
{
	"$schema": "https://opencode.ai/config.json",
	"mcp": {
		"svelte": {
			"type": "local",
			"command": "npx",
			"args": ["-y", "@sveltejs/mcp"]
		}
	}
}
```

which leads to the following error:

```zsh
❯ opencode
Error: Config file at /path/to/project/opencode.json is invalid
↳ Invalid input: expected array, received string mcp.svelte.command
↳ Unrecognized key: "args" mcp.svelte
```

If we open [`https://opencode.ai/config.json`](https://opencode.ai/config.json), we can see that the schema defines the following structure for a local MCP configuration:

```json
"mcp": {
  "description": "MCP (Model Context Protocol) server configurations",
  "type": "object",
  "propertyNames": {
    "type": "string"
  },
  "additionalProperties": {
    "anyOf": [
      {
        "ref": "McpLocalConfig",
        "type": "object",
        "properties": {
          "type": {
            "description": "Type of MCP server connection",
            "type": "string",
            "const": "local"
          },
          "command": {
            "description": "Command and arguments to run the MCP server",
            "type": "array",
            "items": {
              "type": "string"
            }
          },
          "environment": {
            "description": "Environment variables to set when running the MCP server",
            "type": "object",
            "propertyNames": {
              "type": "string"
            },
            "additionalProperties": {
              "type": "string"
            }
          },
          "enabled": {
            "description": "Enable or disable the MCP server on startup",
            "type": "boolean"
          },
          "timeout": {
            "description": "Timeout in ms for fetching tools from the MCP server. Defaults to 5000 (5 seconds) if not specified.",
            "type": "integer",
            "exclusiveMinimum": 0,
            "maximum": 9007199254740991
          }
        },
        "required": ["type", "command"],
        "additionalProperties": false
      },
      ...
    ]
  }
}
```

According to the schema, the correct local configuration should be:

[...]

PR #769, opened 2025-11-06
fix(mcp): generate valid `mcp` local configuration for `opencode`

mastermakrela:
This patch fixes **issue #768**, where `sv add mcp` generated an invalid configuration for the `opencode` client.  
Previously, the addon produced a malformed config using `"command": "npx"` and a separate `args` field, which caused schema validation errors.

Fixes: #768 

#### Changes

- Updated `getLocalConfig` to support both `command` and `args` as flexible inputs.
- Default `command` for local setup for opencode is now correctly defined as an array: `["npx", "-y", "@sveltejs/mcp"]`; for other IDEs nothing has changed
- Made `args` optional, to avoid schema warning in OpenCode, i.e. when using array-style commands.

#### Notes

- I’ve looked at the contributing guidelines, and if I understand correctly, a changeset isn’t needed since this is part of an addon.  
  If I'm wrong - I'll generate one :)
- I noticed there’s already an MCP test, but it doesn’t verify the generated output for specific IDEs. But if wanted, I can create one.

#### Result

The generated `opencode.json` now matches the official schema at  
[`https://opencode.ai/config.json`](https://opencode.ai/config.json), preventing the “expected array, received string” error when running `opencode`.

| Before                                                                                                                                            | After                                                                                                                                             |
| ------------------------------------------------------------------------------------------------------------------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------- |
| Screenshot 2025-11-06 at 22 35 38 | Screenshot 2025-11-06 at 22 36 13 |

### DIFF
--- a/.changeset/hungry-maps-poke.md
+++ b/.changeset/hungry-maps-poke.md
@@ -1,0 +1,5 @@
+---
+"sv": patch
+---
+
+fix(mcp): generate valid `mcp` local configuration for `opencode`


manuel3108:
Looks correct according to their [docs](https://opencode.ai/docs/mcp-servers), thank you!
@jycouet Any idea if the docs were different when this addon was created, or which docs we were following?

jycouet:
Thank you for doing this. 
I added a test to save snapshots, it will be easier to see if things change.

> Any idea if the docs were different when this addon was created, or which docs we were following

I looked at links again, I think it's just me that didn't read well! >< Too many at the same time ^^

---

What do you think?

manuel3108 (approved):
Sounds good, added a changeset, lgtm

PR #771, opened 2025-11-07
fix(cli): Check existing conditions for specified options

### LINKED ISSUE #770: `runAddCommand` doesn't actually test for conditions
When the user runs
```sh
npx sv add drizzle=database:sqlite+client:libsql
```
The user will get:
```
┌  Welcome to the Svelte CLI! (v0.9.9)
│
■  Error: Incompatible 'drizzle' option specified: 'libsql'
```

This is because the following lines of code:
https://github.com/sveltejs/cli/blob/d4640d8007663a6b19879f875a84353e886311b3/packages/cli/commands/add/index.ts#L238-L240
Will *only* check if the `id` or `group` matches, which unfortunately means that with our current `drizzle` definitions:
https://github.com/sveltejs/cli/blob/d4640d8007663a6b19879f875a84353e886311b3/packages/addons/drizzle/index.ts#L27-L60
`.find` will return the first match... which is always `postgresql`, because it has `group: 'client'` and happens to be declared first.
This means `npx sv add drizzle=...` only ever works for `postgresql`, and not `mysql` or `sqlite`.

The fix is to always check the `condition` is valid when it exists:
```diff
 const optionEntries = Object.entries(details.options);
+const specifiedOptionsObject = Object.fromEntries(
+        specifiedOptions.map((option) => option.split(':', 2))
+);
 for (const option of specifiedOptions) {
         let [optionId, optionValue] = option.split(':', 2);
 
         // validates that the option exists
-        const optionEntry = optionEntries.find(
-                ([id, question]) => id === optionId || question.group === optionId
-        );
+        const optionEntry = optionEntries.find(([id, question]) => {
+                // simple ID match
+                if (id === optionId) return true;
+
+                // group match - need to check conditions and value validity
+                if (question.group === optionId) {
+                        // does the value exist for this option?
+                        if (question.type === 'select' || question.type === 'multiselect') {
+                                const isValidValue = question.options.some((opt) => opt.value === optionValue);
+                                if (!isValidValue) return false;
+                        }
+
+                        // if there's a condition, does it pass?
+                        if (question.condition) {
+                                return question.condition(specifiedOptionsObject);
+                        }
+
+                        // finally, unconditional
+                        return true;
+                }
+
+                // unrecognized optionId
+                return false;
+        });
+
         if (!optionEntry) {
                 const { choices } = getOptionChoices(details);
                 throw new Error(
```

GrygrFlzr:
Closes #770 by making sure we actually apply conditions if they exist

### DIFF
--- a/packages/cli/commands/add/index.ts
+++ b/packages/cli/commands/add/index.ts
@@ -231,13 +231,38 @@ export async function runAddCommand(
 		official[addonId] ??= {};
 
 		const optionEntries = Object.entries(details.options);
+		const specifiedOptionsObject = Object.fromEntries(
+			specifiedOptions.map((option) => option.split(':', 2))
+		);
 		for (const option of specifiedOptions) {
 			let [optionId, optionValue] = option.split(':', 2);
 
 			// validates that the option exists
-			const optionEntry = optionEntries.find(
-				([id, question]) => id === optionId || question.group === optionId
-			);
+			const optionEntry = optionEntries.find(([id, question]) => {
+				// simple ID match
+				if (id === optionId) return true;
+
+				// group match - need to check conditions and value validity
+				if (question.group === optionId) {
+					// does the value exist for this option?
+					if (question.type === 'select' || question.type === 'multiselect') {
+						const isValidValue = question.options.some((opt) => opt.value === optionValue);
+						if (!isValidValue) return false;
+					}
+
+					// if there's a condition, does it pass?
+					if (question.condition) {
+						return question.condition(specifiedOptionsObject);
+					}
+
+					// finally, unconditional
+					return true;
+				}
+
+				// unrecognized optionId
+				return false;
+			});
+
 			if (!optionEntry) {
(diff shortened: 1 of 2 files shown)


packages/cli/commands/add/index.ts
```diff
@@ -231,13 +231,38 @@ export async function runAddCommand(
 		official[addonId] ??= {};
 
 		const optionEntries = Object.entries(details.options);
+		const specifiedOptionsObject = Object.fromEntries(
+			specifiedOptions.map((option) => option.split(':', 2))
+		);
 		for (const option of specifiedOptions) {
 			let [optionId, optionValue] = option.split(':', 2);
```
GrygrFlzr, line 238: We could instead do
```ts
for (const option of Object.entries(specifiedOptionsObject)) {
    let [optionId, optionValue] = option;
```
But I'm not sure that's clearer intent.

manuel3108 (approved):
This is perfect, thank you!

PR #774, opened 2025-11-08
fix: add `$schema` to gemini

sacrosanctic:
> Tip: JSON-aware editors can use autocomplete and validation by pointing to the generated schema at schemas/settings.schema.json in this repository. When working outside the repo, reference the hosted schema at https://raw.githubusercontent.com/google-gemini/gemini-cli/main/schemas/settings.schema.json.

https://github.com/google-gemini/gemini-cli/blob/main/docs/get-started/configuration.md#settings-files


sacrosanctic:
> * Do they publish this schema in a more "official" url ? Or it's the only option ?

I couldn't find a more official one, that's also where they link to when you check the `settings.schema.json` itself.

Issue #544, opened 2025-04-23
Set `build.cssTarget` for Tailwind CSS add-on


hyunbinseo:
Tailwind CSS now requires Chrome 111+

> Tailwind CSS v4.0 is designed for and tested on modern browsers, and the core functionality of the framework specifically depends on these browser versions:
> 
> -   **Chrome 111** _(released March 2023)_
> -   **Safari 16.4** _(released March 2023)_
> -   **Firefox 128** _(released July 2024)_
>
> https://tailwindcss.com/docs/compatibility

However, SvelteKit currently builds on `target: 'modules'` which targets Chrome 87:

> Browser compatibility target for the final bundle. The default value is a Vite special value, `'modules'`, which targets browsers with native ES Modules, native ESM dynamic import, and `import.meta` support. Vite will replace `'modules'` to `['es2020', 'edge88', 'firefox78', 'chrome87', 'safari14']`
>
> https://vite.dev/config/build-options#build-target

This causes Vite to target older browsers with its CSS build output.

This can be fixed by setting Vite's `build.cssTarget` out-of-the box.

```ts
import { sveltekit } from '@sveltejs/kit/vite';
import tailwindcss from '@tailwindcss/vite';
import { defineConfig } from 'vite';

export default defineConfig({
  build: { cssTarget: ['firefox128', 'chrome111', 'safari16.4'] },
  plugins: [tailwindcss(), sveltekit()],
});
```

> [!NOTE]  
> Even with this setting, browser native CSS nesting is not flattened since it is supported from Safari 17.2. Reference https://github.com/sveltejs/kit/issues/13732

jycouet:
It looks like today's `vite` `build.target` is closer to tailwind v4 (following the link you used).
We could specify `cssTarget` as you mentioned, but I'm a bit worried to keep it up-to-date with tailwind doc.

In Tailwind docs, when they setup the vite plugin, they don't advertise this. Any idea why?

hyunbinseo:
`vite@7`'s new [baseline target](https://vite.dev/blog/announcing-vite7.html#default-browser-target-changed-to-baseline-widely-available) does make it closer.

I also wonder why the [Tailwind docs](https://tailwindcss.com/docs/installation/using-vite) does not mention `build.cssTarget`.

NOTE: `vite@7`'s baseline target [is static](https://github.com/hyunbinseo/svelte-kitty/issues/32#issuecomment-3363866032) and does not change until `vite@8`. The current value flattens nested CSS.

[...]

PR #780, opened 2025-11-13
feat(cli): co-locate `css` file from usage (`+layout.svelte`)

sacrosanctic:
@benmccann @jycouet following up from the discussion on Discord.

In this PR:
- co-locate css file for kit projects
- add vscode setting for tailwind ([ref](https://github.com/tailwindlabs/tailwindcss/discussions/13881))

### DIFF
--- a/community-addon-template/vitest.config.js
+++ b/community-addon-template/vitest.config.js
@@ -6,8 +6,8 @@ export default defineConfig({
 	test: {
 		include: ['tests/**/*.test.{js,ts}'],
 		exclude: ['tests/setup/*'],
-		testTimeout: ONE_MINUTE * 4,
-		hookTimeout: ONE_MINUTE * 4,
+		testTimeout: ONE_MINUTE * 3,
+		hookTimeout: ONE_MINUTE * 3,
 		globalSetup: ['tests/setup/global.ts'],
 		expect: {
 			requireAssertions: true
--- a/packages/addons/vitest.config.ts
+++ b/packages/addons/vitest.config.ts
@@ -8,8 +8,8 @@ export default defineProject({
 		name: 'addons',
 		include: ['_tests/**/test.{js,ts}'],
 		globalSetup: ['_tests/_setup/global.ts'],
-		testTimeout: ONE_MINUTE * 4,
-		hookTimeout: ONE_MINUTE * 4,
+		testTimeout: ONE_MINUTE * 3,
+		hookTimeout: ONE_MINUTE * 3,
 		retry: env.CI ? 3 : 0,
 		expect: {
 			requireAssertions: true


packages/addons/tailwindcss/index.ts
```diff
@@ -120,6 +120,15 @@ export default defineAddon({
 			});
 		}
 
+		sv.file('.vscode/settings.json', (content) => {
```
jycouet, line 133: I'm not sure how should we deal with all different `ide` support.

A/ Should we have a list of "offical support"?
B/ Should we add a question? Or just try to smart detect (like `TERM_PROGRAM=vscode`)

@manuel3108 @benmccann any insights?

packages/addons/tailwindcss/index.ts
```diff
@@ -130,7 +139,7 @@ export default defineAddon({
 
 				if (!plugins.includes(PLUGIN_NAME)) plugins.push(PLUGIN_NAME);
 
-				data.tailwindStylesheet ??= './src/app.css';
+				data.tailwindStylesheet ??= kit ? './src/routes/layout.css' : './src/app.css';
```
jycouet:
```suggestion
				data.tailwindStylesheet ??= kit ? `${kit?.routesDirectory}/layout.css` : './src/app.css';
```

.changeset/rich-tires-talk.md
```diff
@@ -0,0 +1,5 @@
+---
+'sv': patch
+---
+
+co-locate the css file from `tailwind`
```
jycouet:
```suggestion
feat(cli): co-locate css file from usage (`+layout.svelte`)

**Review by jycouet — CHANGES_REQUESTED** [review]
Thx for the PR
I think that we should update name & location of `packages/create/templates/demo/src/app.css` as well to be consistent

Also https://github.com/sveltejs/cli/blob/4e5e2f3dc0e59f2fe5a41665bfb3336766313cb5/packages/addons/prettier/index.ts#L56 needs to be updated.

File: packages/addons/tailwindcss/index.ts (hunk 1 above)
Comment (sacrosanctic) [reply packages/addons/tailwindcss/index.ts:133]: https://github.com/sacrosanctic/cli/blob/add-agent-md/packages/addons/eslint/index.ts#L44

Eslint doesn't have special logic either. This can be a seperate PR

Comment (manuel3108) [reply packages/addons/tailwindcss/index.ts:133]: Interesting. I thought that in the past we were always against adding any editor-specific coding like this. But since we already have this for eslint, i don't in adding it here, as nobody has complained about that implementation in the past. 

Let's leave it like this for now!

**Comment by manuel3108** [comment]
@jycouet I updated the docs and added a second changeset as I didn't manage to combine both features into one good changeset. Let me know what you think

File: packages/addons/tailwindcss/index.ts (hunk 1 above)
Comment (jycouet) [reply packages/addons/tailwindcss/index.ts:133]: All right 👍
We can move with https://github.com/sveltejs/cli/issues/626 as well later then

**Comment by hyunbinseo** [comment]
Is there a reason why the leading `./` is removed?

```diff
- "tailwindStylesheet": "./src/app.css",
+ "tailwindStylesheet": "src/routes/layout.css",
```

The Prettier [Prettier plugin does](https://github.com/tailwindlabs/prettier-plugin-tailwindcss#options) uses a prefixed one:

```jsonc
// .prettierrc
{
  "tailwindStylesheet": "./resources/css/app.css"
}
```

sacrosanctic:
@hyunbinseo did your app break?

hyunbinseo:
@sacrosanctic I don't think the new template is broken.

> Note that paths are resolved relative to the Prettier configuration file.

But I think it's best to follow the official setup. Similar discussion here: https://github.com/sveltejs/cli/issues/544#issuecomment-3454918659

PR #781, opened 2025-11-16
chore: storybook new cli ux

manuel3108:
Contains minor changes to the way we test the storbook integration. Especially to make CI testing faster. 

Currently uses a canary release. DO NOT MERGE.

### DIFF
--- a/packages/addons/storybook/index.ts
+++ b/packages/addons/storybook/index.ts
@@ -2,7 +2,7 @@ import process from 'node:process';
 import { defineAddon } from '@sveltejs/cli-core';
 import { getNodeTypesVersion } from '../common.ts';
 
-export const STORYBOOK_VERSION = '0.0.0-pr-32717-sha-f340a68b';
+export const STORYBOOK_VERSION = '10.1.0';
 
 export default defineAddon({
 	id: 'storybook',


packages/addons/storybook/index.ts
```diff
@@ -12,7 +12,7 @@ export default defineAddon({
 		runsAfter('eslint');
 	},
 	run: async ({ sv }) => {
-		const args = ['create-storybook@latest', '--skip-install', '--no-dev'];
+		const args = ['create-storybook@0.0.0-pr-32984-sha-9ba02343', '--skip-install', '--no-dev'];
```
JReinhold:
```suggestion
		const args = ['create-storybook@0.0.0-pr-32717-sha-f340a68b', '--skip-install', '--no-dev'];
```

packages/addons/storybook/index.ts
```diff
@@ -12,7 +14,13 @@ export default defineAddon({
 		runsAfter('eslint');
 	},
 	run: async ({ sv }) => {
-		const args = ['create-storybook@latest', '--skip-install', '--no-dev'];
+		const args = [
+			`create-storybook@${STORYBOOK_VERSION}`,
+			'--skip-install',
+			'--no-dev',
+			'--features',
+			'docs'
```
manuel3108: Should be `--no-features` in the future according to Jeppe

manuel3108:
Since this is mainly related to our testing and no user facing code was change, merging without changeset

packages/addons/storybook/index.ts
```diff
@@ -2,6 +2,8 @@ import process from 'node:process';
 import { defineAddon } from '@sveltejs/cli-core';
 import { getNodeTypesVersion } from '../common.ts';
 
+export const STORYBOOK_VERSION = '10.1.0';
```
JReinhold, line 5: I think this was a mistake, as it now version locks the Storybook version, instead of just getting the latest.

So when you run `sv` now with Storybook, you get a warning:

Image

should probably revert back to `latest`.

packages/addons/storybook/index.ts
```diff
@@ -12,7 +14,12 @@ export default defineAddon({
 		runsAfter('eslint');
 	},
 	run: async ({ sv }) => {
-		const args = ['create-storybook@latest', '--skip-install', '--no-dev'];
+		const args = [
+			`create-storybook@${STORYBOOK_VERSION}`,
+			'--skip-install',
+			'--no-dev',
+			'--no-features'
```
JReinhold, line 21: I don't think this should be in production, I think users should be able to choose the feature they want. Storybook will prompt them when this flag is not set.

When we discussed this, I thought you only put in the flag in your tests, to speed them up.

PR #796, opened 2025-11-18
fix: use correct file association for tailwind css

nebwards:
#780 incorrectly set the language id for tailwind css file association as `tailwind`. 

The correct language id is `tailwindcss` as defined by the tailwindcss-intellisense plugin:
https://github.com/tailwindlabs/tailwindcss-intellisense/blob/072a6c96e8067805e4b9b916e7b382d30e2ec0d7/packages/vscode-tailwindcss/package.json#L47

### DIFF
--- a/.changeset/cruel-pets-obey.md
+++ b/.changeset/cruel-pets-obey.md
@@ -1,0 +1,5 @@
+---
+'sv': patch
+---
+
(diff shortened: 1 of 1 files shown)


jycouet (approved):
Yes, source: https://marketplace.visualstudio.com/items?itemName=bradlc.vscode-tailwindcss#recommended-vs-code-settings

Thx!

PR #810, opened 2025-11-22
fix(cli): `kit` projects were detected incorrectly

sacrosanctic:
closes #809

PR #812, opened 2025-11-22
fix: ts detection

sacrosanctic:
Initially brought up in a [discord thread](https://discord.com/channels/457912077277855764/1441488556492324944).

caused by #695

### DIFF
--- a/packages/cli/commands/create.ts
+++ b/packages/cli/commands/create.ts
@@ -37,8 +37,8 @@ import {
 	sanitizeAddons,
 	type SelectedAddon
 } from './add/index.ts';
-import { createWorkspace } from './add/workspace.ts';
 import { commonFilePaths } from './add/utils.ts';
+import { createWorkspace } from './add/workspace.ts';
 
 const langs = ['ts', 'jsdoc'] as const;
 const langMap: Record<string, LanguageType | undefined> = {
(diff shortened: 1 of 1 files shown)


sacrosanctic:
The fix involves repeating logic. Please advice. I don't understand the purpose of `virtualWorkspace` as it seems to overwrite correct settings.

jycouet:
> The fix involves repeating logic. Please advice. I don't understand the purpose of `virtualWorkspace` as it seems to overwrite correct settings.

If you look at the root issue https://github.com/sveltejs/cli/issues/689 the first part is why we started thinking "virtualWorkspace".
I'll continue your work on the other PR https://github.com/sveltejs/cli/pull/810
Let's see

sacrosanctic:
I've rebased the changes from #810, the bug still persists.

For now. I've removed the ts check.

This also brings up another issue. `create` doesn't use `svelte.config.ts` when typescript is selected. I think it should be addressed in a separate PR.

manuel3108:
Thanks for all the research. I've made the diff even smaller by removing the `svelte.config.ts` fo now, as we never use this. The method where this was happening, is only used during `create` and not during `add` so this is fine. I opened #816 for that. 

But you brought up a excellent point regarding the `dependencyVersion()`. We will need to discuss how to handle that!

manuel3108 (approved):

PR #822, opened 2025-11-25
chore: remove `esModuleInterop` from library template

hyunbinseo:
> [ES Module Interop](https://www.typescriptlang.org/tsconfig/#esModuleInterop) is `true` if [`module`](https://www.typescriptlang.org/tsconfig/#module) is `node16`, `nodenext`, or `preserve`; `false` otherwise.

For the library templates the module is set to `nodenext`, so this is a duplicate.

### DIFF
--- a/.changeset/old-tires-lay.md
+++ b/.changeset/old-tires-lay.md
@@ -2,4 +2,4 @@
 'sv': patch
 ---
 
-chore: remove `esModuleInterop` from library template
+chore(create): remove `esModuleInterop` from library template


.changeset/old-tires-lay.md
```diff
@@ -0,0 +1,5 @@
+---
+'sv': patch
+---
+
+chore: remove `esModuleInterop` from library template
```
manuel3108:
```suggestion
chore(create): remove `esModuleInterop` from library template
```

PR #823, opened 2025-11-25
fix(cli): reload workspace before executing each addon

### LINKED ISSUE #821: `prettier-plugin-tailwindcss` is not added automatically
When the following commands are executed:

```shell
pnpm dlx sv create --template minimal --add eslint prettier tailwindcss="plugins:forms" --types jsdoc javascript
pnpm dlx sv create --template minimal --add eslint prettier tailwindcss="plugins:forms" --types ts typescript
```

- `eslint-config-prettier` and `prettier-plugin-tailwindcss` are not installed
- `pnpm format` does not seem to run after creation (or silently fails?)

For diff, see https://github.com/hyunbinseo/svelte-kit-templates/commit/d4c563a8fab4a2c531dc31b2c59e203b73ac68e5

Is this the desired action for `create --add` in one go? Or is it broken?

- https://github.com/sveltejs/cli/pull/756

manuel3108:
Closes #821

### DIFF
--- a/.changeset/afraid-regions-roll.md
+++ b/.changeset/afraid-regions-roll.md
@@ -1,0 +1,5 @@
+---
+'sv': patch
+---
+
+fix(cli): reload workspace before executing each addon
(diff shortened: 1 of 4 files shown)


manuel3108:
Merging, since a lot of the addons that have dependencies on each other should be broken by this.

Issue #825, opened 2025-11-27
Svelte cli installs prettier extension that prevents formatting


Varjovaras:
The Svelte CLI (npx sv create my-app) setup that includes both Prettier and Tailwind CSS currently installs a prettier-plugin-tailwindcss version that causes Prettier formatting to fail on .svelte files.

This failure occurs regardless of whether the file contains Tailwind classes.

Steps to Reproduce

Initialize a new SvelteKit project with tailwind and prettier.

Attempt to format a .svelte file (e.g., src/routes/+layout.svelte) in an editor using the local Prettier configuration (VS Code, Zed, or Vim).

I tested and the previous sv version 0.10.4 did not install the troublesome prettier plugin and it works fine.

Environment

Svelte CLI version: 0.10.5

sacrosanctic:
I cannot reproduce.

```sh
pnpm dlx sv@0.10.5 create --template minimal --types ts --add tailwindcss="plugins:none" prettier --install pnpm ./
echo "<script lang='ts'>let five='5'</script><button class=\"bg-red-500 p-5\">red</button>" > src/routes/Test.svelte
pnpm format
```

Varjovaras:
Running the exact same commands gives me this error which is the one I got earlier too. Only way I found to fix it was to remove the prettier-plugin-tailwindcss even though the error doesn't seem like it at first.

```[error] src/routes/+layout.svelte: TypeError: getVisitorKeys is not a function or its return value is not iterable
[error]     at recurse (file:///Users/kristjan/Documents/github/hcbull/node_modules/.pnpm/prettier@3.7.1/node_modules/prettier/index.mjs:17713:24)
[error]     at printEmbeddedLanguages (file:///Users/kristjan/Documents/github/hcbull/node_modules/.pnpm/prettier@3.7.1/node_modules/prettier/index.mjs:17683:3)
[error]     at printAstToDoc (file:///Users/kristjan/Documents/github/hcbull/node_modules/.pnpm/prettier@3.7.1/node_modules/prettier/index.mjs:17792:9)
[error]     at async coreFormat (file:///Users/kristjan/Documents/github/hcbull/node_modules/.pnpm/prettier@3.7.1/node_modules/prettier/index.mjs:18171:14)
[error]     at async formatWithCursor (file:///Users/kristjan/Documents/github/hcbull/node_modules/.pnpm/prettier@3.7.1/node_modules/prettier/index.mjs:18384:14)
[...]

PR #827, opened 2025-11-29
fix(cli): files will be formatted after create

manuel3108:
fixes https://github.com/sveltejs/cli/issues/821

relates: https://github.com/sveltejs/cli/issues/825

### DIFF
--- a/packages/cli/commands/create.ts
+++ b/packages/cli/commands/create.ts
@@ -37,7 +37,7 @@ import {
 	sanitizeAddons,
 	type SelectedAddon
 } from './add/index.ts';
-import { commonFilePaths, getPackageJson } from './add/utils.ts';
+import { commonFilePaths, formatFiles, getPackageJson } from './add/utils.ts';
 import { createWorkspace } from './add/workspace.ts';
 import { dist } from '../../create/utils.ts';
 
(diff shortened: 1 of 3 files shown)


packages/cli/commands/create.ts
```diff
@@ -272,7 +272,7 @@ async function createProject(cwd: ProjectPath, options: Options) {
 			answersCommunity,
 			options: {
 				cwd: projectPath,
-				install: false,
```
jycouet, line 275: We need to keep `false` here to avoid double install.

You can see it here:
```
npx sv create --template minimal --types ts --add eslint --install npm jyc
```

manuel3108: <details><summary>Old comment</summary>
<p>

We can't otherwise the files are not formatted. 

For me this does not result in an wrongly printed:

input:
```shell
pnpm sv create --template minimal --add eslint prettier tailwindcss="plugins:forms" --types ts --no-dir-check ../cli-temp --install pnpm
```

output: 
```
pnpm dlx sv create --template minimal --types ts --add eslint prettier tailwindcss="plugins:forms" --install pnpm ../cli-temp
```

</p>
</details> 

ohhhh i seee. It's not about the printed command but about the prompts.

jycouet: I mean, it will install 2 times

image

manuel3108: Ok, i gave this a try for a while, but i don't think there is a viable solution here without reworking a bunch of code. I also noticed that `formatFiles` is currently (not this branch, but last release) only called during `add`, as the deps for `create` are installed after the formatting should have run.

If you find a viable solution without changing too much let me know. Otherwise i would suggest rewriting a part of the `cli` directory to make it more future proof and make it more stable and testable based on our current requirements.

PR #831, opened 2025-11-30
add `create-with-all-addons` snapshot

### LINKED ISSUE #830: snapshot testing addons
Regarding #828 I don't think we currently confidence in our testing setup to confidently do this. We should create at least one test which runs all addons with all options and make sure these snapshots do not break while migrating.

jycouet:
Closes #830 

To regenerate the snapshot delete the folder `packages/cli/tests/create-with-all-addons`


jycouet:
Right now it's without `format` _(faster!)_.

I tried it... And I'm sure we can tune a few defaults. Maybe later?

packages/cli/tests/cli.ts
```diff
@@ -34,40 +34,96 @@ describe('cli', () => {
...
+					'packages',
+					'cli',
+					'tests',
+					'create-with-all-addons'
+				);
+
+				// How to update this snapshot?
+				// rm -rf packages/cli/tests/create-with-all-addons
+				// trigger tests again... and check the diff!
+				if (!fs.existsSync(snapPath)) {
```
manuel3108: Maybe I am missing something, but this seems overcomplicated. Why are you not using something like `await expect(output).toMatchFileSnapshot(`${testDirectoryPath}/output.ts`);` directly?

https://github.com/sveltejs/cli/blob/e35827cc70d0cf33899a91f76d37bf62f1f05398/packages/core/tests/js/index.ts#L27

That way, if you run `vitest` in dev mode and press `u`, you can update the snapshots directly.

jycouet: That's too simple!!! haha
I didn't think about it, thx for the suggestion 🎉

packages/cli/tests/cli.ts
```diff
@@ -34,40 +34,70 @@ describe('cli', () => {
...
+				'packages',
+				'cli',
+				'tests',
+				'snapshots',
+				projectName
+			);
+			const relativeFiles = fs.readdirSync(testOutputPath, { recursive: true }) as string[];
+			for (const relativeFile of relativeFiles) {
+				if (!fs.statSync(path.resolve(testOutputPath, relativeFile)).isFile()) continue;
+				if (['.svg', '.env'].some((ext) => relativeFile.endsWith(ext))) continue;
```
manuel3108, line 88: any reason for not comparing them? i mean it's likely they won't change much and will probably not provide any notable benefit, but I wonder what's more confusing - just adding them - or ignoring them

jycouet: For `.env`, since they are not committed, but generated, then they are part of test but not snapshot, so it's failing the test. I started to tweak gitignore, to have it committed there, but it's against the generated gitignore... So we would have to tweak this as well... All in all, I think skipping just the `.env` is a smarter solution.

For the `.svg`, first I think that we don't really care about the content of the svg in `sv`, second, I started to play with storybook, and there, I had some `png`, `svg` and co... And it was taking a lot of time. So it was mainly to speedup things.

PR #838, opened 2025-12-06
chore: update project structure

jycouet:
- [x] moved folders with `git mv ...` to preserve git history (please pull and check!)

## Gist of the PR
- [x] `packages/cli` -> `packages/sv`
- [x] `packages/core` -> `packages/sv/lib/core`
- [x] `packages/create` -> `packages/sv/lib/create`
- [x] `packages/addons` -> `packages/sv/lib/addons`
- [x] `packages/cli/commands` -> `packages/sv/lib/cli`

Preparing work for `community` ;)

### DIFF
--- a/.changeset/loud-gifts-hang.md
+++ b/.changeset/loud-gifts-hang.md
@@ -1,5 +1,0 @@
----
-'sv': patch
----
-
-chore: update internal project structure


.changeset/loud-gifts-hang.md
```diff
@@ -0,0 +1,5 @@
+---
+'sv': patch
+---
+
+chore: update internal project structure
```
manuel3108: We merged a bunch of prs without a changeset that were just internal. I think we should do the same here.

Issue #828, opened 2025-11-30
reduce dependency count


manuel3108:
Now that #751 is pretty ready, one of the next jobs should be to tackle remove a bunch of the dependencies and replacing what they were doing with the svelte compiler. Namely:
* dom-serializer
* domhandler
* domutils
* htmlparser2
* ~postcss~

This should dramatically reduce bundle size as well.

PR #839, opened 2025-12-07
chore: drop postcss dep

manuel3108:
Relates #828 
Will do a couple more over the next week for the other AST's. I was initially thinking of creating one giant pr, but looks like we can properly separate them.

Remove the test for comments, because I don't see a way how they are store in the svelte ast. We can easily re-add that once we know how it should work


manuel3108:
Noooo, another missing node in `print()`. Making this a draft

PR #840, opened 2025-12-08
chore: drop html deps

### LINKED ISSUE #828: reduce dependency count

manuel3108:
Closes #828 
Relates #839 

moves all html related stuff to using the svelte parser and printer.
removed the `createDiv` function and according test. Why did we even had that, it's totally useless since you can just `createElement('div')`

There is one failing test.
`<!doctype html />` -- needs another pr to svelte.
`<html lang="en" lang="%paraglide.lang%">` -- where the heck is the duplicated lang attribute coming from? I guess it's from the paraglide addon and should always have been there, but it wasn't. Why though? Actually that cant be the case. Is it coming from `svelte` as well? Will need to check

Will need to check tomorrow, it's getting late.
image

### DIFF
--- a/packages/sv/package.json
+++ b/packages/sv/package.json
@@ -21,11 +21,11 @@
 	"bin": "./dist/bin.mjs",
 	"exports": {
 		".": {
-			"types": "./dist/lib/index.d.ts",
+			"types": "./dist/index.d.mts",
 			"default": "./dist/lib/index.mjs"
 		},
 		"./testing": {
-			"types": "./dist/lib/testing.d.ts",
+			"types": "./dist/testing.d.mts",
 			"default": "./dist/lib/testing.mjs"
 		},
 		"./core": {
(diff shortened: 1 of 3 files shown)


packages/sv/package.json
```diff
@@ -45,13 +45,9 @@
 		"decircular": "^1.0.0",
 		"dedent": "^1.7.0",
 		"degit": "^2.8.4",
-		"dom-serializer": "^2.0.0",
-		"domhandler": "^5.0.3",
-		"domutils": "^3.2.2",
 		"empathic": "^1.1.0",
 		"esrap": "^2.2.1",
 		"gitignore-parser": "^0.0.2",
-		"htmlparser2": "^9.1.0",
```
jycouet, line 54: love these `minus` !

manuel3108: ohhh yeahhhh!

manuel3108:
Once https://github.com/sveltejs/svelte/pull/17341 is merged, feel free to update and review here. I don't think there should be anything else to do here.

Issue #841, opened 2025-12-09
paraglide: gitignore for cache doesn't get added anymore


Blackwidow-sudo:
In previous versions when choosing paraglide in `sv create` or adding it with `sv add`, there was a .gitignore file added for the paraglide cache. This doesn't seem to be the case anymore.
```.gitignore
# /project.inlang/.gitignore
cache
```
If thats intentional, then may i ask why it was removed?

jycouet:
I’m not aware of any intentional change there.
I checked the history and couldn’t find anything related to this.
Which previous version behaved the way you expect?

Could it be a change related to `paraglide` version?

Blackwidow-sudo:
> I’m not aware of any intentional change there. I checked the history and couldn’t find anything related to this. Which previous version behaved the way you expect?
> 
> Could it be a change related to `paraglide` version?

Im sorry i cant really point to any versions to when this changed. I just noticed that i had to manually create it when i created a new project. Maybe its related to paraglide version, but idk

jycouet:
I looked at a lot of places... and here is what I found:

- There is: `./src/lib/paraglide` that has a `.gitignore` generated by default (that can be disabled https://inlang.com/m/gerre34r/library-inlang-paraglideJs/compiler-options#emitgitignore)
- the `./project.inlang` folder is "owned" by inlang and not intented to be ignored (if I understand correctly, https://github.com/opral/inlang-paraglide-js/issues/201#issuecomment-3498179708)

So I don't think we should do something that they don't recommend. Let me know if you find a doc about it.

---

Since `2.x`, they generate `./src/lib/paraglide/.gitignore` with
```
# ignore everything because the directory is auto-generated by inlang paraglide-js
# for more info visit https://inlang.com/m/gerre34r/paraglide-js
*
```
So we should probably remove this from our `./.gitignore`
```bash
# Paraglide
src/lib/paraglide
```

Blackwidow-sudo:
> the ./project.inlang folder is "owned" by inlang and not intented to be ignored (if I understand correctly)

yes the directory itself shouldn't get ignored, just the cache files under `project.inlang/cache/*` that get generated when developing.

Here is what i found:
When you manually init paraglide in a sveltekit project with `pnpm dlx @inlang/paraglide-js@latest init`, you still get the ignore file that im missing.

So that must mean that somehow when adding paraglide via `sv`, the `project.inlang/.gitignore` doesn't get generated.

[...]

Issue #826, opened 2025-11-28
paraglideJs build multi language static sites failed


shynome:
hi, I have start a multi language static sites with `sv create`, but always failed, today I got the resolve

the `sv create` template lost the follow hidden links. [+layout.svelte](https://github.com/opral/monorepo/blob/main/inlang/packages/paraglide/paraglide-js/examples/sveltekit/src/routes/%2Blayout.svelte#L10)
```svelte
<div style="display:none">
	{#each locales as locale}
		<a href={localizeHref(page.url.pathname, { locale })}>{locale}</a>
	{/each}
</div>
```

jycouet:
I'm looking around, but don't get your question/issue. 
Could you give a bit more infos?

shynome:
run `npm run build`, don't generate `es` language pages, only got `en` pages.
no `es` pages make me  confuse, the paraglideJs work as expect? I think should add the links to +layout.svelte like paraglideJs example do

jycouet:
The demo creates `messages/en.json` and `messages/es.json` and that's what what we expect.
Also, in the demo, we have 2 buttons to show how to switch locale.

In our demo, I'm not sure we want to do more.
Of course, in your apps, you will probably do more.

Or I'm missing something?

shynome:
1. `pnpx sv create QAQ` with paraglideJs
1. switch `@sveltejs/adapter-auto` to `@sveltejs/adapter-static`
1. `echo 'export const prerender = true' >src/routes/+layout.ts`
1.  `pnpm build`
1.  `tree build`

if no hidden links, pnpm build static site don't generate other language dir, not good for SEO

at least we could add some tip about this to https://svelte.dev/docs/cli/paraglide

```log
build
├── demo
│   └── paraglide.html
├── demo.html
├── es (if no hidden links, pnpm build don't generate other language dir, not good for SEO)
│   ├── demo
│   │   └── paraglide.html
│   └── demo.html
├── es.html
├── index.html
└── robots.txt
```

jycouet:
Thank you for adding all these steps.
Here is the doc for SSG: https://inlang.com/m/gerre34r/library-inlang-paraglideJs/sveltekit#static-site-generation-ssg

Now, what should we do... My feeling in random order:
- Adding all the time some stuff with `display:none` in case "adapter-static" & "prerender" doesn't seem right.
- BTW `adapter-static` is not mandatory?
[...]

PR #844, opened 2025-12-11
fix(paraglide): prerender & ignore cache

jycouet:
Closes: #841
Closes: #826


fr33bits:
I don't think adding `project.inlang/cache/` to the root `.gitignore`  is needed since the Paraglide JS compiler auto-generates (on build or when running the dev server) a `./project.inlang/.gitignore` file that ignores everything in the `./project.inlang/` directory (except `./project.inlang/settings.json`), including `./project.inlang/cache`. Unfortunately the Paraglide JS documentation around this is a bit lacking.

jycouet:
Reading your message shows exactly that the documentation around this is a bit lacking :(
I would love to see simpler stuff there! :)

PR #850, opened 2025-12-14
chore: add sv script from contributing.md

ghostdevv:
The CONTRIBUTING.md mentions a `sv` script, which doesn't exist, so this adds it. I'm not 100% sure whether this ever actually existed, or if there was some behavior in pnpm that made running this in the `packages/sv` work without adding a script called `sv` (from what I can tell it doesn't, so the script is necessary).

https://github.com/sveltejs/cli/blob/88d5c9d244a9ae6ece582b6437be00fac9beee86/CONTRIBUTING.md?plain=1#L46-L50

The only alternative to this I can think would be to change the CONTRIBUTING.md to tell users to run `node dist/bin.mjs` in `packages/sv` themselves.

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,6 +14,7 @@
 		"dev": "tsdown -w",
 		"format": "pnpm --parallel format",
 		"lint": "pnpm --parallel lint && eslint --cache --cache-location node_modules/.eslintcache",
+		"sv": "pnpm --filter sv run sv",
 		"test": "vitest run --silent",
 		"test:ui": "vitest --ui",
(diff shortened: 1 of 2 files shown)


package.json
jycouet, line 36: In this root `package.json` there is `sv` as a `devDependencies`.

This is bringing `sv` to be used with `pnpm sv`. You have to build before I think.

Could you double check?

manuel3108:
I just double-checked the readme. The `pnpm sv` command is only shown after the `pnpm build` command was introduced. So I don't think htere is a need to change anything here. 

The explanation provided by @jycouet is correct.

PR #851, opened 2025-12-14
feat: able to fully setup cloudflare workers/pages

ghostdevv:
Pulling some of the changes discussed in https://github.com/sveltejs/kit/issues/15056

This adds a new option when choosing the Cloudflare Adapter to select workers or pages to deploy to. It diverges a little from the docs since they're outdated.


jycouet:
`FOUND`!
Why docs are not deployed in preview?! Because your fork is `svelte-cli` and not `cli`.
We should update the doc script to allow this as it's totally valid :)

PR #858, opened 2025-12-21
chore: unify `language`

jycouet:
- `typescript` `langTs` `ext`... All expressing the same thing. Let's use `ext` all the time?
- ~~`ensureScript` in parseSvelte directly ?~~

### DIFF
--- a/packages/sv/lib/cli/tests/snapshots/create-with-all-addons/src/routes/demo/+page.svelte
+++ b/packages/sv/lib/cli/tests/snapshots/create-with-all-addons/src/routes/demo/+page.svelte
@@ -1,4 +1,4 @@
-<script>
+<script lang="ts">
 	import { resolve } from '$app/paths';
 </script>
 
--- a/packages/sv/lib/cli/tests/snapshots/create-with-all-addons/src/routes/demo/paraglide/+page.svelte
+++ b/packages/sv/lib/cli/tests/snapshots/create-with-all-addons/src/routes/demo/paraglide/+page.svelte
@@ -1,4 +1,4 @@
-<script>
+<script lang="ts">
 	import { setLocale } from '$lib/paraglide/runtime';
 	import { m } from '$lib/paraglide/messages.js';
 </script>


packages/sv/lib/cli/add/workspace.ts
```diff
@@ -82,7 +82,7 @@ export async function createWorkspace({
 	return {
 		cwd: resolvedCwd,
 		packageManager: packageManager ?? (await detect({ cwd }))?.name ?? getUserAgent() ?? 'npm',
-		typescript: usesTypescript,
+		ext: usesTypescript ? 'ts' : 'js',
```
manuel3108: I'm not sure about the naming here. You probably mean `extension` here, right? But that does not make any sense to me, as I would expect an extension to be on a file by file basis and be any random string `myfile.svelte`. 

Would `language` or similar be the better choice here?

Apart from the naming, I think this is a really valuable change.

packages/sv/lib/addons/_tests/mdsvex/test.ts
```diff
@@ -41,9 +41,8 @@ function addFixture(cwd: string, variant: string) {
 	}
 
 	const src = fs.readFileSync(page, 'utf8');
-	const { ast, generateCode } = parseSvelte(src);
-	const scriptAst = svelte.ensureScript(ast);
-	imports.addDefault(scriptAst, { from: './Demo.svx', as: 'Demo' });
+	const { script, generateCode } = parseSvelte(src, { ensureScript: { ext: 'ts' } });
```
manuel3108: I'm not really sure about this change. After handling my other comment regarding the naming, this might make more sense, but I'm not fully convinced this is more logical than before. Also as an addon author I would wonder why there is `ensureScript` but not ensure `css`, but that's a flaw in the current implementation as well. 

I gave this a try in the [playground](https://svelte.dev/playground/542db5eb953c4590a9c703efb465aaac?version=5.46.1) to always call `ensureScript` but this is flawing the `print`ed result.

But sadly I don't see a cleaner way to this properly. So maybe this is fine as well. Just wanted to share my thoughts.

manuel3108 (changes requested):

packages/sv/lib/cli/add/workspace.ts
jycouet: Yeah, I'm not a huge fan of `ext` but it's really convenient as we use it already in a lot of places. And it's really short (I like that) 

But the discoverability is not great for add-on's authors.

Maybe language all the way ?! (I like that parser & ensure stuff is on one line, not sure it's the case with language... Maybe something to test 👍)

packages/sv/lib/addons/_tests/mdsvex/test.ts
jycouet: Anyway we should probably not enforce script. Maybe we are editing a file without script, and we should not add it. No ?
Now that I write this... I'm questioning myself... Because most of the time we will have a script I guess. Maybe we could switch the logic with a `keepNoScriptIfNoScript`

Piting this here to infuse. (also in community add-on, I did a lot of things around api, not this one... But in general !)

manuel3108: That could work, but I'm not sure we are going to be able to convince typescript that way, as we would need to manipulate the return type of type that we do not directly control (`SvelteAst`)

packages/sv/lib/cli/add/workspace.ts
manuel3108: Yeah, I think we should go with `lang` (common from html attributes) or `language` directly. I think I would prefer `language` all the way.

The one line argument already falls apart in multiple places so let's not argue about that 😆

jycouet: Let's go `language` all the way 🚀

packages/sv/lib/addons/_tests/mdsvex/test.ts
jycouet: Yes, the ~~keepNoScriptIfNoScript~~ is probably not the best idea.

I see your playground, that it's quite clear.
But in our code we need to do a bit of limbo with `ast` & `scriptAst` that is not that cool.
```ts
const ast = parse(input, { modern: true })
const scriptAst = ensureScript(ast, { langTs: true })
imports.addDefault(scriptAst, { from: './Demo.svx', as: 'Demo' });
output = print(ast).code
```

Or maybe we should just do:
```ts
const ast = parse(input, { modern: true })
ensureScript(ast, { langTs: true })
imports.addDefault(ast!, { from: './Demo.svx', as: 'Demo' });        // with ! because we ensure before
output = print(ast).code
```

Or we have the ensure in the parse... I'm 50/50 on this... Let me know what you think @manuel3108 :)

manuel3108: Did you mean
```
imports.addDefault(ast.instance!, { from: './Demo.svx', as: 'Demo' }); 
```
by any chance (`ast!` vs `ast.instance!`)?

Because `ast` contains the svelte ast, which can have two js blocks (`instance` & `module`) so that wouldn't work either.

Here is an alternative for that part: https://github.com/sveltejs/cli/pull/867. It literally makes typescript narrow down the type to remove the `| null` from `ast.instance` as soon as `ensureScript` is called. See that pr's description for other thoughts. I think we should revert those changes in here.

jycouet: You probably saw, but I merged yours, and now this one focus on `language`.

Let's see later if we want to have it inside `parseSvelte`?

manuel3108:
https://github.com/sveltejs/cli/pull/869 should fix the ci failures here

jycouet:
Yeahhh ☺️🚀

PR #860, opened 2025-12-22
chore: touch up the codebase

### LINKED ISSUE #859: Colocate image assets in template
> I don't really like that images are currently stored in $lib/images. This seems like it will just become a huge unorganized catch-all. Images can live in the folder where they're used alongside the +page.svelte that includes them. This should help keep things a bit more organized
https://github.com/sveltejs/cli/issues/496

### Comparison

Image
Image

sacrosanctic:
~~closes #859~~

In addition:
- add comment explaining why `svelte.config.ts` isn't used
- minor typos and fixes

I've grouped relevant changes together for easy `git revert`.


packages/sv/lib/cli/add/workspace.ts
```diff
@@ -24,9 +24,8 @@ export async function createWorkspace({
 	const resolvedCwd = path.resolve(cwd);
 
 	// Will go up and prioritize jsconfig.json as it's first in the array
-	const tjsconfig = find.any([commonFilePaths.jsConfig, commonFilePaths.tsConfig], { cwd });
-	// If the file is not ending with jsconfig.json, then we are using typescript
-	const usesTypescript = !tjsconfig?.endsWith(commonFilePaths.jsConfig);
+	const tsConfigPath = find.any([commonFilePaths.jsConfig, commonFilePaths.tsConfig], { cwd });
+	const typescript = tsConfigPath?.endsWith(commonFilePaths.tsConfig) ?? false;
```
jycouet: As we don't know if it will be `js` or `ts`, _nitpicking_
```suggestion
	const jtsConfigPath = find.any([commonFilePaths.jsConfig, commonFilePaths.tsConfig], { cwd });
	const typescript = jtsConfigPath?.endsWith(commonFilePaths.tsConfig) ?? false;
```

packages/sv/lib/cli/add/workspace.ts
```diff
@@ -32,6 +32,9 @@ export async function createWorkspace({
 	const viteConfig = fs.existsSync(viteConfigPath)
 		? commonFilePaths.viteConfigTS
 		: commonFilePaths.viteConfig;
+
+	// ecosystem doesn't support `svelte.config.ts` https://github.com/sveltejs/cli/issues/816#issuecomment-3568032867
+	// but the user may have opt in to it
```
jycouet:
```suggestion
// `sv` uses `svelte.config.js` by default
// but users may have changed to `svelte.config.ts`
```

jycouet:
Hi @sacrosanctic, thx for looking at this.

I'll try to organize feedback ^^

A/ [colocate assets](https://github.com/sveltejs/cli/pull/860/commits/b4ddf48767785ee56b31aad5bc8e8e626dd277dc)
- After this, there is nothing in `lib` folder and we do'nt show `$lib` alias that is a nice default
- there is a ton of files directly under `/routes` and it's not that pretty. (subjective)
- what about a `(home)` level? in order to:
  - have less stuff just under `/routes`
  - show organization folder
image

- what about keeping `svelte-logo.svg` in `lib/images` and use it in `Header.svelte` and as `favicon` as well?

I'm not 100% sure we want all this, but I would look into this direction if there is something to change.
Maybe @benmccann has more opinion on this?

B/ [consistent names](https://github.com/sveltejs/cli/pull/860/commits/abf6329134911bebbae4e3c08947912710dcd0a4)
Is it way more consistent? 😅
- just before it's `xxxTS`
- this is `tsXxx`
- just after it's `xxx` without the `js`

C/ [easier to read](https://github.com/sveltejs/cli/pull/860/commits/2cb17441a3dbcd71730b45d51955e341383dd7f3)
I remember not linking writing this, haha.

D/ [use commonFilePaths](https://github.com/sveltejs/cli/pull/860/commits/4e8bf5ec43f7d48c20da9782c618d93d415d6729)
Looks good 👍

E/ [comment on no ecosystem support for svelte.config.ts](https://github.com/sveltejs/cli/pull/860/commits/fe67658a8113a3ccd6edab55fe89ea068030fb47)
I'm not a big fan of the 3 mentions & link to github comment. I would put something like my suggestion and remove the 2 other places

F/ [typo](https://github.com/sveltejs/cli/pull/860/commits/483cf0f2f71cdec652cf3b5be9c00f65965efa0e)
Thank you 😁   🫣

sacrosanctic:
I figured there would be a discussion about the assets, so I opened an [issue](https://github.com/sveltejs/cli/issues/859#issuecomment-3680974439) for it. I guess it doesn't matter where the discussion happens.

> what about a (home) level? 

I've personally used `/FILES` before, it would be nice to come up with an agreeable convention.

packages/sv/lib/cli/add/workspace.ts
sacrosanctic: `jts` reads like an abomination :(

jycouet: lol, I love it! :)
Maybe love is a strong word... But I like it a lot ^^

benmccann:
Maybe `_assets` if we want to put just the images and `.css` there or perhaps `_includes` for everything including those and the components. I asked in `#ambassadors` to see if others have ideas

packages/sv/lib/cli/add/workspace.ts
sacrosanctic: What about like this, or is it too cheeky? https://github.com/sveltejs/cli/pull/860/changes/1cc1d50d6a6c2aa335d20384c6c259b0fb87ec9f

sacrosanctic:
`_include` sounds pretty good as a catch-all folder name. But I think we also still have a reference to `$lib/...` to showcase it. what about a `$lib/util.js#dateFormat` function?

jycouet:
`T`oday

`H` move page things under (home)

`C` co-locate assets & remove favicon.svg in favor of svelte-logo.svg

`i` _includes folder to group things

image

Like this, `$lib` is still here and we don't have to "make it up" ^^

packages/sv/lib/cli/add/workspace.ts
jycouet: It's not `@deprecated`! Maybe in 2 years it will be the default! _(I don't know ^^)_
It's just not the default today.

I'm not sure why you want to point it out absolutely?
_It feels that you don't like this default?_

benmccann:
`_includes` got the most votes among ambassadors and it sounds like you like it as well, so I say we go with that

benmccann:
Actually, it sounds like `_assets` would be preferred to `_includes`. It wasn't part of the vote, but enough people are speaking up that it seems like the better option

benmccann:
I'm really going back and forth on this, but I just changed my vote to keep things as they are today without any folder for additional organization, which now puts that option in the lead due to the low number of voters. The reason I'm changing my mind is that I noticed the screenshot is not how things appear for me. In the screenshot, files are appearing before folders. I think that makes things look messier because you have to scan through all the files to see the list of folders. That originally caused me to vote for having more organization. However, that's not how things appear for me. And it's not how things appear by default as far as I can tell

jycouet:
I'm also in favor of not changing organization for demo atm.

The PR is still good for the other things you suggest @sacrosanctic 
Let's clean it and I'll merge 👍

packages/sv/lib/cli/add/workspace.ts
sacrosanctic: Ya, I know it's not deprecated, I just want to point how that it shouldn't be used.

sacrosanctic:
Ready

sacrosanctic:
> In the screenshot, files are appearing before folders. I think that makes things look messier because you have to scan through all the files to see the list of folders...However, that's not how things appear for me. And it's not how things appear by default as far as I can tell

Correct, it is not the default but a [vscode configuration](https://www.sveltepatterns.dev/getting-started-with-vscode#put-files-first) to put files first. This change is recommended; without it, it becomes difficult to match reserved files with their routes, as you now have to scan through all subroutes to find the relevant files. The opposite of the current situation.

left is folder first
right is files first
imageimage

PR #864, opened 2025-12-28
feat(cli): add promptless args to `README.md`

### LINKED ISSUE #863: put the reproducible cli command in the readme.md
Replace or add to the contents of the readme with the args version from the cli.

https://github.com/sveltejs/cli/blob/fc8fc8ca57d708dee275dadf3de0a9c9e1e93db5/packages/sv/lib/create/shared/README.md?plain=1#L5-L15

 ```diff
 # create a new project in the current directory 
 npx sv create 
  
 # create a new project in my-app 
 npx sv create my-app 

 # Some comment
+ pnpm dlx sv create --template minimal --types ts --add storybook --install pnpm ./
``` 

Image

### PR DESCRIPTION
**Pull request by sacrosanctic**
closes #863

I cannot reuse the process built into `createKit` to update `README.md` as the order of operation is `createKit` ->`runAddonsApply` -> `buildArgs`, at that point it is too late to use `createKit`. Thus, a manual approach was used to update the `README.md`.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -94,13 +94,12 @@ pnpm preview
 
 ### Update snapshots
 
-Test snapshots are located in `packages/sv/lib/cli/tests/snapshots`. You can update them manually or by using the interactive UI via `pnpm test:ui`. In either case, run `pnpm build` the repo before testing.
+Some snapshots are testing the output of `sv` directly from the generated binary. They are located in `packages/sv/lib/cli/tests/snapshots`. Make sure to generate a new binary before updating these snapshots.
 
-Via interactive UI:
+In one command:
 
 ```sh
-pnpm build
-pnpm test:ui --project cli # core / addons / create / migrate
+pnpm build && pnpm test:ui --project cli
 # Press `u` when prompted to update snapshots.
 ```
 

### CODE REVIEW

**Comment by manuel3108** [comment]
Looks like the tests are failing. You can easily reproduce this by running `pnpm test --project cli` locally. To update the snapshots automatically, you can run `pnpm vitest --project cli` (note `vitest` instead of `test`), which starts the tests in debug mode. After the tests have initially run, you can press `h` or `u` to update the snapshots. 

Apart from that I will leave this open for @jycouet so that he can check if and how this will conflict with the changes from #684

**Comment by sacrosanctic** [comment]
There are still failing tests, but I'm going to leave it as is until @jycouet gives it a pass, as this may be the wrong approach.

File: packages/sv/lib/cli/create.ts [hunk 1]
```diff
@@ -119,7 +122,10 @@ export const create = new Command('create')
 				...initialSteps,
 				`  ${i++}: ${highlight(pmRunCmd)}`,
 				'',
-				`To close the dev server, hit ${highlight('Ctrl-C')}`
+				`To close the dev server, hit ${highlight('Ctrl-C')}`,
+				'',
+				`To recreate this project without prompts, run:`,
+				`  ${pc.cyan(prompt)}`
```
jycouet: When we started this feature we had multiple ideas around where to put this message. But in the "What's next?" section, it's taking too much space!

Image

That's why it was gray, and at a great position.

It's overloading the screen for 100% of users while only a very small % will reuse the cli in the same context.

CONTRIBUTING.md
```diff
@@ -90,6 +92,16 @@ pnpm build
 pnpm preview
 ```
 
+### Update snapshots
+
+Tests use snapshots to verify expected output against actual results. When you make changes that affect test output, update the snapshots. To update snapshots `vitest`'s interactive UI', run:
+
+```sh
+pnpm vitest --project cli # core / addons / create / migrate
```
jycouet: This is working as well:
```sh
pnpm test:ui --project cli
```

For these snapshots update, I would sugges to `build` also before

```ts
pnpm build && pnpm test:ui --project cli
```

packages/sv/lib/cli/add/index.ts
```diff
@@ -658,7 +658,10 @@ export async function runAddonsApply({
 		}
 	}
 
-	if (fromCommand === 'add') common.logArgs(packageManager, 'add', argsFormattedAddons);
+	if (fromCommand === 'add') {
+		const prompt = common.buildArgs(packageManager, 'add', argsFormattedAddons);
+		p.log.message(`To rerun this command without prompts, run:\n${pc.dim(prompt)}`);
```
jycouet: Look at the other comment for location.
Having a small text on top, why not... But very small imo. (I did a few when the feature was crafted, but didn't find something nice)

CONTRIBUTING.md
sacrosanctic: You mean use `test:ui` over `vitest`?

packages/sv/lib/cli/add/index.ts
sacrosanctic: What about:
- To re-run without prompts:
- Reproducible CLI command:
- Non-interactive:
- Automate with:
- Skip prompts:
- Run again:

sacrosanctic: image

CONTRIBUTING.md
jycouet: Yes 👍
And as these tests are using the bin, we need to build before, to make sure we are on latest

packages/sv/lib/cli/add/index.ts
jycouet: - Re-run without questions
- Re-run without prompts
- Run again, no questions asked
- Run again, no prompts

I this that the text should by gray, maybe try with log info as well to have a dot in timeline ? (I think I didn't like it much, but with a title maybe... Maybe not 😅

jycouet:
Right now, it's with the blue dot in the timeline, I tried other style to see:
image

Yeah, the blu is probably the best 👍

I'll update a few thing in the code s00n before merging

packages/sv/lib/cli/tests/cli.ts
```diff
@@ -45,7 +45,10 @@ describe('cli', () => {
 		async (testCase) => {
 			const { projectName, args } = testCase;
 			const svBinPath = path.resolve(monoRepoPath, 'packages', 'sv', 'dist', 'bin.mjs');
-			const testOutputPath = path.resolve(monoRepoPath, '.test-output', 'cli', projectName);
+			const testOutputPath = path.relative(
```
sacrosanctic, line 48: absolute paths aren't tested

packages/sv/lib/cli/utils/common.ts
```diff
@@ -137,24 +139,57 @@ export function parseAddonOptions(optionFlags: string | undefined): string[] | u
...
 	const res = resolveCommand(agent ?? 'npm', 'execute', [...allArgs, ...lastArgs])!;
-	p.log.message(pc.dim([res.command, ...res.args].join(' ')));
+	const message = [res.command, ...res.args].join(' ');
+
+	p.log.info(pc.dim(`Re-run without prompts:\n${message}`));
+
+	return message;
+}
+
+export function updateReadme(projectPath: string, command: string) {
```
sacrosanctic, line 162: This fn is brittle. We will change this down the line?

packages/sv/lib/cli/create.ts
```diff
@@ -307,7 +307,10 @@ async function createProject(cwd: ProjectPath, options: Options) {
 
 	if (argsFormattedAddons.length > 0) argsFormatted.push('--add', ...argsFormattedAddons);
 
-	common.logArgs(packageManager, 'create', argsFormatted, [directory]);
+	const prompt = common.buildAndLogArgs(packageManager, 'create', argsFormatted, [
+		common.normalizePosix(directory)
```
sacrosanctic: Should we be normalizing further up?

https://github.com/sacrosanctic/cli/blob/26f3de3032245ffd74b4d3e3a2fe967cc6f3c017/packages/sv/lib/cli/create.ts#L152


```diff
if (cwd) {
-	return Promise.resolve(cwd);
+ return Promise.resolve(common.normalizePosix(cwd))
}
```

packages/sv/lib/cli/tests/cli.ts
jycouet: I think it's not about testing relative or absolute path.
And this part could probably be reverted.

packages/sv/lib/cli/utils/common.ts
jycouet: Not sure. What do you have in mind? :)

packages/sv/lib/cli/create.ts
jycouet: I didn't wanted to go to more issues, but yes it might be better 👍 

Should work on all `OS`, but we have to triple check ^^

packages/sv/lib/cli/tests/cli.ts
sacrosanctic: Reverting it results in an absolute path on the GitHub test runner. Resulting in non-universal snapshots.

packages/sv/lib/cli/utils/common.ts
sacrosanctic: Nothing ATM. This was the quick fix solution I came up with. I just want to make a note of it.

packages/sv/lib/cli/create.ts
sacrosanctic: The current tests should be sufficient.

PR #866, opened 2025-12-29
fix: Cloudflare adapter integrations

teemingc:
This PR fixes a few issues as a follow up to https://github.com/sveltejs/cli/pull/851

1. The referenced types file needs to be a relative path. Otherwise, TS thinks it's a package name and you get an error in the tsconfig.json file such as:
```
Cannot find type definition file for 'worker-configuration.d.ts'.
  The file is in the program because:
    Entry point of type library 'worker-configuration.d.ts' specified in compilerOptions
```
2. The preview port needs to be 4173 so that it works with the Playwright command which targets the same port. Otherwise, wrangler dev defaults to port 8080 and the Playwright test:e2e script fails

### DIFF
--- a/.changeset/honest-tables-compete.md
+++ b/.changeset/honest-tables-compete.md
@@ -2,4 +2,4 @@
 "sv": patch
 ---
 
-fix: use a relative path for the Cloudflare worker-configuration type in the tsconfig.json file
+fix(cloudflare): use a relative path for worker-configuration type in the tsconfig.json file
--- a/.changeset/smooth-walls-remain.md
+++ b/.changeset/smooth-walls-remain.md
@@ -2,4 +2,4 @@
 "sv": patch
 ---
 
-fix: set Cloudflare local preview port to 4173 so that it works with the Playwright test command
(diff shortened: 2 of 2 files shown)


.changeset/smooth-walls-remain.md
```diff
@@ -0,0 +1,5 @@
+---
+"sv": patch
+---
+
+fix: set Cloudflare local preview port to 4173 so that it works with the Playwright test command
```
jycouet:
```suggestion
fix(cloudflare): local preview is now using port `4173` so that it works with the Playwright test command
```

.changeset/honest-tables-compete.md
```diff
@@ -0,0 +1,5 @@
+---
+"sv": patch
+---
+
+fix: use a relative path for the Cloudflare worker-configuration type in the tsconfig.json file
```
jycouet:
```suggestion
fix(cloudflare): use a relative path for worker-configuration type in the tsconfig.json file
```

manuel3108 (approved):
Thanks!

Issue #846, opened 2025-12-12
ESLint & VS Code configuration


jycouet:
> `dbaeumer.vscode-eslint@3.0.20` has just been [released](https://github.com/microsoft/vscode-eslint/releases/tag/release/3.0.20).
> 
> 1. Update the VS Code ESLint extension
> 2. Delete this line from the VS Code configuration:
> 
> ```json
> "eslint.validate": ["javascript", "javascriptreact", "svelte"]
> ``` 

 _Originally posted by @hyunbinseo in [#626](https://github.com/sveltejs/cli/issues/626#issuecomment-3625686670)_

---

_I would like to track it here._

jycouet:
For me it's not published `3.0.20`
<img width="185" height="68" alt="Image" src="https://github.com/user-attachments/assets/a37c6a51-097b-4c5e-9bee-ac0f8a5a0f3b" />

And in the release note I was not able to find this information of remeving `eslint.validate` 👀

May you help me @hyunbinseo :)

hyunbinseo:
@jycouet reference this PR for better context:

- https://github.com/sveltejs/eslint-plugin-svelte/pull/1396

Does VS Code Marketplace support gradual deployment? I am seeing `3.0.20` as the latest stable.

- [GitHub Release](https://github.com/microsoft/vscode-eslint/releases/tag/release%2F3.0.20)
- [Marketplace Changelog](https://marketplace.visualstudio.com/items/dbaeumer.vscode-eslint/changelog)

<img width="1437" height="944" alt="Image" src="https://github.com/user-attachments/assets/a5bda40f-5a26-4c5e-9139-0d442b672184" />

<img width="1340" height="632" alt="Image" src="https://github.com/user-attachments/assets/409c42f3-7a39-455f-bae2-f2fb07776cc4" />

jycouet:
I'm using Cursor and I have this:
<img width="1963" height="808" alt="Image" src="https://github.com/user-attachments/assets/2a1790b9-38be-4a5e-8f21-d40c19f8ab3e" />

But it's taking settings of vscode! :o

Hummm... Strange!

manuel3108:
<img width="1141" height="345" alt="Image" src="https://github.com/user-attachments/assets/28a9c376-e3c7-4a66-9f02-9abb3b05373d" />
Same for me, it did update.

jycouet:
It's up to date for me too now! 👍

hyunbinseo:
[...]

PR #871, opened 2026-01-08
feat(eslint): remove vscode `eslint.validate`

KiRura:
https://svelte.jp/docs/cli/eslint states:

> What you get
> updated .vscode/settings.json

However, when settings.json already exists and `eslint.validate` is undefined, the settings are not updated, preventing the ESLint extension from linting `.svelte` files.


packages/sv/lib/addons/eslint/index.ts
```diff
@@ -46,10 +46,13 @@ export default defineAddon({
 			if (!content) return content;
 
 			const { data, generateCode } = parseJson(content);
-			const validate: string[] | undefined = data['eslint.validate'];
-			if (validate && !validate.includes('svelte')) {
+			const validate: string[] = data['eslint.validate'] ?? ['typescript', 'javascript', 'svelte'];
```
jycouet: Do we need `typescript` & `javascript` ? You have a reference to this?

We should add it to the list after maybe?
And maybe depending on the project type also?

packages/sv/lib/addons/eslint/index.ts
```diff
@@ -46,10 +46,13 @@ export default defineAddon({
...
+
+			if (!validate.includes('svelte')) {
 				validate.push('svelte');
 			}
+
+			data['eslint.validate'] = validate;
```
jycouet: I think that I want to create a util for this as we have it in multiple places 👍

jycouet (commented):
Thank you for looking at this.
Could you add a changelog? (`pnpm changeset` and looking at others in `CHANGELOG.md`, find something that fit :) )

jycouet:
Actually, according to https://github.com/sveltejs/cli/issues/846 we don't need this `eslint.validate` anymore!

jycouet:
Sorry, I jeopardize your PR @KiRura ! Do you still agree?

Like this it would:
- Close #846 
- Keep your contribution
- Bring some nice refactoring (IMO)

@hyunbinseo could you have a look as well ?
@manuel3108 could you check as well api / naming ? 

Thx :)

KiRura:
Yes, that's fine with me!
I wasn't aware that files.associations could work instead of eslint.validate. I apologize for my lack of knowledge.

Issue #872, opened 2026-01-09
Skip generating wrangler types when building for deployment


teemingc:
The current Cloudflare adapter template adds `wrangler types` to the prepare npm lifecycle script but types are not necessary for deployments. Is it worth finding a way to skip this step in CI and making that change?

cc: @ghostdevv

jycouet:
Today we have: 
```
"prepare": "pnpm run types && svelte-kit sync || echo ''",
```

1/ We should not use `pnpm` but `npm`.

2/ To answer your question, we could move the types generation in `dev` AND `build` ?

teemingc:
I think we only need it for type checking so it could be moved to the check script

EDIT: thinking about it more, we really want to avoid overrunning this. The types would only change when certain fields are changed in the wrangler config which is not very often

jycouet:
I find it strange to generate something in a script called `check`.

We can also just add it to next steps and remove it from `prepare`?
Something like: 

<img width="423" height="297" alt="Image" src="https://github.com/user-attachments/assets/c216fb89-82f2-4fd1-aec7-d19d06a6084b" />

+ It's teaching people that they need to run it again... 

<img width="665" height="63" alt="Image" src="https://github.com/user-attachments/assets/3160c0f2-7517-430a-8881-3fa4211be053" />

ghostdevv:
Only just seeing this unfortunately - moving it out of prepare is tricky because it's gitignored, which then means that when you setup a project you have to both know and remember to run the generate script. I think this is a place we're different from what `create-cloudflare` does, as they do gitignore it - but personally I'm not a big fan of committing generated files. It does feel like a rock and hard place situation though, since it's not a super fast command (1-3s on my system). They did add a `--check` flag recently so you know if it's stale, but it takes the same amount of time to run as just outputting the types 🫠

teemingc:
Do you know if there’s a solution that could be better than the CLI reminding the user to run it manually?

There’s also an issue where you will get type errors after generating the types if you have a wrangler config referencing the generated worker script because it’s not typed

jycouet:
Maybe a bit too fast !
If we go back to the original issue... Is it a problem to have it generating types we deploying ?!
Maybe going back to the original way by default is not that bad ?

ghostdevv:
> There’s also an issue where you will get type errors after generating the types if you have a wrangler config referencing the generated worker script because it’s not typed

Do you have an example of what you mean?

> Is it a problem to have it generating types we deploying ?!

[...]

PR #877, opened 2026-01-19
fix(cloudflare): don't generate types on install

jycouet:
closes: #872


manuel3108 (approved):
LGTM, if @teemingc agrees

PR #881, opened 2026-01-20
chore(internal): flatten addons (+doc link update)

jycouet:


manuel3108 (approved):
As discussed, this totally makes sense. Since this is an internal change, no changeset is required.

Issue #875, opened 2026-01-18
replace `picocolor` with `styleText`


manuel3108:

hyunbinseo:
I would love this to land, but isn't it [added in 20.12](https://nodejs.org/api/util.html#utilstyletextformat-text-options)? `sv` probably needs an `engines` field.

jycouet:
I think that we are already 16+, but we don't enforce it. Maybe we can be 20+ and we don't enforce it ?

I remember that we had issues with engines and package managers. 
-> But we can revisit and check again 👍

PR #882, opened 2026-01-20
chore(deps): drop `picocolors` for `styleText` of `node:util`

jycouet:
Closes: #875

### DIFF
--- a/packages/sv/lib/cli/add/index.ts
+++ b/packages/sv/lib/cli/add/index.ts
@@ -191,7 +191,6 @@ export const add = new Command('add')
 
 			// Addon Options
 			const addonList = addonOptions.map((option) => {
-				// const description = `${pc.dim(`(preset: ${option.preset})`)}\n${option.choices}`;
 				const description = option.choices;
 				return callFormatItem(
 					helper.styleArgumentTerm(option.id),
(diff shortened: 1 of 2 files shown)


jycouet:
There is now ONE central place for colors (with semantic)
I tuned some... and it looks like

image

I'm not bound to these! you can play around

packages/sv/lib/cli/utils/common.ts
```diff
@@ -56,21 +55,21 @@ export const helpConfig: HelpConfiguration = {
 
 		return option.flags;
 	},
-	styleTitle: (str) => pc.underline(str),
-	styleCommandText: (str) => pc.red(str),
-	styleDescriptionText: (str) => pc.gray(str),
-	styleOptionText: (str) => pc.white(str),
-	styleArgumentText: (str) => pc.white(str),
-	styleSubcommandText: (str) => pc.red(str)
+	// styleTitle: (str) => color.underline(str),
```
manuel3108: should we remove this? Same goes for 3 lines below.

manuel3108 (changes requested):
Maybe I'm missing something, but in #875 there were talks about a node version check. I'm not seeing this here. Is there a reasoning?

Edit: There is a conflict as well.

jycouet:
Tentative of a small update
image

manuel3108:
Edit: Not merging, not sure if you don't want to add a changeset. I don't think it's necessary here, but maybe because of the potentially breaking behaviour for some users?

Issue #884, opened 2026-01-21
Duplicate `eslint .` inside the `lint` script


hyunbinseo:
```script
pnpm dlx sv create --template minimal --types jsdoc --add eslint prettier tailwindcss="plugins:forms" --install pnpm javascript
pnpm dlx sv create --template minimal --types ts --add eslint prettier tailwindcss="plugins:forms" --install pnpm typescript
```

With the above script, the following script is created:

```json
"lint": "eslint . && prettier --check . && eslint .",
```

Does the `--add` order matter> I've sorted them in ABC.

What I am used to:

```json
"lint": "prettier --check . && eslint .",
```

jycouet:
It's to be extra sure 🤪

It's gonna be a fun one ^^, I'll look later. Thx

PR #889, opened 2026-01-21
fix(prettier): in `lint` step, prettier will always be first

jycouet:
Closes: #884


packages/sv/lib/core/tooling/json.ts
```diff
@@ -19,11 +19,23 @@ export function arrayUpsert(
 	data[key] = array;
 }
 
-export function packageScriptsUpsert(data: any, key: string, value: any): void {
+export function packageScriptsUpsert(
+	data: any,
+	key: string,
+	value: any,
```
manuel3108:
```suggestion
	value: string,
```
Seems like this was the case previously, but I don't see why honestly.

PR #890, opened 2026-01-21
fix: `app-state` migration finds more files

jycouet:
Closes: #368

### DIFF
--- a/packages/migrate/migrations/app-state/migrate.js
+++ b/packages/migrate/migrations/app-state/migrate.js
@@ -53,10 +53,17 @@ export function transform_svelte_code(code) {
 			const before = modified.slice(0, match.index);
 			const after = modified.slice(match.index + alias.length);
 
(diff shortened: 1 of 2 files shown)


packages/migrate/migrations/app-state/migrate.js
```diff
@@ -39,43 +39,62 @@ export function transform_svelte_code(code) {
...
-		let count_removed = 0;
 
 		while ((match = regex.exec(modified)) !== null) {
 			const before = modified.slice(0, match.index);
 			const after = modified.slice(match.index + alias.length);
 
+			// Skip if inside a comment (check if // appears on the same line before the match)
+			const line_start = before.lastIndexOf('\n') + 1;
+			const line_before_match = before.slice(line_start);
+			if (line_before_match.includes('//')) {
```
manuel3108, line 59: What about for multi-line comments?

Issue #904, opened 2026-02-05
The binary images are corrupted when creating the demo app


MarkMizuguchi:
Create a new project using:
`npx sv create --template demo --types ts --install npm my-app
`

Attempt to open the file `my-app/src/lib/images/svelte-welcome.webp` and notice that the image file is corrupted. The file does not match the size of the file at <https://github.com/sveltejs/kit-template-default/blob/main/src/lib/images/svelte-welcome.webp>.

This issue began with version 0.11.3.

manuel3108:
Can confirm, no clue why though at this point. Files are broken in vscode as well. The file-size doubles after running sv. The images are not shown in the demo app.

<img width="1711" height="323" alt="Image" src="https://github.com/user-attachments/assets/38dfc7bd-9e5a-453d-8e40-5882c2fc8ad4" />

<img width="1717" height="1054" alt="Image" src="https://github.com/user-attachments/assets/4929c62a-00ae-4211-9093-f068fe7ab613" />

PR #905, opened 2026-02-05
fix(cli): raw copy of binary files

jycouet:
closes: #904

### DIFF
--- a/packages/sv/lib/create/index.ts
+++ b/packages/sv/lib/create/index.ts
@@ -1,6 +1,6 @@
 import fs from 'node:fs';
 import path from 'node:path';
-import { mkdirp, copy, dist, getSharedFiles, replace } from './utils.ts';
+import { mkdirp, copy, dist, getSharedFiles, replace, kv } from './utils.ts';
 import { commonFilePaths } from '../cli/add/utils.ts';
 import { sanitizeName } from '../coreInternal.ts';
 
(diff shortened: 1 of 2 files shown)


Conduitry (commented):
Hm. What do you think about whitelisting certain file extensions for having the replacements happen in them, rather than marking some as being copied as-is? The list of file types that we want to be replacing strings in is probably more finite than the list of all the types of files that we don't want to be replacing them in.

jycouet:
> Hm. What do you think about whitelisting certain file extensions for having the replacements happen in them, rather than marking some as being copied as-is? The list of file types that we want to be replacing strings in is probably more finite than the list of all the types of files that we don't want to be replacing them in.

I hesitated with this other side too yes. Since it was more `.css, .html, .js, .json, .map, .mjs, .mts, .txt,  DOT-gitignore, DOT-npmrc`, I was like, maybe it's better the opposite?

Maybe something like this is better as it has no extension list to keep up?
```ts
const buffer = fs.readFileSync(from);
const keys = Object.keys(kv);
// Check if file contains any replacement keys (only text files will)
const needsReplacement = keys.length > 0 && keys.some((key) => buffer.includes(key));
if (needsReplacement) {
	fs.writeFileSync(to, replace(buffer.toString('utf-8'), kv));
} else {
	fs.writeFileSync(to, buffer);
}
```

PR #908, opened 2026-02-06
fix(drizzle): default .env with turso is now using `file:local.db`

jycouet:


packages/sv/lib/addons/drizzle.ts
```diff
@@ -460,12 +460,21 @@ function generateEnvFileContent(
...
-			'Replace with your DB credentials!',
-			{ text: 'A local DB can also be used in dev as well', mode: 'append' },
-			{ text: `${DB_URL_KEY}="file:local.db"`, mode: 'append' }
-		);
+		if (isExample) {
+			value = '"libsql://db-name-user.turso.io"';
+			comment.push(
+				'Replace with your DB credentials!',
+				{ text: 'A local DB can also be used in dev as well', mode: 'append' },
+				{ text: `${DB_URL_KEY}="file:local.db"`, mode: 'append' }
```
manuel3108, line 468: any reason we have the json like stuff here, but not in the else? Let me know if im missing something

jycouet: because by default comments are on top. But here we want them under (append)
So it's `string | { test: string, mode: 'append'}`

PR #909, opened 2026-02-06
chore: use `parseCss` from `svelte/compiler`

### LINKED ISSUE #903: use `svelte` css parser
From the newsletter:
> Svelte’s CSS parser is now exported from svelte/compiler as parseCss. It’s a partial but lightweight CSS AST parser that is currently being used in SvelteKit (svelte@5.48.0, [#17496](https://substack.com/redirect/0dc1655a-fd83-4dda-b371-48221665b938?j=eyJ1IjoiMnJ5cGl2In0.ACRSg6WQZPFdt_6Ap_v2N1J093RYZVHJTklfGho9FP0))

Can we maybe use this here, instead of that workaround?
https://github.com/sveltejs/cli/blob/c056bb0a063f6d709085a800276ecb04740abb40/packages/sv/lib/core/tooling/index.ts#L65-L68

I haven't thought this through, just a random idea.

Edit: That would probably lead to quite a lot of other changes, but maybe it could be worth it?

manuel3108:
Closes #903 
Draft, because i hate that we need to copy this omit everywhere. See thread in discord

### DIFF
--- a/packages/sv/lib/core/tooling/css/index.ts
+++ b/packages/sv/lib/core/tooling/css/index.ts
@@ -1,7 +1,7 @@
 import type { SvelteAst } from '../index.ts';
 
 export function addRule(
-	node: SvelteAst.CSS.StyleSheetRules,
+	node: SvelteAst.CSS.StyleSheetBase,
 	options: { selector: string }
 ): SvelteAst.CSS.Rule {
 	// we do not check for existing rules here, as the selector AST from svelte is really complex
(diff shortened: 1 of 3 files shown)


manuel3108:
Uses https://github.com/sveltejs/svelte/pull/17654 for now. Should update to released version here once published. Way better now.

manuel3108:
This is good to go now

packages/migrate/package.json
```diff
@@ -43,7 +43,7 @@
 		"@types/node": "^20.14.8",
 		"@types/prompts": "^2.4.9",
 		"@types/semver": "^7.7.1",
-		"svelte": "^5.45.10"
+		"svelte": "^5.48.0"
```
jycouet, line 46: in `sv` we use `^5.51.0` now, should we sync these versions?
Maybe moving versions in `pnpm-workspace.yaml` ?

Something for later?

PR #917, opened 2026-02-08
feat(cli): prepare @sveltejs/sv-utils for community add-on

jycouet:


packages/sv-utils/src/tests/svelte/svelte/index.ts
```diff
@@ -2,7 +2,7 @@ import fs from 'node:fs';
 import { join, resolve } from 'node:path';
 import { fileURLToPath } from 'node:url';
 import { describe, expect, test } from 'vitest';
-import { parseSvelte, serializeSvelte } from '../../tooling/index.ts';
+import { parseSvelte, serializeSvelte } from '../../../../../sv-utils/src/tooling/index.ts';
```
manuel3108: shouldn't this be `@sveltejs/sv-utils` Or did i miss something?

jycouet: WoW, good catch! I updated the import path as we are already in `sv-utils` 👍

I looked for `/sv-utils/` and it was the only place

manuel3108 (approved):
This works as expected, feel free to merge at any time 🥳

PR #918, opened 2026-02-08
feat(cli): add new command `docs` to help Humans & Agents to navigate

jycouet:


manuel3108:
I probably missed something, but is that really something we want? Why not add the other two commands from the [mcp cli](https://svelte.dev/docs/mcp/overview)?

Otherwise the code looks good, but I'm not sure I understand the intention behind this.

jycouet:
The idea was to help AI looking in docs. Yes, it's just a wrapper of https://svelte.dev/docs/mcp/tools
Maybe we could document cli usage there?

Or do it like this PR?

I'm not sure what's the best. @paoloricciuti any hint?

paoloricciuti:
I definitely need to add docs to the mcp section for the cli...the only advantage here would be a nicer command but I don't think LLMs care too much about that 😄

jycouet:
Let's not optimize for number of character 🤪

I'm fully with you. *IF* i find the time between x, y, z I will PR the doc ^^ _(But don't wait for me)_

_I'll close here_

Issue #914, opened 2026-02-07
e2e folder in SvelteKit template isn't type-checked


teemingc:
Currently, the Svelte template adds an `e2e` directory for Playwright tests. However, this directory isn't type-checked (see https://github.com/sveltejs/kit/pull/15254). We have 3 options:

1. Rename "e2e" to "tests" which is covered by the generated tsconfig.json
2. Change the generated tsconfig.json to recognise the "e2e" directory
3. Include it via the SvelteKit typescript config option

jycouet:
Yes we should fix this!

Some random points:
- `playwright.config.ts` is today looking at this folder `e2e`
- e2e & vitest will do:
<img width="236" height="246" alt="Image" src="https://github.com/user-attachments/assets/59d2c09f-7824-45d5-9dcf-db7fce860cf8" />

---

Me, I'm always changing stuff to have e2e tests close to the route, so I would move `e2e/demo.test.ts` to `src/routes/page.svelte.e2e.ts` and update the conf to look for tests ending with `e2e.ts`
Bonus, no confusion between unit & e2e files, 

||files|
|-|-|
|unit|.{test,spec}.{js,ts}|
|e2e|.e2e.{js,ts}|

And it would solve typecheck.

What do you think?

wyMinLwin:
I tried to solved by simply renaming it in PR #919

PR #919, opened 2026-02-09
fix: move e2e tests beside the page

wyMinLwin:
Closes: #914

I just simply renamed e2e/ directory to tests/ and updated playwright.config.ts testDir accordingly

### DIFF
--- a/.changeset/ninety-taxis-beg.md
+++ b/.changeset/ninety-taxis-beg.md
@@ -2,4 +2,4 @@
 'sv': patch
 ---
 
-feat(playwright): looking now at `**/*.e2e.{ts,js}` files by default
+feat(playwright): use `**/*.e2e.{ts,js}` files by default


jycouet:
I think that like this there is a confusion between files for Playwright and the one for Vitest.

wyMinLwin:
so "Include it via the SvelteKit typescript config option" would be ok ?

```
 kit: {
    typescript: {
      config: (config) => {
        config.include = [...config.include, '../e2e/**/*.ts'];
      }
    }
  }
 ```

**Comment by wyMinLwin** [comment]
Hope you could review new approach

**Comment by jycouet** [comment]
I'll look Friday

**Comment by jycouet** [comment]
@wyMinLwin , what do you think about this?

**Comment by wyMinLwin** [comment]
@jycouet really great! I think it's cleaner and better

**Comment by jycouet** [comment]
FYI: we are thinking what is the best... And maybe not like this right now!
_We still have the original issue in mind... Let's see how to fix it 👍_

**Comment by sacrosanctic** [comment]
<table>
<tr>
 <td>
 <td> files
<tr>
 <td> unit 
 <td> .{test}.{js,ts}
<tr>
 <td> e2e
 <td> .e2e.test.{js,ts}
</table>

what about naming them all `*.test.js`

**Comment by jycouet** [comment]
> files
> unit	.{test}.{js,ts}
> e2e	.e2e.test.{js,ts}
> what about naming them all `*.test.js`

Like this, `vitest` will pick e2e tests and fail.

File: .changeset/ninety-taxis-beg.md [hunk 1]
```diff
@@ -0,0 +1,5 @@
+---
+'sv': patch
+---
+
+feat(playwright): looking now at `**/*.e2e.{ts,js}` files by default
```
manuel3108:
```suggestion
feat(playwright): use `**/*.e2e.{ts,js}` files by default
```

PR #933, opened 2026-02-19
chore: improve windows testing speed

jycouet:
On `better-auth`, we are always doing a retry....!
First try, `pnpm i`, `pnpm build`, ... timeout!
Second try: `pnpm i` very fast... And we manage to finish

So if we give more time to finish the first time, it's will be maybe better?!
Result: YES! :)

---

I'm thinking... Should `packages/sv/src/addons/tests/all-addons/test.ts` run at all?
Now that we have `packages/sv/src/cli/tests/cli.ts` tests?
-> Would be a net 20sec win.


manuel3108:
> I'm thinking... Should packages/sv/src/addons/tests/all-addons/test.ts run at all?

That sounds really reasonable. Removed!

PR #949, opened 2026-02-27
fix: Change postgres.js dependency to devDependency

sacrosanctic:
The other SQL packages are also added as `dependencies`, they could be candidates to be moved as well.


jycouet:
All sv.devDependency except `better-sqlite3` (I remember soemthing there).
What do you think?

sacrosanctic:
it makes sense to me

PR #950, opened 2026-02-27
fix: install `@better-auth/cli` as a dev dependency

### LINKED ISSUE #937: [Better-Auth] Use the selected package manager for the `auth:schema` script
https://github.com/sveltejs/cli/blob/02e855e40a262826584fc0bedc0497f47f431329/packages/sv/src/addons/better-auth.ts#L137

This line enforces `npx`, but if I selected `pnpm` during the installation, I'd prefer it to use `pnpx` instead

sacrosanctic:
closes #937

in this pr:
- remove reliance on a package manager by installing the cli as a dev dependency
- update pnpm install instructions to use corepack ([src](https://github.com/orgs/pnpm/discussions/9044) / [src](https://pnpm.io/installation#using-corepack))
- use `pnpm dlx` ([src](https://github.com/orgs/pnpm/discussions/6477))
- lock `create` and `add` args to a specific version for reproducibility

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -18,8 +18,7 @@ Please keep your pull requests focused to feature or issue. Focused smaller chan
 This is a monorepo, meaning the repo holds multiple packages. It requires the use of [pnpm](https://pnpm.io/). You can [install pnpm](https://pnpm.io/installation) with:
 
 ```sh
-npm i -g corepack@latest
-corepack enable pnpm
+npm i -g pnpm
 ```
 
 _(Optional)_ For running certain packages and tests locally you will need to install [docker](https://docs.docker.com/get-started/get-docker).


paoloricciuti:
An even better way could be to add `@better-auth/cli` as a dev dependency and have the command be `better-auth generate --config src/lib/server/auth.ts --output src/lib/server/db/auth.schema.ts --yes`...just tested and it works (and I prefer that to always install)

.changeset/sweet-moose-raise.md
```diff
@@ -0,0 +1,5 @@
+---
+"sv": patch
+---
+
+fix: install `@bettter-auth/cli` as a dev dependency
```
WarningImHack3r:
```suggestion
fix: install `@better-auth/cli` as a dev dependency
```
;)

documentation/docs/10-introduction/20-faq.md
```diff
@@ -7,7 +7,7 @@ title: Frequently asked questions
 Running `sv` looks slightly different for each package manager. Here is a list of the most common commands:
 
 - **npm** : `npx sv create`
-- **pnpm** : `pnpx sv create` or `pnpm dlx sv create`
+- **pnpm** : `pnpm dlx sv create`
```
WarningImHack3r, line 10: Even though as per your link in the PR description, `pnpx` is not _always_ there by default, I'm pretty sure it exists and is both used and preferred by the vast majority of pnpm users, and probably won't go away anytime soon. I'd argue to leave it as it was, or at least invert the original commands ("`pnpm dlx ...` or `pnpx ...`")

CONTRIBUTING.md
```diff
@@ -18,7 +18,8 @@ Please keep your pull requests focused to feature or issue. Focused smaller chan
 This is a monorepo, meaning the repo holds multiple packages. It requires the use of [pnpm](https://pnpm.io/). You can [install pnpm](https://pnpm.io/installation) with:
 
 ```sh
-npm i -g pnpm
+npm i -g corepack@latest
```
WarningImHack3r: Maybe add a line or two to explain why specifically use corepack over the user's install, if it even is mandatory to use the corepack version?
Your sources explain why corepack is still a viable install method, but don't mention any specific advantage over any other installation method. Maybe removing this block altogether and let the user install it however they want is simpler to manage?

sacrosanctic: > Corepack automatically installs and activates the required version when commands are executed. Developers no longer need to manually manage versions.
> https://corepack.org/

It is not typical in the repo to add that kind of context. By default, users expect us to provide best in class and best in practice information. If it is not, we simply update it, as I am here.

documentation/docs/10-introduction/20-faq.md
sacrosanctic: There are many reasons to use `pnpm dlx` over `pnpx`, the link has outlined some of them. But that's besides the point. Users should know how to use their own dev tools.

This list is only a reference to which package managers are supported. Everywhere else in the docs, we only reference `npm/npx`.

CONTRIBUTING.md
WarningImHack3r: Yes, but pnpm [does this by default](https://arc.net/l/quote/wosnkoda) since v10 (Jan 2025), no matter the installation method, hence my comment

But it's fair to enforce corepack, but maybe adding an explanation line would be nice to explain that

packages/sv/src/core/common.ts
```diff
@@ -144,7 +144,7 @@ export function buildAndLogArgs(
 	args: string[],
 	lastArgs: string[] = []
 ): string {
-	const allArgs = ['sv', command, ...args];
+	const allArgs = [`sv@${pkg.version}`, command, ...args];
```
jycouet, line 147: Hummm, not sure why we do this in this PR?
There is a specific need for it?

CONTRIBUTING.md
jycouet: I hear about `corepack` for a long time... and finally took the time to check (thx to this comment).

Yes, I find a bit strange to write "You need to install pnpm, do corepack". 

It's funny that we don't speak about `node` and `nvm` or `devbox` before that.
So, not sure about this one.

documentation/docs/10-introduction/20-faq.md
jycouet: Yes to `pnpm dlx`

packages/sv/src/core/common.ts
sacrosanctic: Was just tapping away and ended up grouping a few things together. I think its necessary for reproducibility.

CONTRIBUTING.md
sacrosanctic: I think the discourse largely stems from that fact that pnpm is not clear what is the proper way to install it. I believe corepack currently still handles edge cases better. 

Seeing there is push back, I will revert for now and request the maintainers to solidify their stance.

packages/sv/src/core/common.ts
jycouet: Ok, let's keep it in 👍

PR #952, opened 2026-02-27
fix: default to rune mode

### LINKED ISSUE #944: Should legacy mode be explained in the tutorial?
As long as legacy mode exists, there is a reasonable chance that it will cause problems.
I have seen a lot of confusion leading up the release of Svelte 5 due to the automatic mode switches.

If new projects are initialized with a global flag for runes mode, I think it would be fine to not explain it, but last time I checked, this would cause issues with component library dependencies that are legacy since the flag affects those as well, unless dynamic compiler options are used which exclude `node_modules`.

I think the lack of fine-grained reactivity in legacy is not that much of an issue, but people might work on imported, unreactive data without any runes and assignments trigger reactivity via legacy mode, so everything looks correct but when they later introduce a rune it mysteriously breaks.

Relates to:
- sveltejs/svelte.dev#1273

sacrosanctic:
closes #944

### DIFF
--- a/.changeset/yellow-hoops-tickle.md
+++ b/.changeset/yellow-hoops-tickle.md
@@ -1,5 +1,0 @@
----
-'@sveltejs/sv-utils': patch
----
-
-feat: creation of `@sveltejs/sv-utils` to build addons _(experimental)_
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -29,7 +29,7 @@ Linux users, you will have to ensure 'sudo' is not required. See [docker post in
 ```sh
 git clone https://github.com/sveltejs/cli.git
 cd cli
-pnpm install
+pnpm i
 ```
 
 ## Build and run
--- a/documentation/docs/10-introduction/10-overview.md
+++ b/documentation/docs/10-introduction/10-overview.md
@@ -6,7 +6,7 @@ The command line interface (CLI), `sv`, is a toolkit for creating and maintainin
 
 ## Usage
 
-The easiest way to run `sv` is with [`npx`](https://docs.npmjs.com/cli/v8/commands/npx) (or the equivalent command if you're using a different package manager — for example, `pnpx` if you're using [pnpm](https://pnpm.io/)):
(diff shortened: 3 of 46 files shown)


sacrosanctic:
why doesnt the snapshots fail?

jycouet:
> why doesnt the snapshots fail?

These files are not used _(I mean, not in creation)_
The idea originally was to be able to start these template directly in the repo.

You should probably tune these `packages/sv/src/create/shared/**/svelte.config.js`

sacrosanctic:
> These files are not used (I mean, not in creation)

That's confusing, can we not do that. They're not real repos.

jycouet:
> > These files are not used (I mean, not in creation)
> 
> That's confusing, can we not do that. They're not real repos.

I'm torn on this one. Yes, they are not real repos... But it's kinda nice to be able to start it, tweak it and save the update.
At the same time, since we do a lot of replacement, it's probably not that usefull!
I think I would like also to tweak a bit the strategy and have a `created` hook.
This should probably be part of another PR anyway.

sacrosanctic:
the tests pass, finally!

> I'm torn on this one.

we can move that discussion to https://github.com/sveltejs/cli/issues/956

packages/sv/src/cli/tests/snapshots/create-only/svelte.config.js
```diff
@@ -7,6 +7,9 @@ const config = {
 		// If your environment is not supported, or you settled on a specific environment, switch out the adapter.
 		// See https://svelte.dev/docs/kit/adapters for more information about adapters.
 		adapter: adapter()
+	},
+	vitePlugin: {
+		dynamicCompileOptions: ({ filename }) => ({ runes: !filename.includes('node_modules') })
```
JReinhold, line 12: I don't think this is 100% correct. Right now this is saying:

_"Enable runes-mode across all project files, but explicitly **disable** it for all files in `node_modules`, even those using runes"_.

I think it should be `undefined` in the `node_modules` case to go back to the auto-infer mode

```suggestion
		dynamicCompileOptions: ({ filename }) => ({ runes: filename.includes('node_modules') }) ? undefined : true
```

The newest release of `sv` is causing our CI to fail in Storybook for this exact reason, because we're using runes in `@storybook/svelte` and all the other packages. When I change the project's config to the above suggestion, it works fine.

PR #953, opened 2026-02-27
feat: add vscode extension recommendations

### LINKED ISSUE #792: Add VS Code `extensions.json` file
If VS Code specific files are going to be added, why not add recommended extensions?

- https://github.com/sveltejs/cli/pull/780

```json
{
  "recommendations": [
    "bradlc.vscode-tailwindcss",
    "dbaeumer.vscode-eslint",
    "esbenp.prettier-vscode",
    "svelte.svelte-vscode"
  ]
}
```

The newly added `settings.json` is part of the [VS Code Taliwind CSS extension setup](https://github.com/tailwindlabs/tailwindcss-intellisense#recommended-vs-code-settings):

```json
{
  "files.associations": {
    "*.css": "tailwind"
  }
}
```

Personally, I am using Zed and the `.vscode` directory is just a bloat for me.

sacrosanctic:
closes #792

aka give vscode first party support, and by extensions their forks.

### DIFF
--- a/packages/sv/src/addons/eslint.ts
+++ b/packages/sv/src/addons/eslint.ts
@@ -151,10 +151,7 @@ export default defineAddon({
 
 		sv.file(files.vscodeExtensions, (content) => {
 			const { data, generateCode } = parse.json(content);
-
-			data['recommendations'] ??= [];
-			data['recommendations'].push('dbaeumer.vscode-eslint');
-
+			json.arrayUpsert(data, 'recommendations', 'dbaeumer.vscode-eslint');
 			return generateCode();
 		});
 
--- a/packages/sv/src/addons/prettier.ts
+++ b/packages/sv/src/addons/prettier.ts
@@ -79,10 +79,7 @@ export default defineAddon({
 
 		sv.file(files.vscodeExtensions, (content) => {
 			const { data, generateCode } = parse.json(content);
-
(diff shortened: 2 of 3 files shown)


packages/sv/src/addons/eslint.ts
```diff
@@ -149,6 +149,15 @@ export default defineAddon({
 			return generateCode();
 		});
 
+		sv.file(files.vscodeExtensions, (content) => {
+			const { data, generateCode } = parse.json(content);
+
+			data['recommendations'] ??= [];
+			data['recommendations'].push('dbaeumer.vscode-eslint');
```
jycouet:
```suggestion
			json.arrayUpsert(data, 'recommendations', 'dbaeumer.vscode-eslint');
```

With `import { json } from '@sveltejs/sv-utils';`

jycouet (changes requested):
Thx for this batch 🎉

I think that you can update to use `json` utility.

sacrosanctic:
I would also use the shared folder but i dont think it works with folders atm

jycouet:
> I would also use the shared folder but i dont think it works with folders atm

Correct, it doesn't work like that today. I'm thinking that the `shared` folder could even go away since we have a lot of utils now to manage different update. With like a created hook? But this would be for later ✌️

Issue #941, opened 2026-02-23
Consider creating dedicated demo component for testing


Rich-Harris:
I just ran `sv add vitest` on an existing site and was slightly surprised to see it create `src/routes/page.svelte.spec.ts`. For one thing, the test it contained didn't pass, because it was making assumptions that might hold true for a freshly-baked app but won't necessarily hold true for existing apps.

But also: it feels weird to test components like `+page.svelte`. We don't know what props it might accept, or how it might interact with the layout, etc — if I want to test what happens when I visit `/` or any other route, that's really a job for an e2e test. I would typically expect a component test to belong to something in `lib`.

What if we created something like `src/lib/Demo.svelte` and `src/lib/Demo.svelte.spec.ts` instead? Maybe `demo.spec.ts` should go in `lib` too? (and also have a corresponding `demo.ts` module? It's slightly odd that the demo test isn't actually testing anything)

PR #960, opened 2026-03-01
feat(vitest): better real world vitest examples

jycouet:
Closes: #941

### DIFF
--- a/.changeset/cool-boxes-juggle.md
+++ b/.changeset/cool-boxes-juggle.md
@@ -1,0 +1,5 @@
+---
+"sv": patch
+---
+
+fix(better-auth): resolve `auth:schema` failure on fresh installs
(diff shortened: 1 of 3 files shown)


manuel3108:
seems like the tests are failing pretty harshly currently

jycouet:
> seems like the tests are failing pretty harshly currently

Better-auth main issue is fixed now in main, I think that it should be better

manuel3108:
Ok, i think that looks good now. Should we do the same for playwright then? I guess it would make sense right?

jycouet:
What same do you have in mind ? 👀😁

manuel3108:
Sorry, it makes the same assumptions vitest was making. It expects that +page.svelte contains an h1, which might not be given for existing apps

jycouet:
So we could do a demo page for it ?
(in another PR ?)

PR #968, opened 2026-03-03
fix(better-auth: auth:schema command fails on fresh install due to broken @better-auth/cli

### LINKED ISSUE #961: fix(better-auth): auth:schema command fails on fresh install due to broken @better-auth/cli
## Bug Description

When scaffolding a new project with the `better-auth` addon via `sv create`,
the generated `auth:schema` script fails immediately on a fresh install —
no user error involved.

## Steps to Reproduce
```sh
pnpx sv create
# select: better-auth addon
cd my-project
pnpm run auth:schema
```

## Error Output
```
> better-auth generate --config src/lib/server/auth.ts --output src/lib/server/db/auth.schema.ts --yes

SyntaxError: The requested module 'better-auth/db' does not provide an export named 'getMigrations'
    at ModuleJob._instantiate (node:internal/modules/esm/module_job:226:21)
```

## Root Cause

The `better-auth` addon installs `@better-auth/cli` at `^1.4.18`.
Versions `1.4.6`–`1.4.20` of `@better-auth/cli` internally resolve to a
version of `better-auth` that does not export `getMigrations` from
`better-auth/db`, causing the crash on a fresh install.

A `1.4.21` patch was released that resolves the immediate crash.

Full investigation and maintainer response here:
https://github.com/better-auth/better-auth/issues/8257

## Proposed Fix

**Option 1 — Migrate to the new `auth` CLI:**

The better-auth team appears to have moved to a new `auth` CLI, as their
documentation has been updated to use it:
https://better-auth.com/docs/concepts/cli

Remove the `@better-auth/cli` dev dependency:
```diff
- sv.devDependency('@better-auth/cli', '^1.4.18');
```

Update the `auth:schema` script:
```diff
- `better-auth generate --config ${authConfigPath} --output ${authSchemaPath} --yes`
+ `npx auth@latest generate --config ${authConfigPath} --output ${authSchemaPath} --yes`
```

`npx auth@latest` is used here rather than `pnpm dlx auth` so the script
works regardless of which package manager the user chose.

**Option 2 — Bump `@better-auth/cli` to `>=1.4.21`:**

Bump the pinned version to avoid the broken version range.

---

Happy to open a PR for whichever approach maintainers prefer.

## Environment

- `sv` version: `0.12.4`
- `better-auth` version: `^1.4.18`
- `@better-auth/cli` installed version: `1.4.20`
- Node.js: `v22.21.0`
- OS: Windows 11

sacrosanctic:
closes #961

### DIFF
--- a/packages/sv/src/addons/better-auth.ts
+++ b/packages/sv/src/addons/better-auth.ts
@@ -50,7 +50,7 @@ export default defineAddon({
 
 		let drizzleDialect: Dialect;
 
-		sv.devDependency('better-auth', '^1.4.18');
+		sv.devDependency('better-auth', '~1.4.21');
 		sv.devDependency('@better-auth/cli', '~1.4.21');
 
 		sv.file(`drizzle.config.${language}`, (content) => {
--- a/packages/sv/src/cli/tests/snapshots/create-with-all-addons/package.json
+++ b/packages/sv/src/cli/tests/snapshots/create-with-all-addons/package.json
@@ -22,7 +22,7 @@
 		"auth:schema": "better-auth generate --config src/lib/server/auth.ts --output src/lib/server/db/auth.schema.ts --yes"
 	},
 	"devDependencies": {
-		"@better-auth/cli": "^1.4.21",
+		"@better-auth/cli": "~1.4.21",
 		"@eslint/compat": "^2.0.2",
 		"@eslint/js": "^9.39.2",
 		"@inlang/paraglide-js": "^2.10.0",
@@ -35,7 +35,7 @@
 		"@tailwindcss/typography": "^0.5.19",
 		"@tailwindcss/vite": "^4.1.18",
 		"@vitest/browser-playwright": "^4.0.18",
-		"better-auth": "^1.4.18",
+		"better-auth": "~1.4.21",
 		"drizzle-kit": "^0.31.8",
 		"drizzle-orm": "^0.45.1",
 		"eslint": "^9.39.2",


packages/sv/src/addons/better-auth.ts
```diff
@@ -51,7 +51,7 @@ export default defineAddon({
 		let drizzleDialect: Dialect;
 
 		sv.devDependency('better-auth', '^1.4.18');
-		sv.devDependency('@better-auth/cli', '^1.4.18');
+		sv.devDependency('@better-auth/cli', '^1.4.21');
```
bennajah: use latest cli version 

                sv.devDependency('auth', '^1.5.2');

packages/sv/src/cli/tests/snapshots/create-with-all-addons/package.json
```diff
@@ -22,7 +22,7 @@
 		"auth:schema": "better-auth generate --config src/lib/server/auth.ts --output src/lib/server/db/auth.schema.ts --yes"
 	},
 	"devDependencies": {
-		"@better-auth/cli": "^1.4.18",
+		"@better-auth/cli": "^1.4.21",
```
bennajah: "auth": "^1.5.2",

packages/sv/src/addons/better-auth.ts
sacrosanctic: This is a quick fix. Ideally, both dependencies should be in sync. And we shouldn't bump to 1.5 until we migrate the whole plugin

jycouet: I update from `^` to `~` to be sure to say on patch level.

In a next PR we should thing about fully going to `1.5` (cli & lib!)

PR #971, opened 2026-03-05
fix(sv): remove spread operator from ESLint config generation

### LINKED ISSUE #967: spread syntax in eslint's `defineConfig` is redundant
It took me a long time to find this information.

https://github.com/eslint/eslint/discussions/17834#discussioncomment-13171715
https://eslint.org/blog/2025/03/flat-config-extends-define-config-global-ignores/#introducing-defineconfig()-for-eslint

sacrosanctic:
closes #967

### DIFF
--- a/packages/sv/src/cli/tests/cli.ts
+++ b/packages/sv/src/cli/tests/cli.ts
@@ -45,7 +45,7 @@ describe('cli', () => {
 
 	it.for(testCases)(
 		'should create a new project with name $projectName',
-		{ timeout: 51_000 },
+		{ timeout: 57_000 },
 		async (testCase) => {
 			const { projectName, args, template = 'minimal' } = testCase;
 


manuel3108:
@sacrosanctic could you maybe have another look at this? Looks like the tests are failing because of this.

sacrosanctic:
~~do not merge!~~

i believe `defineConfig` should be used for js template as well.

sacrosanctic:
It appears that in ts, you dont need a root array,
```ts
// https://typescript-eslint.io/getting-started/#step-2-configuration
export default defineConfig(...)
```
but in js, you do
```js
// https://eslint.org/docs/latest/use/getting-started#configuration
export default defineConfig([...])
```
This is supported in their does but not explicitly called out. And it [isn't always consistent](https://eslint.org/blog/2025/03/flat-config-extends-define-config-global-ignores/#introducing-defineconfig()-for-eslint).

PR #975, opened 2026-03-07
chore: bump tsdown `0.21`

jycouet:
Main change is: `tsdown.config.ts` (and bump of tsdown)


jycouet:
> I still don't agree the changes that were introduced by tsdown. But since this seems to be the way forward we will probably just have to follow that for now.
> 
> Please merge if you don't see any other chances in getting rid of that config.

I'm also not a big fan of this change. But, I did a PR there to at least be informed when one dep is not needed anymore...
Let's merge for now and see where we go after :)

PR #976, opened 2026-03-10
feat(drizzle): add cloudflare `D1` database

### LINKED ISSUE #951: cloudflare D1 database with drizzle
### Describe the problem

when init a new svelte project the CLI ask how u would like setup your new project.
this is great DX, but i think this would be better if you add **cloudflare D1** and **D1 driver** with drizzle 

### Describe the proposed solution

|  sveltekit-adapter: Which SvelteKit adapter would you like to use?
│  cloudflare
│
◇  sveltekit-adapter: Are you deploying to Workers (assets) or Pages?
│  Workers
drizzle: Which database would you like to use?
│  cloudflare D1
│
◇  drizzle: Which SQLite client would you like to use?
│  d1

### Importance

would make my life easier

sacrosanctic:
closes #951

`auth` is now a factory. Would like it not to be, is there another way to get platform info.

## todo
- [x] update `wrangler`
- [x] update `better-auth`
- [x] show `D1` only for cloudflare adapter or no/auto adapter

### DIFF
--- a/packages/sv/src/cli/tests/snapshots/create-with-all-addons/src/lib/server/auth.ts
+++ b/packages/sv/src/cli/tests/snapshots/create-with-all-addons/src/lib/server/auth.ts
@@ -16,5 +16,7 @@ export const auth = betterAuth({
 			clientSecret: env.GITHUB_CLIENT_SECRET
 		}
 	},
-	plugins: [sveltekitCookies(getRequestEvent)] // make sure this is the last plugin in the array
+	plugins: [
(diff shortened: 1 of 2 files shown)


jycouet:
@sacrosanctic I think that it's the other way around, not `show D1 only for cloudflare adapter or no/auto adapter`, but add the good adapter if you select D1

I let you have a look :)
Thx for looking at this

sacrosanctic:
@jycouet ready

Issue #987, opened 2026-03-13
Runes mode enforcement is currently too hamfisted


GrygrFlzr:
Followup to #944 #952 and #984 
Currently, the default config generated by sv cli uses this function:
```js
dynamicCompileOptions: ({ filename }) =>
			filename.includes('node_modules') ? undefined : { runes: true }
```

This currently just checks for the existence of `node_modules` in the filename path for determining if the compiler should use runes mode.
As a result:
1. It does not respect any existing `<svelte:options runes={false}>` in `.svelte` files
2. It unintentionally forces runes mode for generated files in the `.svelte-kit` folder
3. It unintentionally skips a path like `src/lib/leafnode_modules`, or all paths if the username happens to be `node_modules`

A more complete solution could be instead to scope the changes to the `src` folder, and also skip `.svelte` files that contain `<svelte:options` as a "I know what I'm doing" hint.
Example function:
```js
import { extname, join, relative } from 'node:path';
const srcDir = join(import.meta.dirname, 'src');
/** @type {import('@sveltejs/vite-plugin-svelte').PluginOptions['dynamicCompileOptions']} */
const dynamicCompileOptions = ({ filename, code }) => {
	const isOutOfSrc = relative(srcDir, filename).startsWith('..');
	// .svelte.ts and .svelte.js definitely require runes mode
	const isSFCWithOptionDeclared =
		extname(filename) === 'svelte' && code.includes('<svelte:options');

	return isOutOfSrc || isSFCWithOptionDeclared ? undefined : { runes: true };
};
```
However, I think sv cli is not the right place to apply this anyway, because if e.g. we want to update the function, that requires all users with generated configs to manually change their config; it is an inherently fragile execution of the concept.

Even the function I provided would be fragile against the user configuring a custom `kit.files.src` path.

jycouet:
Thx for looking into this @GrygrFlzr 

On your points
1. I was thinking that `<svelte:options` in `.svelte` files is taking precedence to `dynamicCompileOptions`. It's not the case? I'll check that.
2. In `.svelte-kit`, we have `generated/root.svelte` with `<svelte:options runes={true} />`, can we have other files? I'll investagate
3. You are right, your `isOutOfSrc` look more robust.

Yes, if it start to be a complexe thing, I fully agree, it should not be in `sv`.
Here originally, the intent was: "Hey, in new projects, default to rune". It seems to not be that straight forward to do that, and we should fix it somewhere.

I'll report back later today ✌️

[...]

PR #991, opened 2026-03-17
fix(sv): stricter logic for rune mode

sacrosanctic:
#987

code sourced from https://github.com/sveltejs/cli/issues/987#issuecomment-4053378160

Also add a comment to remove it in svelte 6


manuel3108:
We might want to wait for https://github.com/sveltejs/svelte/pull/17951 to be released, so that we can make the changes in the correct place as well

manuel3108:
Looks fine to me and work, would like a review from @jycouet as well, since he was part of the initial implementation if i remember correctly. Should also solve the issue Jeppe (storybook) reported

PR #943, opened 2026-02-23
feat(eslint): add `recommendedTypeChecked` config

jycouet:
Closes #938

PR #994, opened 2026-03-18
fix(eslint): account for all config formats

sacrosanctic:
I spotted the problem when looking at https://github.com/hyunbinseo/svelte-kit-templates/blob/b200ae76d7bb2c61e6e695c4d90912666d39256c/javascript/eslint.config.js. I had to test eslint by hand.

We should add a test, could someone lend me hand?

That repo was quite helpful, maybe we can expand on the templates we take snapshots for?

- [x] missing type

### DIFF
--- a/packages/sv/src/addons/eslint.ts
+++ b/packages/sv/src/addons/eslint.ts
@@ -122,7 +122,7 @@ export default defineAddon({
 				});
 				comments.add(rulesOverride.properties[0].key, {
 					type: 'Line',
-					value: " 'svelte/rule-name': 'error'"
+					value: " 'svelte/button-has-type': 'error'"
 				});
 
 				const exportExpression = js.functions.createCall({ name: 'defineConfig', args: [] });
--- a/packages/sv/src/cli/tests/snapshots/create-with-all-addons/eslint.config.js
+++ b/packages/sv/src/cli/tests/snapshots/create-with-all-addons/eslint.config.js
@@ -38,7 +38,7 @@ export default defineConfig(
 	},
 	{
 		// Override or add rule settings here, such as:
-		// 'svelte/rule-name': 'error'
+		// 'svelte/button-has-type': 'error'
 		rules: {}
 	}
 );


packages/sv/src/addons/common.ts
```diff
@@ -38,23 +41,39 @@ export function addEslintConfigPrettier(content: string): string {
...
-	const idx = elements.findIndex(
-		(el) =>
-			el?.type === 'MemberExpression' &&
-			el.object.type === 'MemberExpression' &&
-			el.object.property.type === 'Identifier' &&
-			el.object.property.name === 'configs' &&
-			el.object.object.type === 'Identifier' &&
-			el.object.object.name === svelteImportName
-	);
+	// i dont know the type
```
manuel3108: TODO

packages/sv/src/addons/common.ts
```diff
@@ -38,23 +41,39 @@ export function addEslintConfigPrettier(content: string): string {
...
+		return (
+			maybeSpread?.object?.type === 'MemberExpression' &&
+			maybeSpread?.object?.property?.type === 'Identifier' &&
+			maybeSpread?.object?.property?.name === 'configs' &&
+			maybeSpread?.object?.object?.type === 'Identifier' &&
+			maybeSpread?.object?.object?.name === svelteImportName
+		);
+	};
+
+	// i dont know the type
```
manuel3108: TODO

sacrosanctic:
> Which problem exactly are you trying to solve in this file?

@manuel3108 

`prettier` and `svelte.configs.prettier` are supposed to be right after `svelte.configs.recommended`.

The original only partially supported the following cases, `[]`, `defineConfig()`, and `defineConfig([])`. It now also handles `svelte.configs.recommended` whether it is spread or not.

sacrosanctic:
@jycouet

Added some of the changes from https://github.com/sveltejs/cli/pull/943 to lighten the load there.

Is the CI telling me i need to make that change in the svelte.dev repo?

Should we have more snapshots?

packages/sv/src/addons/better-auth.ts
```diff
@@ -305,8 +305,8 @@ export default defineAddon({
 						? `
 						signInEmail: async (event) => {${d1AuthLine}
 							const formData = await event.request.formData();
-							const email = formData.get('email')?.toString() ?? '';
-							const password = formData.get('password')?.toString() ?? '';
+							const email = typeof formData.get('email') === 'string' ? formData.get('email') : '';
+							const password = typeof formData.get('password') === 'string' ? formData.get('password') : '';
```
jycouet: All these changes were problematic in some eslint config?!

sacrosanctic: removed

sacrosanctic:
Not too familiar with the AST, cannot move the comment inside the object.

jycouet:
> Not too familiar with the AST, cannot move the comment inside the object.

Nah, I don't manage too. `esrap` manage before or after node, but not inside. (I think it's not needed, it's ok)
I updated with 1 existing rule. WDYT? (if you have a better rule idea, that's good for me too)

Let me know

PR #999, opened 2026-03-21
rename to `file`, `isKit`, `directory`

jycouet:
Singular form — `file.viteConfig` reads better than `files.viteConfig` since each property refers to a single file path.

Pure mechanical rename, no logic changes.

Prep for community add-on improvements.

---

I'm thinking also to switch: 
```ts
/** If we are in a kit project, this object will contain the lib and routes directories */
kit: { libDirectory: string; routesDirectory: string } | undefined;
```
to 
```ts
isKit: boolean
directory: {
  lib: string,
  routes: string
}
```
So that directory has good default in non-kit as well!

### DIFF
--- a/packages/sv/src/core/workspace.ts
+++ b/packages/sv/src/core/workspace.ts
@@ -49,8 +49,9 @@ export type Workspace = {
 	isKit: boolean;
 	directory: {
 		src: string;
+		/** In SvelteKit taking `kit.files.lib` automatically. Falls back to `src/lib` in non-Kit projects */
 		lib: string;
-		/** SvelteKit routes directory, taking `kit.files.routes` automatically. Falls back to `src/routes` in non-Kit projects */
+		/** In SvelteKit taking `kit.files.routes` automatically. Falls back to `src/routes` in non-Kit projects */
 		kitRoutes: string;
 	};
 	/** The package manager used to install dependencies */


jycouet:
> But having directory.routes in a non kit project is irritating for me.

Me too :D But it was "simple". Maybe I should have just added a "good" `JSDoc` on top.
Instead I tried with another thing: `kitRoutes` to be more explicite. (with a fallback even in non kit... to not fail silently. I tried with type narrowing, but since we destructure everything... it doesn't work)

WDYT like this? I'm hesitant with this or just `routes` with JSDocs.
I'll wait to merge :D

packages/sv/src/core/workspace.ts
```diff
@@ -46,8 +46,13 @@ export type Workspace = {
 		/** Get the relative path between two files */
 		getRelative: ({ from, to }: { from?: string; to: string }) => string;
 	};
-	/** If we are in a kit project, this object will contain the lib and routes directories */
-	kit: { libDirectory: string; routesDirectory: string } | undefined;
+	isKit: boolean;
+	directory: {
+		src: string;
+		lib: string;
+		/** SvelteKit routes directory, taking `kit.files.routes` automatically. Falls back to `src/routes` in non-Kit projects */
```
manuel3108: If I remember correctly, the same applies for kit.files.lib right? We should either be consistent with the comments or leave it :D

jycouet: So heuuu, let's not merge this right now, I have to look more into https://github.com/sveltejs/kit/discussions/14240

manuel3108: Based on the last comment, we are not deprecating this in the near future: https://github.com/sveltejs/kit/discussions/14240#discussioncomment-15982544

PR #1003, opened 2026-03-21
ci: shard Windows tests across 2 runners

jycouet:
Speed up CI by splitting Windows tests into 2 in parallel.

Cutting like 2 minutes... not sure it's worth it!


manuel3108:
It's just ci, don't think we need it more complicated than necessary.

PR #1006, opened 2026-03-24
Add configurable prettier options

langpavel:
Expose `useTabs`, `singleQuote`, `trailingComma`, and `printWidth` as addon options so users can customize prettier defaults via CLI or interactive prompts.


jycouet:
That's a lot of questions, I'm not really in favor of this one.
I'll close for now

PR #1014, opened 2026-03-30
clean things up

sacrosanctic:
edits on #1001

### DIFF
--- a/packages/sv/src/addons/drizzle.ts
+++ b/packages/sv/src/addons/drizzle.ts
@@ -141,14 +141,12 @@ export default defineAddon({
 				'docker-compose.yml' // for backward compatibility
 			];
 
-			const composeFile = composeFileOptions.find((option) =>
-				fs.existsSync(path.resolve(cwd, option))
-			);
-
-			if (!composeFile) throw new Error('unreachable state...');
+			const composeFile =
+				composeFileOptions.find((option) => fs.existsSync(path.resolve(cwd, option))) ??
+				'compose.yaml';
 
 			sv.file(composeFile, (content) => {
-				// `transform.yaml` not implemented. Therefore, abort if file exist.
+				// `transforms.yaml` not implemented. Therefore, abort if file exist.
 				if (content.length > 0) return false;
 
 				const imageName = options.database === 'mysql' ? 'mysql' : 'postgres';


packages/sv/src/addons/better-auth.ts
```diff
@@ -296,11 +289,11 @@ export default defineAddon({
 
 			sv.file(
 				`${directory.kitRoutes}/demo/better-auth/login/+page.server.${language}`,
-				transforms.text(({ content }) => {
+				(content) => {
 					if (content) {
 						const filePath = `${directory.kitRoutes}/demo/better-auth/login/+page.server.${language}`;
 						log.warn(`Existing ${color.warning(filePath)} file. Could not update.`);
-						return false;
+						return content;
```
jycouet: We should maybe allow to return false meaning "not modified".
WDYT?

packages/sv/src/addons/drizzle.ts
```diff
@@ -514,68 +505,69 @@ export default defineAddon({
...
-		content = text.upsert(content, 'CLOUDFLARE_ACCOUNT_ID', {
-			value: '""',
-			comment: ['Cloudflare D1'],
-			separator: true
-		});
-		content = text.upsert(content, 'CLOUDFLARE_DATABASE_ID', { value: '""' });
-		content = text.upsert(content, 'CLOUDFLARE_D1_TOKEN', { value: '""' });
-		return content;
-	}
+) => (content: string) => string;
```
jycouet: I see what you did there ^^. 
I'm not a big fan of `=> ... => ...`, I would like to limit as much as possible.

Maybe we could pass `text` if we want to avoid the top import?

jycouet: `TransformFn` is helping ;)
Perfect

packages/sv/src/addons/better-auth.ts
sacrosanctic: Ya, I think that's better. Otherwise, switching between raw and transform is confusing.

packages/sv/src/addons/drizzle.ts
sacrosanctic: should intellisense show `TransformFn` too?

jycouet: I think it's ok

packages/sv/src/addons/better-auth.ts
jycouet: But both should have `string | false` maybe?
I think we run prettier there to all files touched. Maybe when it's false we can not register these files?

PR #1024, opened 2026-04-01
feat(sv-utils): `color.command` now also accepts `string[]`

sacrosanctic:


jycouet:
@sacrosanctic what do you think about this? 👀
_experiment!_

sacrosanctic:
looks good, wish the fn name wasnt so long

jycouet:
> looks good, wish the fn name wasnt so long

Same here ! 
What about `formatCommand` ? (and a small JSDoc)?

I'm not sure about your last 2 commits? 👀
Is it ok with intellisense? (I didn't check yet)

sacrosanctic:
> What about formatCommand ? (and a small JSDoc)?

I feel like we have to keep it as `resolveCommandArray`. To discourage similar types of interface changing behaviour.

> I'm not sure about your last 2 commits? 👀

They're non functional changes, feel free to reverse it, I like that the logic and data is now cleanly separated.

> Is it ok with intellisense? (I didn't check yet)

Yes? What the problem?

jycouet:
With `Record<string, ...`, I could do `color.hello("")` and types will be happy.
I tried with `as const` or with `satisfies`, but none are working as we want.

Let's just go back to a simpler solution _(adding a few characters)_

PR #1057, opened 2026-04-08
fix(adapter-cloudflare): wrangler types in check/build scripts

### LINKED ISSUE #872: Skip generating wrangler types when building for deployment

jycouet:
Closes #872

### Description

-  wrangler types are no more gitignored
- use `--check` flag for wrangler types in check/build scripts

### Checklist

### DIFF
--- a/.changeset/brave-clouds-check.md
+++ b/.changeset/brave-clouds-check.md
@@ -1,0 +1,5 @@
+---
+'sv': patch
+---
+
+fix(adapter-cloudflare): use `--check` flag for wrangler types in check/build scripts
--- a/packages/sv/src/addons/sveltekit-adapter.ts
+++ b/packages/sv/src/addons/sveltekit-adapter.ts
@@ -178,21 +178,13 @@ export default defineAddon({
 			);
 
 			if (file.typeConfig) {
-				sv.file(
-					file.gitignore,
-					transforms.text(({ content }) => {
-						if (content.length === 0) return false;
-						return text.upsert(content, '/worker-configuration.d.ts', {
-							comment: 'Cloudflare Types'
-						});
-					})
-				);
-
-				// Setup wrangler types command
+				// Setup wrangler types command and prepend to check/build
 				sv.file(
 					file.package,
 					transforms.json(({ data, json }) => {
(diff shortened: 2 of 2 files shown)


petebacondarwin:
This actually breaks the create-cloudflare ecosystem tests for SvelteKit: 

https://github.com/cloudflare/workers-sdk/actions/runs/24253310618/job/70830164771?pr=13330#step:8:627

The `--check` requires that the types had previously been generated.

jycouet:
:( can't look right now
It's because types are not commited ? 
Can you provide what scripts do you suggest ?

petebacondarwin:
When the project is first created no typings have been generated - but now running `npm run build` will fail.
This `--check` command is designed to fail if the types are out of date, perhaps in CI jobs where you want to ensure that the checked in types are accurate.

But in this case I would argue that you really just want to run `wrangler types` without the `--check` to just generate types as part of the build.

jycouet:
In a "normal" case don't you want to break the build if types are out-of-date ?

petebacondarwin:
OK, so I see that you stopped generating it automatically due to the time that can take (although I believe that we should check the hash stored in the file and not re-run the generation if it has not changed so should be super fast in that case).

Because of that it is necessary to ensure that you have run `pnpm run gen` at least once before you run `pnpm run build`. In this case, we can change Wrangler's autoconfig setup to run this as part of setting up a project for Cloudflare.

But I would argue that if things are working correctly, there should be very little cost to running `wrangler types` regularly. In which case you don't need to run the `--check` command and will not see any errors.

jycouet:
All I would like is: 
- Best practice from expert
- Easy onboarding for new users
(unfortunately, I'm not a cloudflare user today) 

What's the normal way to work ? Generate often ? Break build ? Break check ? Never break ? 

Auto generating at build time would silence potential issue later at runtime, no ?

We can do everything, we just need to decide.

petebacondarwin:
We fixed it in our CI tests by running the type generation before running the build. If your users are happy with this flow then it's fine by us.

PR #1064, opened 2026-04-17
fix(sv): scope @deprecated tag to legacy create(cwd, options) overload

### LINKED ISSUE #1063: Types for `create` deprecate function entirely
The types changes in #1046 that were intended to just deprecate the old `create(cwd, options)` call appear to apply to the export as a whole. I'm getting a strikethrough on `create` in `import { create } from 'sv'`, and not just on the places where I'm calling `create` in the old way.

mvanhorn:
Closes #1063

## Problem

The `@deprecated` JSDoc tag added in #1046 was intended to mark only the legacy `create(cwd, options)` positional form, but in practice editors strike through every reference to the `create` export — including correct calls like `import { create } from 'sv'`.

That happens because VS Code and tsserver resolve a symbol's JSDoc from the **first** overload in source order. Since the deprecated overload was listed first, its `@deprecated` tag became the canonical documentation for the symbol itself.

## Fix

Reorder the overloads so the non-deprecated `create(options)` signature is declared first. The `@deprecated` tag now annotates only the legacy `create(cwd, options)` overload, matching the original intent of #1046.

Before:
- `import { create }` → shown as deprecated
- `create({ cwd, ... })` → shown as deprecated
- `create('./', { ... })` → shown as deprecated

After:
- `import { create }` → not deprecated
- `create({ cwd, ... })` → not deprecated
- `create('./', { ... })` → deprecated (as intended)

## Testing

- `pnpm build` regenerates `packages/sv/api-surface.md` with the overload order swapped (included in the diff)
- `pnpm check` passes
- `pnpm lint` passes

Changeset (`patch`) included.

### DIFF
--- a/.changeset/fix-create-deprecation-scope.md
+++ b/.changeset/fix-create-deprecation-scope.md
@@ -1,13 +1,5 @@
 ---
-"sv": patch
+'sv': patch
 ---
 
-fix: scope `@deprecated` tag to the legacy `create(cwd, options)` overload
-
-The JSDoc `@deprecated` on the first overload signature caused editors to
-mark the entire `create` export as deprecated, including the supported
-`create({ cwd, ...options })` call. Reorder the overloads so the
-non-deprecated signature comes first; editors now only strike through the
-legacy positional form.
-
-Closes #1063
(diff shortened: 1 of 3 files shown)


jycouet (approved):
Thx

PR #1065, opened 2026-04-17
fix(sv): shift order of function signatures to avoid incorrect import-level deprecation

### LINKED ISSUE #1063: Types for `create` deprecate function entirely

GrygrFlzr:
Closes #1063

### Description

VS Code's import strikethrough uses the first-found function declaration, so we just re-order to fix the issue.
Note this doesn't impact call-site strikethrough, which already do check that the arguments match.

### Checklist

- ~~Update snapshots~~ (not applicable)
- [x] Add a changeset

### DIFF
--- a/.changeset/lovely-hotels-yawn.md
+++ b/.changeset/lovely-hotels-yawn.md
@@ -1,0 +1,5 @@
+---
+'sv': patch
+---
+
+fix(sv): declare non-deprecated function definition first to avoid import-level deprecation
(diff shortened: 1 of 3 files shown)


GrygrFlzr:
Looks like #1064 was submitted already

PR #989, opened 2026-03-15
chore(sv): bump ESLint to v10 and update related dependencies

jycouet:
Closes #930

## Summary

- Bump `eslint` from `^9.39.2` to `^10.0.3` and `@eslint/js` from `^9.39.2` to `^10.0.1`
- Bump `eslint-plugin-svelte` to `^3.15.2`, `typescript-eslint` to `^8.57.0`, `@eslint/compat` to `^2.0.3`, `globals` to `^17.4.0`
- Bump `@sveltejs/eslint-config` to `^9.0.0` (repo devDependency)
- Update snapshot and lockfile

PR #1069, opened 2026-04-28
fix(sv): align eslint version to `10` accross all addons

jonasgeiler:
Firstly, I apologize for not creating an issue before making this PR, I hope that's okay.
If you want I can still create and link an issue, instead of describing the problem in this PR.

### Description

I have noticed that while creating a new project using the `sv create` command, and _**first**_ selecting the `eslint` add-on, and _**afterwards**_ selecting the `prettier` add-on (or just using the command `pnpm dlx sv@0.15.1 create --template library --no-types --add eslint prettier --install pnpm example-project`) that it always shows the following warning message:
```
An older major version of eslint was detected. Skipping eslint-config-prettier installation.
```
I thought that didn't make any sense, since I was creating a fresh new project, so I dug deeper and saw that the `SUPPORTED_ESLINT_VERSION` constant in the `prettier.ts` add-on definition had an outdated ESLint major version `'9'`, which caused a version mismatch when installing it _**after**_ the ESLint add-on.
Installing the Prettier add-on first didn't cause this problem, since the add-on script doesn't detect `eslint` yet and skips installing `eslint-config-prettier`, handing it over to the ESLint add-on, which happily installs `eslint-config-prettier` without checking it's own version.

I guess the problem was introduced in #989, where they bumped ESLint to v10, but forgot to update the (somewhat hidden) `SUPPORTED_ESLINT_VERSION` constant.

Now I could have just set the `SUPPORTED_ESLINT_VERSION` constant to `'10'` and call it a day, but I thought I'd rather future-proof this, so it doesn't happen again in the future.
For this, I have create a common `ESLINT_VERSION` constant (which I made sure also gets updated automatically by `pnpm run update-deps`), shared between the ESLint and Prettier add-ons.
The ESLint add-on just installs the version defined by that constant, but the Prettier add-on extracts the major version from that constant, and uses it in place of the previously hard-coded value.
I have also heavily modified the Prettier add-on tests, to make sure `eslint-config-prettier` gets installed, depending on the installed `eslint` version.

### Checklist

- [X] Update [snapshots](../CONTRIBUTING.md#update-snapshots) (if applicable)
- [X] Add a [changeset](../CONTRIBUTING.md#generating-changelogs) (if applicable)
- [X] Allow maintainers to edit this PR
- [X] I care about what I'm doing, no matter the tool I use (Notepad, Sublime, VSCode, AI...)
  > Note: I have used GitHub Copilot's auto-complete feature for this contribution, however, no AI Agents where used. Most of the changes were written and double checked by myself, to the best of my knowledge.


packages/sv/src/core/common.ts
```diff
@@ -282,6 +282,11 @@ export const normalizePosix = (dir: string) => {
 	return path.posix.normalize(dir.replace(/\\/g, '/'));
 };
 
+export function stripVersionRange(versionRange: string): string {
+	// Removes the version range (e.g. `^` is removed from: `^9.0.0`)
+	return versionRange.replaceAll(/[^\d|.]/g, '');
+}
```
jycouet: I'm wondering if this could go in `packages/sv-utils/src/common.ts` ?

packages/sv/src/core/tests/common.ts
```diff
@@ -51,3 +51,11 @@ describe('parseAddonOptions', () => {
...
 	});
 });
+
+describe('stripVersionRange', () => {
+	it('removes range prefixes and non-version characters', () => {
+		expect(stripVersionRange('^9.0.0')).toBe('9.0.0');
+		expect(stripVersionRange('~1.2.3')).toBe('1.2.3');
+		expect(stripVersionRange('workspace:^5.4.3')).toBe('5.4.3');
+	});
+});
```
jycouet: Maybe it should throw on `latest`, `workspace:*`, what do you think?

packages/sv/src/addons/prettier.ts
```diff
@@ -103,7 +104,7 @@ export default defineAddon({
 	}
 });
 
-const SUPPORTED_ESLINT_VERSION = '9';
+const SUPPORTED_ESLINT_VERSION = stripVersionRange(ESLINT_VERSION).split('.')[0];
```
jycouet: Maybe it could then use `splitVersion` from `sv-utils`? And manage the undefined case?

packages/sv/src/addons/prettier.ts
```diff
@@ -103,7 +104,7 @@ export default defineAddon({
...
 
 function hasEslint(version: string | undefined): boolean {
 	return !!version && version.startsWith(SUPPORTED_ESLINT_VERSION);
```
jycouet: This was not changed now, but looking at it, we could probably refactor a bit to return 2 things from this function: `installed` & `supported`. 
Like this `SUPPORTED_ESLINT_VERSION` goes inside of the function also.

jycouet (commented):
Thank you for your PR and looking at it. 
Very well explained 👍

I let you have a look at the comments? Then will test it out

packages/sv/src/addons/prettier.ts
jonasgeiler: Yeah I thought so too, because we're basically doing `version.startsWith(SUPPORTED_ESLINT_VERSION)` twice at the moment.
I didn't do it in the end because I tried to keep the changes to a minimum in this PR, but since you suggested the same, I can definitely work out a cleaner solution for this.

packages/sv/src/core/common.ts
jonasgeiler: I initially put it there, but then I thought that the function is a bit too "rough" (or "dumb", if you will) to be exposed to add-on developers as a utility, and should stay internal for now.
A proper version range parsing/stripping function would be much more involved than this, potentially even returning an array for cases where `||` is used, like `<1.0.0 || >=2.3.1 <2.4.5 || >=2.5.2 <3.0.0`. It can get really complicated really fast 😅

So if you want a proper production-ready utility, it would probably be best to use an existing solution, like [npm's own `semver` library](https://github.com/npm/node-semver), which simplifies working with version specifiers and ranges.
They provide a `semver.minVersion(...)` function that would probably be the best suited for this task:
```js
const semverMinVersion = require('semver/ranges/min-version'); // I hope this work with ESM too...

semverMinVersion('>=1.0.0 || >=2.3.1 <2.4.5 || >=2.5.2 <3.0.0').version // '1.0.0'
semverMinVersion('^9.0.0').version // '9.0.0'
semverMinVersion('2.x').version // '2.0.0'
semverMinVersion('~1.2.3').version // '1.2.3'
semverMinVersion('*').version // '0.0.0'
```
There is also a `semver.clean(...)` function but it seems to not work with version ranges, only exact versions like `=9.0.0`.

Another wholly different solution could be using the lockfiles content, or literally looking into the `node_modules` folder, to get the installed version?
Or maybe, calling a command like `npm list eslint@^10.2.0` to check if the required version is installed?

If you want, I can create a new PR to specifically work on a better solution to use in place of `version.replaceAll(/[^\d|.]/g, '')`, and we stall merging this PR until the other one is merged first?

jycouet: I totally get you, haha.
We can't really look at lock files or node_modules as in the create scenario, it will exist only later. And in the add you are not sure it's there... It's probably there.

In `packages/sv-utils/src/common.ts` we already have some fragile version check. I was hesitant to add semver pkg, but it's not ESM, and it like 100kb! So I'll keep the good enough for now.

I was suggesting to switch mainly because you added a test 😅
But yes, it's kinda internal `sv`

packages/sv/src/addons/prettier.ts
jycouet: Yeah let's do this one 👍

packages/sv/src/core/tests/common.ts
jonasgeiler: If we went with the solution of using [npm's `semver` library](https://github.com/npm/node-semver), [as proposed in the comment above](https://github.com/sveltejs/cli/pull/1069#discussion_r3155636741), the `semver.minVersion(...)` function would throw an error for `latest` and `workspace:*` and such.

Nonetheless: Yeah I think it would be good to throw in those cases. Otherwise the caller just gets an empty string returned.

packages/sv/src/core/common.ts
jycouet: I'll try quickly with semver to see if esm is an issue... And how much we can treeshack

jonasgeiler: We can also extract just the required, minimal code for the `minVersion(...)` function and provide THAT as a `sv-utils` function. I see a lot of things that can be stripped away, like the debug features. But all that would slowly wander into bonkers territory 😅

jonasgeiler: Actually I'll quickly try that. Extracting and trimming down everything that is needed for `minVersion(...)` and put that in a Gist. Let's see...

jycouet: I'm running the experiment already! :p
And it's very tree shakable.

I'll push here to see in a bit, ok ?

jonasgeiler: Yeah go ahead!
I have tried doing the above and it turns into like 1250 LoC... So probably better to just outright use the `semver` library if possible.

Still, I created a Gist here: https://gist.github.com/jonasgeiler/a480c9637a0e96c80050b4424536da8a
But I guess I probably wasted my time on this 😅 Had fun though!

packages/sv/src/addons/prettier.ts
```diff
@@ -89,23 +87,24 @@ export default defineAddon({
...
 
-const SUPPORTED_ESLINT_VERSION = stripVersionRange(ESLINT_VERSION).split('.')[0];
-
-function hasEslint(version: string | undefined): boolean {
-	return !!version && version.startsWith(SUPPORTED_ESLINT_VERSION);
+function checkEslint(version: string | undefined): { installed: boolean; supported: boolean } {
+	if (!version) return { installed: false, supported: false };
+	const supportedMajor = coerceVersion(ESLINT_VERSION).major;
+	const installedMajor = coerceVersion(version).major;
+	return { installed: true, supported: installedMajor === supportedMajor };
```
jonasgeiler, line 109: I didn't even think about using `semver.coerce(...)`. This is much cleaner! :D

jonasgeiler (commented):
@jycouet Looking much better now! Nice work :)

packages/sv/src/core/common.ts
jycouet: That's a lot of lines!!!
Not _wasted_, I hope you had fun! :)

Hoooo, just saw that you are in Vienna?! let's have a coffee! :)

jycouet:
Thank you, you! :)

jycouet (approved):

hyunbinseo:
So `eslint-config-prettier` being removed in `sv@0.13.0` was actually a bug?

- https://github.com/sveltejs/cli/releases/tag/sv%400.13.0
- https://github.com/hyunbinseo/svelte-kit-templates/commit/1d5dbb1a19aed58bb1f0ca086247c3f475c3e772

jycouet:
Probably due to this missing `9` to `10`.

But it's here now: https://github.com/sveltejs/cli/blob/main/packages/sv/src/cli/tests/snapshots/create-with-all-addons/package.json#L42

PR #1070, opened 2026-04-28
test(cli): using `pnpm`

jonasgeiler:
### Description

While working on #1069, I have noticed that the `pnpm test:ui --project cli` command fails with the following message, when you have [Node.js Corepack](https://github.com/nodejs/corepack) installed and enabled for `pnpm`:
```
 FAIL   cli  tests/cli.ts:46:19 > cli > should create a new project with name '@my-org/sv'
AssertionError: Error addon test: 'i' -> {
  "stderr": "This project is configured to use pnpm because /home/jonas/repos/_contrib/cli/package.json has a \"packageManager\" field\n",
  "stdout": "",
  "exitCode": 1
}: expected 1 to be +0 // Object.is equality

- Expected
+ Received

- 0
+ 1

 ❯ tests/cli.ts:167:8
    165|       res.exitCode,
    166|       `Error addon test: '${cmd}' -> ${JSON.stringify(res, null, 2)}`
    167|      ).toBe(0);
       |        ^
    168|     }
    169|    }
```

~~Since I can't find a good reason to use `npm` here, I just changed the commands to use `pnpm`, which fixes the problem and prevents the test from failing. A very simple PR.~~
__Edit:__ To fix this, I have now opted to instead just disable Corepack strict-mode by passing an `COREPACK_ENABLE_STRICT=0` environment variable when running commands with npm.

### Checklist

- [X] Update [snapshots](../CONTRIBUTING.md#update-snapshots) (if applicable)
  > Note: I hope I did it correctly, since there were no changes shown in git?
- [X] Add a [changeset](../CONTRIBUTING.md#generating-changelogs) (if applicable)
- [X] Allow maintainers to edit this PR
- [X] I care about what I'm doing, no matter the tool I use (Notepad, Sublime, VSCode, AI...)

### DIFF
--- a/packages/sv/src/cli/tests/cli.ts
+++ b/packages/sv/src/cli/tests/cli.ts
@@ -158,16 +158,8 @@ describe('cli', () => {
 					['run', 'test']
 				];
 				for (const cmd of cmds) {
-					const res = await exec('npm', cmd, {
-						nodeOptions: {
-							stdio: 'pipe',
-							cwd: testOutputPath,
-							env: {
-								...process.env,
-								// Disable warning about npm instead of pnpm
-								COREPACK_ENABLE_STRICT: '0'
-							}
-						}
+					const res = await exec('pnpm', cmd, {
+						nodeOptions: { stdio: 'pipe', cwd: testOutputPath }
 					});
 					expect(
 						res.exitCode,
--- a/packages/sv/src/cli/tests/cli.ts
+++ b/packages/sv/src/cli/tests/cli.ts
@@ -158,8 +158,16 @@ describe('cli', () => {
 					['run', 'test']
 				];
 				for (const cmd of cmds) {
-					const res = await exec('pnpm', cmd, {
-						nodeOptions: { stdio: 'pipe', cwd: testOutputPath }
+					const res = await exec('npm', cmd, {
+						nodeOptions: {
+							stdio: 'pipe',
+							cwd: testOutputPath,
+							env: {
+								...process.env,
+								// Disable warning about npm instead of pnpm
+								COREPACK_ENABLE_STRICT: '0'
+							}
+						}
 					});
 					expect(
 						res.exitCode,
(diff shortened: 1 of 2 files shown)


jycouet:
What version of `pnpm` are you using?
I remember using `npm` here to avoid mixing monorepo lockfiles & generated lockfiles.

I'll have a look, thx

jonasgeiler:
@jycouet 
Corepack is using `pnpm v10.17.0`, as defined here:
https://github.com/sveltejs/cli/blob/92d7ca121efc85577521d4c92d816dd7ba94c59a/package.json#L45

However, I have now found and commited a different solution for this problem, which is much less intrusive.
I basically just disable the Corepack strict mode with the `COREPACK_ENABLE_STRICT=0` environment variable, leaving everything else as-is.
Let me know if you like this solution better!

---

One thing I have noticed while working up a better solution, is that the `pnpm-workspace.yaml` file contains an exclusion for `.test-tmp/**`, however I have not found any usages of `.test-tmp` anywhere in the code base?
https://github.com/sveltejs/cli/blob/92d7ca121efc85577521d4c92d816dd7ba94c59a/pnpm-workspace.yaml#L3
Is it possible this was intended to exclude the `.test-output/` folders? If so, the `pnpm-workspace.yaml` file should probably look more like this:
```diff
packages:
  - 'packages/*'
-  - '!.test-tmp/**'
+  - '!**/.test-output/**'

onlyBuiltDependencies:
  - esbuild
```
As documented here: https://pnpm.io/pnpm-workspace_yaml

I'll leave that up to you to change though.

jycouet:
I'm using `.test-tmp` to test from time to time the cli locally, like "manual checks". _Maybe we could add it to the contributing guide as info?_

But adding `!.test-output` & switch to `pnpm` could improve the pipeline time... Let's see!

Like this we don't need the corepack tweak? Correct? @jonasgeiler

jonasgeiler:
> @jycouet wrote: Like this we don't need the corepack tweak? Correct?

Yup, that was the originally proposed solution.
I changed it to use the `COREPACK_ENABLE_STRICT=0` because you mentioned the need to "avoid mixing monorepo lockfiles & generated lockfiles", which I am not sure if it will happen now? But if it works on both of our setup, I guess it should be fine?

jycouet (approved):
Wanted to be sure, thx 🚀

PR #1072, opened 2026-04-28
fix(playwright): move `playwright install` from `prepare` to `test:e2e`

### LINKED ISSUE #1062: `playwright install` is added to `prepare` script on unsupported systems.
### Problem

When I create a new sveltekit application with playwright, or install any library afterwards an error flashes but it disappears immediately leaving me confused. 

<details>
<summary>image</summary>

Image

</details>

### The cause

When you run `playwright install` on an unsupported system (like Arch) it fails. `playwright install` is added to `prepare` script which runs every time you install a new library. 

It was implemented recently to address this issue https://github.com/sveltejs/cli/issues/1038 

@Rich-Harris wrote
> I just set up a project with Playwright, tried pnpm test and hit a giant wall of console spam. The pnpm exec playwright install instruction was buried in the middle of a bunch of other junk — if I didn't know better I could well have missed it. Made me think we should just do this automatically when installing the Playwright add-on?

But after bootstrapping a SvelteKit application there's an instruction that says "install playwright bro"

<details>
<summary>image</summary>

Image

</details>

If you use Storybook addon it installs browsers for you if you choose so. 

### The solution

Remove `playwright install` from the prepare script and either:
1) make "Run npx playwright install to download browsers" text bigger so it's easier to spot the instruction
2) ask if user wants to install playwright as an additional installation step (similar to when Storybook addon is used - and remove playwright installation from storybook step)

OR
Conditionally add `playwright install` to `prepare` script on systems that support it. 

OR
Add an instruction to remove `playwright install` from prepare script on unsupported systems to avoid confusing errors.

jycouet:
Closes #1062

### Description

Move `playwright install` out of the `prepare` script and into `test:e2e`.

- `prepare` runs on every `npm install`, including in prod / CI environments that probably don't run e2e tests. No reason to download browsers there.
- It also fails noisily on unsupported systems (Arch, NixOS, ...) every time you add a dep.
- Moving it to `test:e2e` ties the install to the only place it's actually needed, without annoying the test launch (playwright skips when browsers are already cached).

### Checklist

- [x] Update [snapshots](../CONTRIBUTING.md#update-snapshots) (if applicable)
- [x] Add a [changeset](../CONTRIBUTING.md#generating-changelogs) (if applicable)


paoloricciuti:
What happens if the browsers are already installed, though? How fast it skips?

jycouet:
It takes less than a second

PR #1110, opened 2026-06-02
chore: remove `{@const ...}` tags from template

Rich-Harris:
Replaces `{@const ...}` with declaration tags. Tests fail locally for me, not clear why, hoping they pass in CI


Conduitry (dismissed):
One issue is that we need to bump the devdep version of Svelte that's used in tests. And then also probably ESLint and Prettier plugins.

But also, if we're introducing new Svelte features into the demo apps, we should bump the ranges of the compiler, ESLint, and Prettier plugins that are set in package.json in the various templates.

Rich-Harris:
@jycouet are the test failures expected?

jycouet:
> @jycouet are the test failures expected?
   
Looking at it now its because of minimum age of packages. 
I think that we don't whan people to add in their conf non default age neither excludes. 

We should probably stick to defaults of 1440 https://pnpm.io/fr/settings#minimumreleaseage ? Today we have https://github.com/sveltejs/cli/blob/main/pnpm-workspace.yaml#L1 (2 days)

Should be soon green. (svelte `5.56.1` will be 48 hours old in 14h 😅)

Issue #969, opened 2026-03-03
migrate `better-auth` to `1.6`


sacrosanctic:
https://better-auth.com/blog/1-5

sacrosanctic:
1.5 seems a bit rough around the edges, we should wait a bit for them to squash some bugs first.

jycouet:
Too bad in a minor. 
Thx for the note

jycouet:
Even 1.6 now https://github.com/better-auth/better-auth/releases/tag/v1.6.0

PR #1115, opened 2026-06-04
chore: bump deps

manuel3108:
### Description

Let's hope that tests still pass...
Skipping `better-auth`, see #969 
Skipping `create-vite` because the all versions above `^8.2.0` have strange issues around font-sizes and it seems impossible to make all 8 test-tempaltes pass at the same time.

### Checklist


package.json
```diff
@@ -21,25 +21,25 @@
...
-		"eslint": "^10.4.0",
-		"eslint-plugin-svelte": "^3.17.0",
+		"eslint": "^10.4.1",
+		"eslint-plugin-svelte": "^3.19.0",
 		"magic-string": "^0.30.21",
-		"prettier": "^3.8.1",
+		"prettier": "^3.8.3",
 		"prettier-plugin-packagejson": "^2.5.22",
-		"prettier-plugin-svelte": "^3.4.1",
+		"prettier-plugin-svelte": "^3.5.2",
```
hyunbinseo: Shouldn't the root package be updated to `prettier-plugin-svelte@4` as well?

manuel3108: Good catch, i thought running `pnpm update -r` was enough, but the `--latest` flage also updates majors.

manuel3108:
I think the remaining ci failures are due to `minimumReleaseAge: 2880`. So let's wait two days and see if it's better.

manuel3108:
@jycouet This bumps our ci to `node@v22` to avoid problems with tsdown below `node@20`. Given that 22 is more than 2 years old and has recently become EOL, i think this should be fine. [source](https://nodejs.org/en/about/previous-releases)

Apart from that, we are finally error free.

PR #1116, opened 2026-06-04
feat(sv): add prisma as first class add-on

devharshthakur:
This adds Prisma as a first-class database ORM add-on in `sv`.

`sv` already provides Drizzle ORM as an official addon. In the TypeScript and Node.js ecosystem, Prisma is the other widely used choice, which is why I added this add-on.

This implementation supports:

- PostgreSQL
- SQLite

These are the most commonly used Prisma database options and provide a strong starting point for most projects.

## Why this matters

The goal is to make Prisma available through the same built-in add-on workflow that already exists for other core tooling in `sv`, so contributors can choose their preferred ORM without manual setup.

It is also inspired from `tanstack` cli. It provides both `drizzle` and `prisma` as first class addons

## Notes

- I tested the generated sveltekit projects in every format and everything is working as expected so far.
- MySQL is intentionally left out for now to keep the initial contribution focused.
- If there is community demand, I can add MySQL support in a follow-up.

### DIFF
--- a/documentation/docs/30-add-ons/07-prisma.md
+++ b/documentation/docs/30-add-ons/07-prisma.md
@@ -1,0 +1,55 @@
+---
+title: prisma
+---
+
+[Prisma](https://www.prisma.io/) is a another TypeScript ORM for Node.js applications, with a schema-driven workflow and generated client.
+
+## Usage
+
+```sh
+npx sv add prisma
+```
+
+## What you get
+
+- a Prisma setup for SvelteKit projects
+- a generated Prisma client wired into the app
+- a `prisma/schema.prisma` schema file
+- a `prisma.config.ts` or `prisma.config.js` config file
+- an `.env` file with a database connection string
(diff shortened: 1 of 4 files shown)


jycouet:
That would be a perfect [community add-on](https://svelte.dev/docs/cli/community)

devharshthakur:
@jycouet `sv` cli has first class support for `dirzzle`, and `prisma` is only other most used alternative, any future plans for adding this or it is left as a community add on permanently

jycouet:
The team selected some packages that are widely adopted to be officially supported _(this can change, but right now we have this list)_.

That's why we created the [community add-ons](https://svelte.dev/docs/cli/community) feature to support more tools and integrations. (and your code is a good base to switch to a community one 🎉)

_Side note: I personnaly don't use Drizzle nor Prisma, and I plan to create another community add-on to support something else 🤫_

PR #1117, opened 2026-06-05
feat(eslint): support loading config from `vite.config.*`

manuel3108:
Relates #1111 
(let's try and create multiple small prs)

### Description

Sources:
* https://github.com/sveltejs/eslint-plugin-svelte/issues/1550
* https://github.com/sveltejs/language-tools/pull/3031
* https://github.com/sveltejs/kit/pull/15944

### Checklist

- Update [snapshots](../CONTRIBUTING.md#update-snapshots) (if applicable)
- Add a [changeset](../CONTRIBUTING.md#generating-changelogs) (if applicable)
- Allow maintainers to edit this PR
- I care about what I'm doing, no matter the tool I use (Notepad, Sublime, VSCode, AI...)

Issue #1111, opened 2026-06-04
prepare for kit@3


manuel3108:
This is a non exhaustive list that we will need to add and remove stuff as we go.

## `sv create`
* [x] Templates should use kit@3
* [x] get rid of `svelte.config.js` in favor of `vite.config.js`
* [x] ~overhaul `demo` template to use remote functions if relevant, see https://github.com/sveltejs/cli/issues/641~

## `sv migrate`
* [x] migrate package.json and stuff like that to kit@3
* [x] get rid of `svelte.config.js` in favor of `vite.config.js`
* [x] allow migrating single files, in example #1103 
* [x] make `sv migrate` the single source of truth for migrations, as vscode migration in example do not apply everything that was part of the `sv migrate svelte-5` thingy, see https://github.com/sveltejs/cli/issues/1104

## `sv add`
* [x] check compatibility of relevant official add-ons
* [x] get rid of `$env/*` imports in `drizzle` and `better-auth`

bnn1:
hey not sure if i should open a new issue or it's ok to report there, but anyway since there's no `svetle.config.js` alises are now defined in `vite.config.js` but the latest `sv create` command doesn't put them there so depending on selected configuration `sv create` may produce broken code because `import ... from '$lib'` would not be resolved

jycouet:
> hey not sure if i should open a new issue or it's ok to report there, but anyway since there's no `svetle.config.js` alises are now defined in `vite.config.js` but the latest `sv create` command doesn't put them there so depending on selected configuration `sv create` may produce broken code because `import ... from '$lib'` would not be resolved

In `svetle.config.js`, the lib alias was not needed but available in your code. Same for vite config.
I just checked latest sv and you can see it in `.svelte-kit/tsconfig.json`.
Do you have a more specific case to share?

bnn1:
@jycouet 
1. `bunx sv@latest create test-project`
2. select all plugins/addons/libraries during installation
3. after bootstrap finished find:
  - 'The `config.experimental.handleRenderingErrors` option has been removed. Please see the list of breaking changes for your major release' error when running `svelte-kit sync` error - fixed by removing `handleRenderingErrors` from `vite.config.js`
  - 'Error: `defineEnvVars` has moved — import it from `@sveltejs/kit/env` instead' error - fixed by changing import path in `env.ts` file
[...]

PR #1119, opened 2026-06-05
feat(create): provide svelte config in `vite.config.js` during create

manuel3108:
Relates #1111 
Discards: https://github.com/sveltejs/cli/pull/1117
(let's try and create multiple small prs)

### Description

I absolutely do not understand why we had so many `svelte.config.js`'s flying around, especially since they were all close to identical. Therefore removed most of them. Since everything in the `vite.config` is js based, there is no need to bother about `.ts` extensions during create.

---

JYC update: 
- even if `vite.config.ts` has no typescript today, when I select ts, I expect to have it in `ts` (and it's today's behavior). Let's keep it here.
- added a util for svelteConfig & updated all addons

Todos:
- [x] playground test reading vite config, needs `@sveltejs/load-config`
- [x] snapshots
- [x] changeset

Sources:
* https://github.com/sveltejs/eslint-plugin-svelte/issues/1550
* https://github.com/sveltejs/language-tools/pull/3031
* https://github.com/sveltejs/kit/pull/15944

### Checklist


documentation/docs/50-api/20-sv-utils.md
```diff
@@ -232,44 +232,42 @@ Namespaced helpers for AST manipulation:
 
 ## Svelte config
 
-A SvelteKit project's config can live in two places: the default export of a `svelte.config.{js,ts}`, or the object passed to `sveltekit()` in a `vite.config.{js,ts}`. `svelteConf` edits it wherever it is, so add-ons don't have to care which.
+As of SvelteKit 2.62, the svelte/kit config can be passed straight to the `sveltekit()` plugin in `vite.config.{js,ts}`, and a separate `svelte.config.{js,ts}` is no longer required. **This is now the default for projects created by `sv`** - generated projects keep their config inside `vite.config.js` and ship no `svelte.config.js`.
```
manuel3108: im not sure we should point out that it's no longer required. That documentation won't age well i think

documentation/docs/50-api/20-sv-utils.md
```diff
@@ -232,44 +232,42 @@ Namespaced helpers for AST manipulation:
...
 
-### `svelteConf`
+`svelteConfig` lets add-ons read and edit that config wherever it lives - the `sveltekit()` argument in `vite.config.{js,ts}` (the new default), or a `svelte.config.{js,ts}` default export (still supported) - without having to know which.
```
manuel3108: `(the new default)` same here

packages/sv-utils/src/svelte-config.ts
```diff
@@ -0,0 +1,240 @@
...
+ * svelteConfig.edit({ sv, cwd }, ({ override, js }) => {
+ *   js.imports.addDefault(ast, { from: '@sveltejs/adapter-node', as: 'adapter' });
+ *   override({ adapter: js.functions.createCall({ name: 'adapter', args: [], useIdentifiers: true }) });
+ * });
+ * ```
+ */
+function edit({ sv, cwd }: { sv: SvFileApi; cwd: string }, editFn: SvelteConfEdit): void {
+	const location = find(cwd) ?? {
+		path: 'svelte.config.js',
+		kind: 'svelte'
```
manuel3108: Shouldn't we fallback to a vite.config.js by default?

packages/sv/src/cli/tests/snapshots/create-with-all-addons/e2e/demo.test.ts
manuel3108, line 1: Im failing to understand why this file shouldnt be part of the snapshot anymore??? We didn't change anything in that regard, right?

manuel3108:
Ran some local tests, and everything is working perfectly fine, as soon as those (optional) comments are adressed, im fine by you merging this!

Also lets keep it minor as you suggested.

Can't approve since im the author

packages/sv/src/cli/tests/snapshots/create-with-all-addons/e2e/demo.test.ts
jycouet: It's an outdated file that shuold have been removed when we changed vitest files in https://github.com/sveltejs/cli/pull/960

hyunbinseo:
Does `eslint.config.js` file no longer need `svelteConfig` imported from `svelte.config.js`?

The import is removed but the following issue is not closed yet:

- https://github.com/sveltejs/eslint-plugin-svelte/issues/1550
- https://github.com/hyunbinseo/svelte-kit-templates/commit/5193ee58286c9031e2351e2ba53a017d83a65419 - see diff here or in this PR's snapshot

Would love a clearer instruction for those who want to upgrade.

e.g. the `svelteConfig` is no longer needed if peer deps are upgraded to ~

jycouet:
The import isn't gone - it's conditional.

`sv` now stores config in `vite.config.js` and ships no `svelte.config.js`. The eslint addon only adds the `svelteConfig` import when a real `svelte.config.{js,ts}` exists. With config in `vite.config.js` there's nothing importable, so we omit it on purpose: `svelte-eslint-parser` falls back to its defaults.

Upgrading an existing project:
- Keep your `svelte.config.js`? Nothing to do - the import stays wired.
- Moved config into `vite.config.js` and deleted `svelte.config.js`? Drop the `import svelteConfig ...` line and the `svelteConfig` entry in `parserOptions`. The parser defaults take over.

hyunbinseo:
> explicitly importing allows for better compatibilty and functionality with rules and other tooling that depend on the config file. [source](https://github.com/sveltejs/eslint-plugin-svelte)

If I understood correctly:

1. The explicit import is for better compatibility
2. Svelte config defined in `vite.confit.{js,ts}` cannot be imported
3. So the default `eslint-plugin-svelte` options are used for now
4. If the following issue is resolved, explicit importing can be done, changing the status of 2

- https://github.com/sveltejs/eslint-plugin-svelte/issues/1550

PR #1122, opened 2026-06-08
feat: support kit's explicit environment variables

jycouet:
### Description

- [x] `drizzle` and `better-auth` now emit kit's explicit environment variables (`src/env.ts` + `$app/env/private`) when the project uses them in kit 3 / `next`, or kit 2 with `explicitEnvironmentVariables`. And fall back to legacy `$env/dynamic/private` otherwise. 
- [x] Adds a `defineEnv` helper to `sv-utils`.
- [x] diff is big because of a new snapshot testing... that covers envs

The new util will be reuse for a future `sv migrate env` ! 

### Checklist


Rich-Harris (approved):
nice!

PR #1138, opened 2026-06-13
feat(migrate): rework migrations and initial `sveltekit-3` migration

### LINKED ISSUE #1124: `sv migrate`
Relates #1111 

Given that we will be re-using the migration script for kit@3 we want to overhaul it beforehand. The current setup is mainly regex-based and mostly uses `magic-string` to do replacements. This has proven to be error prone.

Based on the setup we build for `sv` and `sv-utils` we can do better. We should be re-using all of the well proven and tested ast utils we have accumulated over time. and for printing we will use sveltes new `print` functionality and `esrap` for non svelte files.

Based on the learnings from the previous migrations, the following things should be possible:
* migrete files or subdirs one by one (flexible cmd args)
* allow selecting different migrations to run
* allow more flexibility in what the migration executes (you might not want to move all features immediately to seperate commits and stuff like that)

The code could look something like this, but this only a really early, un-tested and un-proven idea.

<details><summary>WIP, very early idea</summary>
<p>

```js
export default defineMigration({
    id: 'kit-3',
    changelog: 'https://svelte.dev/blog/whatever'
    run: () => {
        // ask the users more questions if required

        return  {
            pkg: ...pkgTask,
            config: ...moveConfigToViteTask,
            env: ...coolNewEnvTask
        }
    }
});


function pkgTask({sv, files}) {
    // ask more questions for the migration task

    return defineMigrationTask({
        name: 'Update package.json',
        isRequired: true,
        isApplied: () => {
            const isApplied = false;
            sv.file(
                files.match('package.json'),
                transform.json({data} => {
                    const version = getDependencyVersion(data, '@sveltejs/kit');
                    isApplied = version && semver.gte(version, '3.0.0'); // pseudo

                    // no return, so not modifying the file
                })
            )
        },
        run: () => {            
            sv.file(
                files.match('package.json'),
                transform.json({data} => {
                    data.dependencies['@sveltejs/kit'] = '^3.0.0';
                    return data;
                })
            )
        }
    })
}
```

</p>
</details>

manuel3108:
Relates #1111
Closes #1124 

### Description

This PR lays the groundwork for making `sv migrate` the new home for Svelte migrations.

It does a few things:

* bumps `svelte-migrate` to a new major version so it can be deprecated in favor of `sv migrate`
* rewrites the migration system to re-use the tooling and utilities we have built for `sv add` over the last two years
* removes previous migrations from the source code
* keeps previous migrations available to users as legacy migrations by delegating to `npx svelte-migrate@1`
* introduces a new migration system that is closely aligned with the existing `sv add` architecture

### Migration system

The new system is built around two concepts: migrations and migration tasks.

A migration is the top-level wrapper. It is similar to what we had before and will usually be added alongside a new version release, though not necessarily only for major releases.

A migration task is the actual execution unit responsible for applying part of a migration. Larger migrations can be split into multiple tasks. For example, `sveltekit-3` currently consists of separate tasks for `package.json`, `svelte.config.js`, and environment variable changes.

Tasks can be required or optional. Optional tasks can be selected or deselected by the user, which has a few important benefits:

* users can migrate step by step
* users can split migration changes into multiple commits more easily
* each task can represent a clear testing surface

Each task receives almost the same parameters as an addon, with only minor differences. This allows migrations to re-use addon functionality directly. For example, the `sveltekit-3` migration calls the `experimental` addon to update `package.json` and the relevant peer dependencies.

### Multi-file migrations

To support migrations that need to edit multiple files, this PR introduces `sv.files` in addition to `sv.file`:

```js
sv.files(
	{
		// glob expression evaluated with fs.globSync by sv
		include: '**/*.svelte',

		// early string-based content filter to avoid running expensive edit functions for unrelated files
		where: (content) => content.includes('$env/')
	},
	transforms.svelteScript({ language }, ({ ast }) => {
		return changeImports(ast.instance.content);
	})
);
```

### Migrating

```bash
# basics (will prompt everything that is not explicitely provided)
pnpx sv migrate sveltekit-3

# with path
pnpx sv migrate sveltekit-3 --cwd ../your-work-dir 

# only required tasks (like package.json)
pnpx sv migrate sveltekit-3 --tasks required

# only required and selected optional task
pnpx sv migrate sveltekit-3 --tasks env-vars

# no prompt workflow
pnpx sv migrate sveltekit-3 --cwd ../your-work-dir --tasks all --confirm --install pnpm

```

### Pending work

* [x] Decide whether we want additional verification steps before running migrations
  * [x] ~~lint check~~ We decided against this for now
  * [x] ~~`svelte-check`~~ We decided against this for now
* [x] Add docs and readmes
  * [x] `sv.files`
  * [x] `sv migrate`
* [x] Implement the `env-vars` migration
* [x] ~~Fix and bump `esrap`~~
  * ~~It currently generates a lot of new lines where the previous content did not have them.~~
  * ~~This may be difficult and could potentially require a major bump for `esrap`.~~
  * This is not possible in `esrap` as it takes the `ast` as an argument and is unaware of the input. Therefore implemented based on a text [diff](https://npmx.dev/package/diff) library that does the job pretty good `preserveOriginalNewlines`
* [x] Add tests for migrations
  * This should be fairly straightforward once a few migration testing helpers are in place.
* [x] Add more command-line arguments
  * Currently not all prompts can be skipped through CLI arguments.
* [x] allow users to pass a glob expression / directory / file and only migrate this one
* [x] migration for `$app/environment` to `$app/env`
* [x] ~~check migration for app-state (legacy) and see if we can use that one or if we should rewrite it~~ Should be fine, migration was published over 1.5 years ago
* [ ] ~~expose an api for migrations, so that we can use it from the playground, vs code extension and whatever~~ will be implemented at a later point

### Checklist


manuel3108:
Ok, just tried this out on a couple of real-world apps. Overall the technology and idea seems to work really well. No parsing errors whatsoever

There is one major showstopper though: formatting. As soon as we touch and moidify a file and then print it through esrap a lot of the formatting is totally gone which leads to massive diffs. My fix-attempt with `preserveOriginalNewlines` is only working for new lines and has bugs. Not sure if / how this is fixable in `esrap`, will need to investigate.

Also formatting of a handful of files took forever after migrating. We will need to check this out as well.

.changeset/wise-beers-switch.md
```diff
@@ -0,0 +1,5 @@
+---
+'svelte-migrate': major
+---
+
+breaking: the `svelte-migrate` package has been replaced by the `sv` package
```
jycouet, line 5: - [x] Should it be a major to say "please don't use it anymore" ?
- [x] Should we add a log as the start of the bumped `svelte-migrate` to say to use `sv` _(EDIT: haha, it's already done 👍)_
- [ ] We should not forget to deprecate in npm

documentation/docs/20-commands/40-sv-migrate.md
```diff
@@ -2,7 +2,7 @@
 title: sv migrate
 ---
 
-`sv migrate` migrates Svelte(Kit) codebases. It delegates to the [`svelte-migrate`](https://www.npmjs.com/package/svelte-migrate) package.
+`sv migrate` migrates Svelte(Kit) codebases. Some migrations are built into `sv`; older migrations are run through [`svelte-migrate`](https://www.npmjs.com/package/svelte-migrate).
```
jycouet: we should pin the version that we will use?

documentation/docs/20-commands/40-sv-migrate.md
```diff
@@ -46,3 +55,43 @@ Upgrades a library using `@sveltejs/package` version 1 to version 2. See the [pu
...
+
+Run the migration in a different working directory.
+
+### `--no-git-check`
+
+By default, `sv migrate` prompts before making changes when the git working tree is dirty. Use `--no-git-check` to skip this check.
+
+### `--files <glob>`
+
+Limit the files considered for migration to those matching a glob. A migration may still try to edit or create files outside the glob, in which case `sv migrate` will show a warning.
```
jycouet, line 71: Maybe we can give a few glob examples ?

`--files` vs `--include` ? (maybe `--files` is good ^^)

packages/sv-utils/src/tests/transforms.ts
```diff
@@ -8,12 +8,14 @@ describe('transforms', () => {
 			const result = transforms.json(({ data }) => {
 				data.name = 'new';
 			})(input);
+			expect(result).not.toBe(false);
+			if (result === false) return;
 			expect(JSON.parse(result)).toEqual({ name: 'new' });
 		});
 
-		it('abort: returns original content on false', () => {
+		it('abort: returns fals if transform is cacelled', () => {
```
jycouet:
```suggestion
		it('abort: returns false if transform is cacelled', () => {
```

packages/sv-utils/src/files.ts
```diff
@@ -43,7 +43,19 @@ export function loadFile(cwd: string, filePath: string): string {
  * Writes the file. Will make parent directories as needed.
  * @param filePath - Resolves paths relative to the workspace.
  */
-export function saveFile(cwd: string, filePath: string, content: string): void {
+export function saveFile(
+	cwd: string,
+	filePath: string,
+	content: string,
+	saveFileInfix?: string
```
jycouet, line 51: What is the need for this? 👀

packages/sv/src/core/package-manager.ts
```diff
@@ -119,5 +119,5 @@ export function addPnpmAllowBuilds(
 	const filePath = found ?? path.join(cwd, 'pnpm-workspace.yaml');
 	const content = found ? fs.readFileSync(found, 'utf-8') : '';
 	const newContent = pnpm.allowBuilds(...packages)(content);
-	if (newContent !== content) fs.writeFileSync(filePath, newContent, 'utf-8');
+	if (newContent && newContent !== content) fs.writeFileSync(filePath, newContent, 'utf-8');
```
jycouet, line 122: aybe not in this PR, but each tim eI see `fs.writeFileSync` I'm thinking that we should use our `saveFile`. no?

packages/sv/src/migrate/migrations/sveltekit-3/tasks/package-json.ts
```diff
@@ -0,0 +1,16 @@
...
+	run: (args) => {
+		// instead of duplicating the logic of the experimental addon, lets just call it.
+		// migrates kit to it's 'next' version.
+		experimentalAddon.run({
+			...args,
+			options: { versions: ['kit'], features: [] },
+			cancel: () => {}
+		});
+	}
+});
```
jycouet, line 16: I'm not sure to get the comment. 
Actually, now looking at `experimental.ts` I get it... But it's kinda cryptic from here.
Now `kit` is -> `@sveltejs/kit@next`, but tomorrow... how will it evolve in addon? in migrate?

Maybe the option should have been `kit-3-next` ? (we can still do it probably)

Let's think 🤗

packages/sv/src/migrate/migrations/sveltekit-3/tasks/svelte-config.ts
```diff
@@ -0,0 +1,47 @@
...
+				(node) => node.type === 'ImportDeclaration'
+			);
+			ast.body.unshift(...originalImports);
+
+			// kit-level options are flattened into the root of the vite config, so merge the properties
+			// and then move the properties into the new config object, skipping the unflattened `kit` wrapper
+			const newConfigProperties = [
+				...originalConfigObject.config.properties,
+				...originalConfigObject.kit.properties
+			];
```
jycouet, line 65: This should be handled directly in `svelteConfig.edit`. To be checked

jycouet (commented):
I did a first pass on Github mainly. Great job!!! 🎉

For the formating question... Maybe it should be in "next step" ? Or we could have a task "should we run pnpm format" at the end? (We can detect a format script?

In a lot of cases, there is a lint & format script... We could propose to check before and run after? and if not there, we put in next step or something?

manuel3108:
> For the formating question... Maybe it should be in "next step" ? Or we could have a task "should we run pnpm format" at the end? (We can detect a format script?

We are already doing this. But this still leads to wild results sometimes

.changeset/wise-beers-switch.md
manuel3108: > Should it be a major to say "please don't use it anymore" ?

Wdym? That's exactly the reason i made this a major.

documentation/docs/20-commands/40-sv-migrate.md
manuel3108: That is done, it always uses `svelte-migrate@1`

documentation/docs/20-commands/40-sv-migrate.md
manuel3108: Haha, I had a long discussion with Codex about this, but this is the best that came up. Include was also what was proposed but since it's a glob you can technically also use it to exclude stuff (no idea why you should do that btw)

packages/sv-utils/src/files.ts
manuel3108: For snapshot testing, so that it does not overwrite the files that its reading, otherwise you will have a mess in the diffs and snapshots

packages/sv/src/core/package-manager.ts
manuel3108: Agree. But since that one is deprecated I either way I wouldnt bother

packages/sv/src/migrate/migrations/sveltekit-3/tasks/package-json.ts
manuel3108: Yeah definitely. I just wanted to avoid duplicating the work you already did. (Btw this addon is the reason i started using exactly the same api, and I think it turned out great)

packages/sv/src/migrate/migrations/sveltekit-3/tasks/svelte-config.ts
manuel3108: I remember there was a reason, but i dont remember it.....

.changeset/wise-beers-switch.md
jycouet: I'm thinking, should it really be a major? Or minor is enough?

packages/sv/src/migrate/migrations/sveltekit-3/tasks/package-json.ts
jycouet: So we switch to `kit-3-next` and we are good?

jycouet:
```ts
┌  Welcome to the Svelte CLI! (v0.16.1)
│
◇  Select a migration to run
│  sveltekit-3
│
▲  Make sure to read the changelog for this migration before running it: https://svelte.dev/blog/whatever
│
■  Error: @sveltejs/kit is not listed as a devDependency in package.json
│      at Object.setup (file:///home/jycouet/udev/gh/yved/sveltejs_cli/pr-1058/packages/sv/dist/bin.mjs:372:53)
│      at determineTasks (file:///home/jycouet/udev/gh/yved/sveltejs_cli/pr-1058/packages/sv/dist/bin.mjs:458:12)
│      at file:///home/jycouet/udev/gh/yved/sveltejs_cli/pr-1058/packages/sv/dist/bin.mjs:426:23
│      at process.processTicksAndRejections (node:internal/process/task_queues:103:5)
│      at async runCommand (file:///home/jycouet/udev/gh/yved/sveltejs_cli/pr-1058/packages/sv/dist/engine-DpLfNKm_.mjs:1457:3)
```

In this case, should we:
- have an error without stacktrace?
- prompt to set the `cwd` ?
- suggest to set the --cwd ?

.changeset/wise-beers-switch.md
manuel3108: We did major when we deprecated [create-svelte](https://npmx.dev/package/create-svelte/versions)

packages/sv/src/migrate/migrations/sveltekit-3/tasks/package-json.ts
manuel3108: yeah that should work

manuel3108:
Mhmm good catch. Not sure why this is happening.
Also how did you even manage to get there? did you run this on a svelte-only project? Because there should be another cleaner check before that epects either `svelte` or `@sveltejs/kit` installed

.changeset/wise-beers-switch.md
jycouet: You are right, it's way better! :)

jycouet:
> Mhmm good catch. Not sure why this is happening. Also how did you even manage to get there? did you run this on a svelte-only project? Because there should be another cleaner check before that epects either `svelte` or `@sveltejs/kit` installed

I ran it directly in this repo 🤣
```ts
pnpm build && node ./packages/sv/dist/bin.mjs migrate
```

packages/sv/src/migrate/migrations/sveltekit-3/index.ts
```diff
@@ -0,0 +1,29 @@
+import { coerceVersion } from '@sveltejs/sv-utils';
+import { defineMigration } from '../../index.ts';
+import appEnv from './tasks/app-env.ts';
+import envVars from './tasks/env-vars.ts';
+import packageJson from './tasks/package-json.ts';
+import svelteConfig from './tasks/svelte-config.ts';
+
+export default defineMigration({
+	id: 'sveltekit-3',
+	changelog: 'https://svelte.dev/blog/whatever',
```
jycouet, line 10: To remember

packages/sv/src/migrate/migrations/sveltekit-3/tasks/environment/explicit-env-vars.ts
```diff
@@ -0,0 +1,410 @@
+import { Walker, js, type AstTypes, type Comments, type SvelteAst } from '@sveltejs/sv-utils';
```
jycouet, line 1: We should manage `const { ORIGIN } = await import('$env/static/private')`

packages/sv/src/addons/experimental.ts
```diff
@@ -24,8 +24,8 @@ const options = defineAddonOptions()
 	.add('versions', {
 		question: 'Which packages should use their `next` (pre-release) version?',
 		type: 'multiselect',
-		default: ['kit'],
-		options: [{ value: 'kit', label: '@sveltejs/kit@next' }],
+		default: ['kit-3-next'],
```
manuel3108, line 27: since renaming this option is technically a breaking change we should at least add a changeset for it. (even if it's patch only)

packages/sv/src/migrate/migrations/sveltekit-3/tasks/svelte-config.ts
```diff
@@ -32,16 +35,78 @@ export default defineMigrationTask({
...
+			let trustsAllOrigins = false;
 
 			for (const prop of newConfigProperties) {
 				if (prop.type !== 'Property') continue;
 				if (prop.key.type !== 'Identifier') continue;
 				if (prop.key.name === 'kit') continue;
 
+				// `csrf: { checkOrigin: false }` is deprecated; the equivalent is now `trustedOrigins: ['*']`
+				if (prop.key.name === 'csrf' && isCheckOriginDisabled(prop.value)) {
+					keyedConfig['trustedOrigins'] = ['*'];
```
manuel3108: As far as i can tell this migrates the `trustedOriginins` key to to the root of the svelte-kit config, which is incorrect according to my local type definitions. Semms like it should still be part of the `csrf` object.

jycouet: done

documentation/docs/20-commands/40-sv-migrate.md
sacrosanctic: Can we specify a version number so users know which one to use?

documentation/docs/20-commands/40-sv-migrate.md
```diff
@@ -46,3 +55,43 @@ Upgrades a library using `@sveltejs/package` version 1 to version 2. See the [pu
...
+
+When used with `--cwd`, the glob is relative to the working directory.
+
+### `--tasks <task...>`
```
sacrosanctic, line 75: is this like `--tasks task1 task2 task3 ...`?

documentation/docs/50-api/10-sv.md
```diff
@@ -32,6 +32,15 @@ export default defineAddon({
 			return 'export const foo = true;';
 		});
 
+		// edit multiple existing files
+		sv.files(
+			{
+				include: 'src/**/*.{js,ts,svelte}',
```
sacrosanctic, line 38: mention exclude here

packages/sv/src/core/engine.ts
```diff
@@ -36,26 +40,36 @@ function alphabetizeRecord(obj: Record<string, string>) {
...
+	const pkgPath = filePaths.packageJson;
+	sv.file(
+		pkgPath,
+		transforms.json<Package>(({ content, data }) => {
+			if (!content) throw new Error(`Invalid workspace: missing '${pkgPath}'`);
+
+			let modified = false;
+			for (const dependency of dependencies) {
+				if (dependency.dev && !data.devDependencies) data.devDependencies = {};
+				if (!dependency.dev && !data.dependencies) data.dependencies = {};
```
sacrosanctic: these 2 lines seems redundant considering the next 2 two does the same thing

packages/sv/src/core/config.ts
```diff
@@ -20,7 +23,29 @@ export type SvApi = {
...
 	 */
-	file: (path: string, edit: (content: string) => string | false) => void;
+	file: (path: string, edit: FileEdit) => void;
+	/**
+	 * Edits matching files in the workspace.
+	 * The `include` and `exclude` patterns are glob patterns relative to the workspace root.
+	 * For each matching file, the `edit` callback is called with the file content,
+	 * and should return the new content (or `false` to abort editing that file).
+	 *
+	 * Note: always adds excludes for `node_modules` and dot-prefixed directories
```
sacrosanctic, line 33:
```suggestion
	 * Note: `node_modules` and dot-prefixed directories are always excluded.
```
Should this also mention `build` and `dist`? Maybe link to the source for the curious? Should there be an option to override this?

packages/sv/src/core/formatFiles.ts
```diff
@@ -1,22 +1,68 @@
...
+		}
+	}
+
+	const cmd = resolveCommand(options.packageManager, 'execute-local', [
+		'prettier',
+		'--write',
+		'--ignore-unknown',
+		...options.filesToFormat
+	])!;
+	await run(cmd.command, cmd.args, options.cwd, 'Formatting modified files');
```
sacrosanctic, line 40:
```suggestion
  if (options.strategy === 'files-only') {
    const cmd = resolveCommand(options.packageManager, 'execute-local', [
      'prettier',
      '--write',
      '--ignore-unknown',
      ...options.filesToFormat
    ])!;
    await run(cmd.command, cmd.args, options.cwd, 'Formatting modified files');
    return;
  }

  void options.strategy satisfies never;
```
The second `if` is not an early return, we can make each enum explicit. And future enums will have less git diffs.

documentation/docs/20-commands/40-sv-migrate.md
manuel3108: exactly!

packages/sv/src/core/engine.ts
manuel3108: Indeed, good catch. That was a leftover

packages/sv/src/core/config.ts
manuel3108: Mhmm yeah, maybe it should. On the other hand it wouldn't cause any problems, because your next build is just going to override it. But in `node_modules` it could cause problems as this is sometimes treated as the source.

packages/sv/src/core/formatFiles.ts
manuel3108: Typescript is not happy with that one
image

sacrosanctic: whoops, I missed that the previous check has a fall through logic. Ignore this.

documentation/docs/20-commands/40-sv-migrate.md
```diff
@@ -2,7 +2,7 @@
 title: sv migrate
 ---
 
-`sv migrate` migrates Svelte(Kit) codebases. It delegates to the [`svelte-migrate`](https://www.npmjs.com/package/svelte-migrate) package.
+`sv migrate` migrates Svelte(Kit) codebases. Some migrations are built into `sv`; older migrations are run through [`svelte-migrate@1`](https://www.npmjs.com/package/svelte-migrate).
```
sacrosanctic: I meant `sv migrate` vs `svelte-migrate`? How does a user decide when to use which?

packages/sv/src/core/engine.ts
```diff
@@ -36,26 +40,35 @@ function alphabetizeRecord(obj: Record<string, string>) {
...
+
+				const dependencies = dependency.dev ? data.devDependencies! : data.dependencies!;
+
+				if (!dependencies[dependency.pkg] || dependencies[dependency.pkg] !== dependency.version) {
+					modified = true;
+					dependencies[dependency.pkg] = dependency.version;
+				}
+			}
```
sacrosanctic:
```suggestion
			for (const { dev, pkg, version } of dependencies) {
				const dependency = dev ? 'devDependencies' : 'dependencies';
				data[dependency] ??= {};

				if (data[dependency][pkg] !== version) {
					modified = true;
					data[dependency][pkg] = version;
				}
			}
```

documentation/docs/20-commands/40-sv-migrate.md
manuel3108: Users should always use `sv migrate`. `sv migrate` will delegate to `svelte-migrate@1` if required.

manuel3108:
Updated to latest svelte. `.svelte` files migrated by the migration are now way better and do not totally mess up the indentation. There are still a few improvements that will need to be made inside `svelte`'s `print` functionality, especially when running against `immich`. But that's not something that should prevent us from moving forward with this.

manuel3108:
The failing tests works perfectly locally, but locally another is failing. Maybe @jycouet could you have a look at this and see if you are able to reproduce this?
```
pnpm vitest --project cli
```

Edit: Haaa, this is unrelated to this branch, see https://github.com/sveltejs/cli/pull/1178. Happy :tada: !

PR #1140, opened 2026-06-15
chore: skip CI when updating non-relevant files

teemingc:
We added this to the Kit repo a while back and it's useful for saving CI time https://github.com/sveltejs/kit/blob/main/.github/workflows/ci.yml


manuel3108 (approved):
LGTM

Issue #80, opened 2024-10-10
Convert to JavaScript


Rich-Harris:
The first thing I experience when I clone this repo and try to install dependencies is this:

<img width="907" alt="image" src="https://github.com/user-attachments/assets/1a4e4dd6-1fad-4181-90a4-212667a728fc">

IIUC what this is telling me is that I can't install anything because `pnpm` tries to link up the executables as part of the installation process, but it can't do that because the executable hasn't been built. I can't build the executable because I haven't installed dependencies.

I'm sure this is fixable and if I bash my head against a wall for a couple of minutes I'll figure it out, but: **we don't have to live this way**. Svelte and SvelteKit are written in JS and don't have a build step (beyond generating `.d.ts` files during prepublish, or bundling the compiler for the playground in Svelte's case), and aside from avoiding this sort of headache it's been a real productivity win — I don't need to mess around with file watchers during dev, I don't need to deal with sourcemap flakiness or any of that nonsense, I can just `pnpm link` and I'm good to go. I can even make changes to this package inside another project's `node_modules`, and once I've done whatever I needed to do the changes are sitting there waiting for me to commit them — no more editing transpiled files and then painstakingly reapplying them to the source they were transpiled from.

Not something we need to act on immediately, but I promise future us will thank us for making this change

PR #1142, opened 2026-06-19
chore: capture full type signatures in api-surface snapshots

jycouet:
### Description

Build each public `sv` entry as a single-entry DTS bundle so the api-surface snapshots inline every referenced declaration instead of emitting name-only re-exports from the shared engine chunk. Now tracks signatures for re-exported symbols (`defineAddon`, `add`, ...) and internal types like `OfficialAddons`.

---

Instead of a custom solution, I would like to switch to `tsnapi`. But we need first these 3 merged
- https://github.com/antfu/tsnapi/pull/28
- https://github.com/antfu/tsnapi/pull/29
- https://github.com/antfu/tsnapi/pull/30
So let's merge this first before the switch.

I'm also considering more and more https://github.com/sveltejs/cli/issues/80 to remove tsdown, and not need any `tsnapi` or surface check... as the file split/org would be sufficient! 

### Checklist

- [x] Update [snapshots](../CONTRIBUTING.md#update-snapshots) (if applicable)
- [ ] Add a [changeset](../CONTRIBUTING.md#generating-changelogs) (if applicable) - N/A, dev tooling only


manuel3108:
Agree, #80 is defintely something we should consider sooner than later. Since the current api surface is missing so much, lets go ahead with this as a temporary solution

manuel3108 (approved):

PR #1147, opened 2026-06-24
feat: migration-task comment management

jycouet:
Targets `feat/sv-migrate`.

- `@migration-task` comments now also work in `.svelte` files. The Svelte printer ignores the JS `Comments` registry, so we insert real `Comment` nodes into the fragment. Fixes the environment task silently dropping these comments.
- One `addMigrationTask(message, target)` API, where the target picks the placement: `{ comments, node }` for JS/TS, `{ fragment, anchor? }` for `.svelte`.
- Split `migrate/index.ts` into `index.ts` (the `defineMigration*` framework) + `migration-task.ts` (the comment API, paired with its spec).


manuel3108 (approved):
This is perfect!

PR #1148, opened 2026-06-24
feat: reborn app-state migration

jycouet:
Rebuilds the `$app/stores` -> `$app/state` migration as a proper AST task (was a regex legacy shell-out).

- Runnable standalone: `sv migrate app-state`
- Also required inside the `sveltekit-3` migration

Bug fixes over the legacy version:
- truthy `$navigating` (e.g. `{#if $navigating}`) -> `navigating.to`
- bail cases (`$:`, `derived(store)`, `getStores()`) now leave a `@migration-task` instead of silently skipping
- drop spurious TODO when `.to` is already accessed

Stacked on #1147 (uses its shared `addMigrationTask`); merge that first.

### DIFF
--- a/.changeset/format-strategy-script.md
+++ b/.changeset/format-strategy-script.md
@@ -1,0 +1,5 @@
+---
+'sv': patch
+---
+
+feat(migrate): defer formatting to the project's own `format`/`fmt` script when present, falling back to `prettier` on modified files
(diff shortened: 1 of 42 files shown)


manuel3108 (approved):
LGTM, even found some more places to migrate in immich, the previous migration did not find / bailed.

PR #1154, opened 2026-06-27
fix(sv-utils): dedupe re-homed comments at attach time

jycouet:


manuel3108 (changes requested):
I'm not seeing any change in the `pkg.svelte.dev` repo.
Running `pnpx --config.block-exotic-subdeps=false https://pkg.svelte.dev/sv/c/f6db281 migrate`

still changes `src/lib/github.ts` by duplicating comments.

Since the repo is not public, i wont share too much code, but i think this should make it reproducable.
```ts
export async function list_recent_pulls(
	opts: { 
		/**
		 * Octokit to authenticate the REST call. Pass an
		 * `octokit_for_user(...)` instance from a remote function so the
		 * request counts against the user's GitHub budget, not the
		 * App-installation budget.
		 */
		octokit: Octokit;
		owner: string;
		repo: string;
		per_page?: number;
		page?: number
	 }
): Promise<PullSummary[]> { /* ... */ }
```

jycouet:
Reproduced: the parser reports a comment twice when it backtracks over TS type annotations (JSDoc inside an inline object type param), so the duplicate entered at parse time - a different path than the attach-time dedupe. Fixed in 3904cce by deduping by source position at ingest, with a round-trip test for your snippet.

Retest with: `pnpx --config.block-exotic-subdeps=false https://pkg.svelte.dev/sv/c/3904cce migrate`

manuel3108:
Note to myself: 
The fix works, although im keen to understand what is causing the problem here. Maybe `@sveltejs/acorn-typescript` or something? Since that's also something in our control, this is probably something we should investigate.

jycouet:
Yes, with your merge in acorn, let's update `feat/sv-migrate` and close this one

manuel3108:
Just checked locally, adding the following to `pnpm-workspace.yml` makes the problem go away. So this is no longer a problem. Will need to update the dependency in svelte to make that happen.

```yml
overrides:
  '@sveltejs/acorn-typescript': 1.0.11
```

Closing this for now.

PR #1157, opened 2026-06-30
feat: add `valibot` addon

### LINKED ISSUE #1152: add an option for validation library
### Proposal 1
`sveltejs/cli` maintains the `valibot` addon and make it optional in `sv create`.

This makes the templates easier to generate.

### Proposal 2
`sveltejs/cli` maintains the `valibot` addon and make it optional in `sv create`.

### Additional Info
Add a note: Svelte supports any validation library that implements standard Schema.

sacrosanctic:
Closes #1152

### Description

I don't know what test to write.

Removed write guards for demo paths. We already warn on dirty and they're demos, not real code.

### Checklist

### DIFF
--- a/packages/sv/src/addons/valibot.ts
+++ b/packages/sv/src/addons/valibot.ts
@@ -1,0 +1,11 @@
+import { defineAddon } from '../core/config.ts';
+
+export default defineAddon({
+	id: 'valibot',
+	shortDescription: 'schema validation',
+	homepage: 'https://valibot.dev',
+	options: {},
+	run: ({ sv }) => {
+		sv.devDependency('valibot', '^1.4.1');
+	}
+});
--- a/.changeset/valibot-addon.md
+++ b/.changeset/valibot-addon.md
@@ -1,0 +1,5 @@
+---
+"sv": minor
+---
+
+feat: add `Valibot` addon
--- a/packages/sv/api-surface-testing.md
+++ b/packages/sv/api-surface-testing.md
@@ -17,6 +17,7 @@ type OfficialAddons = {
 	storybook: Addon<any>;
 	mcp: Addon<any>;
 	experimental: Addon<any>;
+	valibot: Addon<any>;
 };
 declare const officialAddons: OfficialAddons;
 type BooleanQuestion = {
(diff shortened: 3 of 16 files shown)


packages/sv/src/addons/valibot.ts
```diff
@@ -0,0 +1,11 @@
+import { defineAddon } from '../core/config.ts';
+
+export default defineAddon({
+	id: 'valibot',
+	shortDescription: 'schema validation',
+	homepage: 'https://valibot.dev',
+	options: {},
+	run: ({ sv }) => {
+		sv.devDependency('valibot', '^1.4.1');
```
manuel3108, line 9: Not sure if you saw my note in the ticket early enough. In the maintainers meeting it was brought up that the addon would basically be a one liner as displayed here and therefore making it an addon would be an overstatement.

The idea was to make that a question during create, since you will always be capable of adding on line to your `package.json` by hand. Honestly, IMHO I wouldn't even bother asking a question and just always add it. But maybe @jycouet has an opinion here as well.

Most of your PR is still valid though and can probably stay unchanged.

sacrosanctic: I did see your msg, but it seems I have misinterpreted it. 

> IMHO I wouldn't even bother asking a question and just always add it

I am also in favour of not asking. But that seems a bit too opinionated considering `standard-schema` allows for different validation libraries. Users shouldn't have to reverse what the CLI has done. So I made it optional.

manuel3108: Assuming we keep it this way we would need to ensure that `better-auth` runs after `valibot` by running `runsAfter('valibot')` in `better-auth`'s setup.

I'm still not convinced what is the right forward here:
1. optional addon (as of now)
2. question during create
3. no question during create

I do get your point about the different libs. However we have used `valibot` everywhere (docs, demos, etc). But yes, users might have other choices.

sacrosanctic: To avoid the rolling back code problem, my original idea was to provide a list of validators. But that added maintenance burden to the CLI team and opted for this instead.

Side note: It would be nice if validation libraries create their own community addon and users can select it from a list within the CLI.

You can discuss this again in the next maintainers meeting.

PR #1160, opened 2026-07-03
docs: remove non-existent link

teemingc:
The link https://svelte.dev/docs/svelte/compiler-options doesn't exist and is causing docs build to fail.


manuel3108 (approved):
Approved, seems nonsense at that place as well

PR #1168, opened 2026-07-11
fix: add reactivity to paraglide

sacrosanctic:
Closes #

### Description

- add reactivity to paraglide
- fails without unrun
- remove `.gitignore`, the generated folder had a `.gitignore`

### Checklist

Co-authored-by: cofl <366548+cofl@users.noreply.github.com>

### DIFF
--- a/packages/sv/src/addons/paraglide.ts
+++ b/packages/sv/src/addons/paraglide.ts
@@ -207,11 +207,13 @@ export default defineAddon({
 			`src/hooks.client.${language}`,
 			transforms.script(({ ast, comments, js }) => {
 				js.imports.addNamed(ast, { imports: ['Locale'], from: '$lib/paraglide.svelte' });
-				js.imports.addNamed(ast, {
-					imports: ['ClientInit'],
-					from: '@sveltejs/kit',
-					isType: true
-				});
+				if (language === 'ts') {
+					js.imports.addNamed(ast, {
+						imports: ['ClientInit'],
+						from: '@sveltejs/kit',
+						isType: true
+					});
+				}
 
 				const init = dedent`
 					export const init${ts(': ClientInit')} = () => {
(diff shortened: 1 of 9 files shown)


packages/sv/src/addons/paraglide.ts
```diff
@@ -161,15 +170,56 @@ export default defineAddon({
...
+					from: '$lib/paraglide/runtime',
+					isType: true
+				});
+				js.imports.addNamed(ast, { imports: ['page'], from: '$app/state' });
+				js.imports.addNamed(ast, { imports: ['goto'], from: '$app/navigation' });
+				env.importEnv(ast, js, ['browser']);
+
+				const codeBlock = dedent`
+				export class Locale {
+					#current${ts(': _Locale')} = $state(toLocale(browser && document.querySelector('html')?.lang) ?? 'en');
```
cofl: Probably want to use baseLocale instead of hardcoding English as the fallback.

```suggestion
					#current${ts(': _Locale')} = $state(toLocale(browser && document.querySelector('html')?.lang) ?? baseLocale);
```

zsh:
what's the migration plan to existing project?

sacrosanctic:
> what's the migration plan to existing project?

Don't need one, it overrides the behaviour without needing to adjust any APIs. It is a breaking change in the sense that it is now reactive, but I think that's more of a fix than anything.

manuel3108 (approved):
This is way cleaner than what i implemented in my private project and works flawlessly!

Thank you!

zsh:
> > what's the migration plan to existing project?
> 
> Don't need one, it overrides the behaviour without needing to adjust any APIs. It is a breaking change in the sense that it is now reactive, but I think that's more of a fix than anything.

thank you for the follow up. i assume i have to rerun `npx sv add paraglide` once to get new helper code injected into project, right?

manuel3108:
Exactly, and you will have to wait for the next release, it's only merged but not yet published.

manuel3108:
Actually, i triggered the release right now, so you should be able to use this new version in about 5 minutes or so

PR #1174, opened 2026-07-17
fix(sveltekit-adapter): `static-adapter` follows the same setup as the docs

sacrosanctic:
Closes #

### Description

https://svelte.dev/docs/kit/adapter-static#Usage

### Checklist

### DIFF
--- a/.changeset/ready-mice-dream.md
+++ b/.changeset/ready-mice-dream.md
@@ -1,0 +1,5 @@
+---
+"sv": patch
+---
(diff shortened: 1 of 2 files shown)


manuel3108:
But that doesn't change any behavior, if I understand the docs correctly, as those are just the default options. If this is the case I don't think we should follow the docs here. I do understand that the docs point out those options, but i don't see the benefit of following them to the point.

PR #1184, opened 2026-07-22
breaking: remove deprecated utils and properties

manuel3108:
All of those utils and properties have been deprecated for a while - so im not bothered about them


jycouet (approved):
So good to see this :)

PR #1186, opened 2026-07-22
chore: replace semver with verkit

jycouet:
Replaces `semver` with [`verkit`](https://github.com/sxzz/verkit), a modern ESM-only, tree-shakeable SemVer toolkit following node-semver semantics. Part of e18e/ecosystem-issues#277.

Size impact:
- `sv` bundle (`sv-utils` dist/index.mjs): -15.9 kB / -2.8 kB gzip (1367.6 kB -> 1351.7 kB)
- `svelte-migrate` install: -21.8 kB unpacked (semver 100.2 kB -> verkit 78.4 kB), plus `@types/semver` no longer needed

All used APIs map 1:1 (`validRange`->`isValidRange`, `gtr`->`isGreaterThanRange`, `minVersion`->`findMinimumForRange`, ...); only real reshape is `coerce` returning a string instead of a SemVer object.


jycouet:
@manuel3108 should we target v1? or?

manuel3108:
Let's go with v1 to avoid later conflicts because of changes to the migrate package.

manuel3108:
Not sure why the actions decided to not run here, but LGTM

PR #1190, opened 2026-07-23
feat: smart package manager selection

### LINKED ISSUE #1159: Filter package managers by what's installed
Being a man of good taste, I don't have `yarn` installed on my machine. I accidentally fat-fingered the package manager choice when setting up a project just now, and so the installation errored.

What if we only showed the options that are installed? i.e. if `which yarn` doesn't return anything, omit it from the list — same for `bun` and `deno`. Downside: it's less clear that the CLI supports those things. But for someone who doesn't have them installed that's probably better anyway?

Another option: include them, but gray them out and make them unselectable:

```bash
◆  Which package manager do you want to install dependencies with?
│  ○ None
│  ○ npm
│  ○ yarn (not installed)
│  ● pnpm
│  ○ bun
│  ○ deno (not installed)
└
```

sacrosanctic:
image

Closes #1159

### Description

- order is `None`, `pnpm`, abc, not installed
- `command -v <pm>` over `which <pm>` based on [this](https://unix.stackexchange.com/questions/85249/why-not-use-which-what-to-use-then) discussion.

### todo
- [x] handle non interactive mode selecting an invalid pm

### Checklist

### DIFF
--- a/packages/sv/src/core/package-manager.ts
+++ b/packages/sv/src/core/package-manager.ts
@@ -73,6 +73,15 @@ export async function packageManagerPrompt(cwd: string): Promise<AgentName | und
 }
 
 export async function installDependencies(agent: AgentName, cwd: string): Promise<void> {
+	if (!isInstalled(agent)) {
+		p.log.error(
+			`Package manager ${color.command(agent)} is not installed. Please install it first.`
+		);
+		p.log.message();
+		p.cancel('Operation failed.');
+		process.exit(1);
+	}
+
 	const task = p.taskLog({
 		title: `Installing dependencies with ${color.command(agent)}...`,
 		limit: Math.ceil(process.stdout.rows / 2),
--- a/packages/sv/src/core/package-manager.ts
+++ b/packages/sv/src/core/package-manager.ts
@@ -13,32 +13,56 @@ import * as find from 'empathic/find';
 import fs from 'node:fs';
 import path from 'node:path';
 import process from 'node:process';
-import { exec } from 'tinyexec';
+import { exec, execSync } from 'tinyexec';
 
 export const AGENT_NAMES: AgentName[] = AGENTS.filter(
 	(agent): agent is AgentName => !agent.includes('@')
-);
-const agentOptions: PackageManagerOptions = AGENT_NAMES.map((pm) => ({ value: pm, label: pm }));
-agentOptions.unshift({ label: 'None', value: undefined });
+	// put `pnpm` first
+).sort((a, b) => {
+	if (a === 'pnpm') return -1;
+	if (b === 'pnpm') return 1;
+	return a.localeCompare(b);
+});
 
 export const installOption: Option = new Option(
 	'--install <package-manager>',
 	'installs dependencies with a specified package manager'
 ).choices(AGENT_NAMES);
+type PackageManagerOptions = Array<{
+	value: AgentName | undefined;
+	label: AgentName | 'None';
+	disabled?: boolean;
+	hint?: string;
+}>;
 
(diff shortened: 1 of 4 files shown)


packages/sv/src/core/package-manager.ts
```diff
@@ -13,32 +13,56 @@ import * as find from 'empathic/find';
...
+}>;
 
-type PackageManagerOptions = Array<{ value: AgentName | undefined; label: AgentName | 'None' }>;
 export async function packageManagerPrompt(cwd: string): Promise<AgentName | undefined> {
 	const detected = await detect({ cwd });
-	const agent = detected?.name ?? getUserAgent();
+	const detectedAgent = detected?.name ?? getUserAgent();
+	const installedAgents = AGENT_NAMES.filter(isInstalled);
+	const initialAgent =
+		detectedAgent && installedAgents.includes(detectedAgent) ? detectedAgent : installedAgents[0];
```
sacrosanctic: `detected` pm isn't always installed. If the user runs `sv add` in a `pnpm` project, and if the user doesn't have `pnpm` installed `sv` won't let you select it. Which doesn't really make sense.

jycouet: You managed to do "not selectable" in a multi select? (I didn't know it exist). 
Maybe even if you don't have it installed, this one should be selected _(and even selected my default)_ to keep the current behavior

sacrosanctic: ya, turns out `p.select` has a disabled property. But this is not multi-select. `pm` is just regular select.

sacrosanctic:
Currently, there are 2 situations:
- `--install <pm>` but `pm` is not installed
- project is detected as `pm` based on heuristics (like lockfile and package.json properties) but `pm` is not installed

In both cases, I've decided to just throw. Users have to install the `pm` in question or update the project to fit their desired `pm`.

jycouet:
> Currently, there are 2 situations:
> - `--install <pm>` but `pm` is not installed
> - project is detected as `pm` based on heuristics (like lockfile and package.json properties) but `pm` is not installed
> 
> In both cases, I've decided to just throw. Users have to install the `pm` in question or update the project to fit their desired `pm`.

I think that it's reasonable 👍
Or maybe we add in the next step ? 

I'll look today 👍 thx

sacrosanctic:
Oh ya, skipping works too. I just went with one so it is in a reviewable state.

Hmmm, now that I think about it, that might be better. You take a look first.

sacrosanctic:
Another thought:
- `sv add` shouldn't ask the user to select a pm, it should be able to detect one, just a binary question to install or not
- `--install <pm>` should overwrite detection?

jycouet:
> sv add shouldn't ask the user to select a pm, it should be able to detect one, just a binary question to install or not

Today it's auto selecting the detected pm, I think that it's great like this (it's acting like do you want to run install)

> --install <pm> should overwrite detection?

I think that it's already the case, and it's good.

jycouet:
Pushed a few tweaks on top of this:

- **Cross-platform `isInstalled`**: replaced `sh -c command -v` (breaks on Windows, no `sh`) with `execSync(agent, ['--version'])` - tinyexec resolves `.cmd` shims. Results are memoized so each pm is checked at most once per run.
- **No more hard exit when a pm is missing**: options are selectable with a dimmed `pnpm (not installed)` label instead of disabled, and `installDependencies` warns + skips instead of throwing. `create`/`add` then show "Install \<pm\>" + "\<pm\> install" in next steps.
- Removed the pnpm-first ordering (kept installed-first in the prompt) and the fallback to the first installed pm - with nothing detected, "None" stays the default.
- Non-interactive path returns early before any install checks.

WDYT @sacrosanctic ?

sacrosanctic:
lgtm

PR #1195, opened 2026-07-25
feat: skip package manager prompt when nothing needs installing

### LINKED ISSUE #1088: Don't show package manager install prompt if not needed
Adding the remote Svelte MCP server for Claude Code does not require any installation.

```
┌  Welcome to the Svelte CLI! (v0.15.3)
│
◇  Which client would you like to use?
│  claude code
│
◇  What setup would you like to use?
│  Remote
│
▲  CLAUDE.md already exists, we didn't touch it. See https://svelte.dev/docs/mcp/overview#Usage for manual setup.
│
◆  Successfully setup add-ons: mcp
│
│  None
│
│  To skip prompts next time, run:
●  npx sv@0.15.3 add mcp="ide:claude-code+setup:remote" --no-install
│
└  You're all set!
```

Seems to be the same for the local version. I thought it would be installed locally.

```json
{
  "mcpServers": {
    "svelte": {
      "type": "stdio",
      "command": "npx",
      "env": {},
      "args": ["-y", "@sveltejs/mcp"]
    }
  }
}
```

jycouet:
Closes #1088

`sv add` now only asks which package manager to use when an add-on actually adds or widens a dependency in `package.json` (e.g. `sv add mcp=ide:claude-code+setup:remote` needs nothing installed). Files are still formatted when the install is skipped for that reason.

Adds `isRangeWithin` to `@sveltejs/sv-utils`, so an existing stricter range (`^9.2.0` vs a requested `^9.0.0`) is kept and doesn't trigger an install.


sacrosanctic:
Darn, I was working on the issue too, but yours look better.

packages/sv/src/cli/add.ts
```diff
@@ -689,12 +689,12 @@ export async function runAddonsApply({
...
-			? null
-			: options.install === true
-				? await packageManagerPrompt(options.cwd)
-				: options.install;
+	// nothing new landed in `package.json`, so there is nothing to install and nothing to ask about
+	let packageManager: AgentName | null | undefined = null;
+	if (options.install !== false && installNeeded) {
+		packageManager =
+			options.install === true ? await packageManagerPrompt(options.cwd) : options.install;
+	}
```
sacrosanctic, line 697:
```suggestion
	let packageManager: AgentName | null | undefined = null;
	if (options.install === true) {
		// prompt user when `package.json` has changed
		if (installNeeded) {
			packageManager = await packageManagerPrompt(options.cwd);
		}
	} else if (options.install === false) {
		// user choose not to install
	} else {
		packageManager = options.install;
	}
```

packages/sv/src/cli/add.ts
```diff
@@ -755,8 +755,12 @@ export async function runAddonsApply({
 	if (packageManager) {
 		workspace.packageManager = packageManager;
 		await installDependencies(packageManager, options.cwd);
+	}
+
+	// still format when the install was skipped only because no new dependency was added
+	if (packageManager || (!installNeeded && options.install !== false)) {
```
sacrosanctic: really hard to understand this condition

packages/sv-utils/src/semver.ts
```diff
@@ -36,6 +36,19 @@ export function coerceVersion(str: string): Version {
...
+ * e.g. `^9.2.0` is within `^9.0.0` but `^8.0.0` is not.
+ * Unparseable inputs (`latest`, `workspace:*`, ...) return `false`.
+ */
+export function isRangeWithin(range: string, target: string): boolean {
+	try {
+		return isRangeSubset(range, target);
+	} catch {
+		return false;
+	}
+}
```
AdrianGonz97, line 50: the param names are a bit too ambiguous. i think sticking with `isRangeSubset`'s naming makes the most sense.
```suggestion
/**
 * Returns `true` when every version matching `subset` also matches `superset`,
 * e.g. `^9.2.0` is within `^9.0.0` but `^8.0.0` is not.
 * Unparseable inputs (`latest`, `workspace:*`, ...) return `false`.
 */
export function isRangeWithin(subset: string, superset: string): boolean {
	try {
		return isRangeSubset(subset, superset);
	} catch {
		return false;
	}
}
```

packages/sv-utils/src/semver.ts
```diff
@@ -36,6 +36,19 @@ export function coerceVersion(str: string): Version {
...
+
 export function isVersionUnsupportedBelow(
 	versionStr: string,
 	belowStr: string
```
AdrianGonz97, line 54: drive-by change
```suggestion
export function isVersionUnsupportedBelow(
	version: string,
	below: string
```

packages/sv-utils/src/tests/semver.ts
```diff
@@ -74,3 +74,22 @@ describe('minVersion', () => {
...
+		{ range: 'latest', target: '^9.0.0', expected: false },
+		{ range: 'workspace:^9.0.0', target: '^9.0.0', expected: false }
+	] as const;
+	it.each(combinations)(
+		'($range within $target) should be $expected',
+		({ range, target, expected }) => {
+			expect(isRangeWithin(range, target)).toEqual(expected);
+		}
+	);
+});
```
AdrianGonz97, line 95: ...continues the optional name change
```suggestion
describe('isRangeWithin', () => {
	const combinations = [
		{ subset: '^9.0.0', superset: '^9.0.0', expected: true },
		{ subset: '^9.2.0', superset: '^9.0.0', expected: true },
		{ subset: '9.2.0', superset: '^9.0.0', expected: true },
		{ subset: '^9.0.0', superset: '^9.2.0', expected: false },
		{ subset: '^8.0.0', superset: '^9.0.0', expected: false },
		{ subset: '*', superset: '^9.0.0', expected: false },
		{ subset: 'latest', superset: '^9.0.0', expected: false },
		{ subset: 'workspace:^9.0.0', superset: '^9.0.0', expected: false }
	] as const;
	it.each(combinations)(
		'($subset within $superset) should be $expected',
		({ subset, superset, expected }) => {
			expect(isRangeWithin(subset, superset)).toEqual(expected);
		}
	);
});
```

packages/sv/src/core/engine.ts
```diff
@@ -36,35 +37,40 @@ function alphabetizeRecord(obj: Record<string, string>) {
...
 function updatePackages(
 	dependencies: Array<{ pkg: string; version: string; dev: boolean }>,
 	sv: SvApi
-) {
-	if (dependencies.length === 0) return;
+): boolean {
+	if (dependencies.length === 0) return false;
 
+	let installNeeded = false;
 	const pkgPath = filePaths.packageJson;
```
AdrianGonz97, line 49: this is very much a nitpick, but returning an object should remove the need for the comment here
```suggestion
/** Returns `true` when `package.json` gained or widened a dependency, i.e. an install is required. */
function updatePackages(
	dependencies: Array<{ pkg: string; version: string; dev: boolean }>,
	sv: SvApi
): { installNeeded: boolean } {
	let installNeeded = false;
	if (dependencies.length === 0) return { installNeeded };

	const pkgPath = filePaths.packageJson;
```

packages/sv/src/core/engine.ts
```diff
@@ -36,35 +37,40 @@ function alphabetizeRecord(obj: Record<string, string>) {
...
 
-			if (!modified) return false; // do not edit the file if no changes were made
+			if (!installNeeded) return false; // do not edit the file if no changes were made
 
 			if (data.dependencies) data.dependencies = alphabetizeRecord(data.dependencies);
 			if (data.devDependencies) data.devDependencies = alphabetizeRecord(data.devDependencies);
 		})
 	);
+
+	return installNeeded;
```
AdrianGonz97, line 74: a follow-up from the previous suggestion
```suggestion
	return { installNeeded };
```

packages/sv/src/core/engine.ts
```diff
@@ -386,11 +409,12 @@ export function prepareSvApi(
 	return {
 		sv,
 		finalize: () => {
-			updatePackages(dependencies, sv);
+			const installNeeded = updatePackages(dependencies, sv);
```
AdrianGonz97, line 424: another follow-up
```suggestion
			const { installNeeded } = updatePackages(dependencies, sv);
```

packages/sv/src/cli/add.ts
```diff
@@ -755,8 +755,12 @@ export async function runAddonsApply({
...
 		await formatFiles({
-			packageManager,
+			packageManager: packageManager ?? workspace.packageManager,
 			cwd: options.cwd,
 			filesToFormat,
 			strategy: 'files-only'
```
AdrianGonz97: im not exactly sure when this changed, but iirc, we used to guard formatting files behind whether we ran the deps install ourselves because we couldn't guarantee that the formatter was installed locally. if it wasn't, then it would result in an error when attempting to format. 

I think that seems to be the case now. for example:
1. create a new project
2. install prettier and eslint with their respective add-ons
3. delete node_modules
4. try running the eslint add-on

packages/sv/src/core/engine.ts
jycouet: I'll do now by hand... as GH has issue to merge suggestion

packages/sv/src/cli/add.ts
jycouet: Good catch. Decoupled it rather than re-coupling: `sv` now checks that `prettier` actually resolves from the project before formatting, instead of inferring it from "did we install". Your repro → no `node_modules` → format skipped.

Minor: it doesn't error out, `formatFiles` catches and logs `Failed to format files`.

.changeset/quiet-pans-invent.md
```diff
@@ -0,0 +1,5 @@
+---
+'sv': patch
+---
+
+feat: skip the package manager prompt when no dependency needs installing
```
manuel3108:
```suggestion
feat(cli): skip the package manager prompt when no dependency needs installing
```

manuel3108 (approved):
LGTM

PR #1185, opened 2026-07-22
feat(create): use `#lib` instead of `$lib`

manuel3108:
Relates #1111 

### Checklist

- Update [snapshots](../CONTRIBUTING.md#update-snapshots) (if applicable)
- Add a [changeset](../CONTRIBUTING.md#generating-changelogs) (if applicable)
- Allow maintainers to edit this PR
- I care about what I'm doing, no matter the tool I use (Notepad, Sublime, VSCode, AI...)

PR #1196, opened 2026-07-25
chore: merge main into version-1

jycouet:
Brings `main` into `version-1` in a single merge commit.

Conflict resolutions:
- keep the v1 side (verkit instead of semver, deprecated APIs removed), while taking main's newer `package-manager-detector@^1.8.0`
- `create`/`add`: combine main's "skip formatting when install failed" with v1's `strategy: 'files-only'`
- `svelte-migrate` is a deprecated stub on v1, so its changeset entry and deps are dropped

PR #1197, opened 2026-07-25
feat(migrate): retarget ts/jsconfig at $app/tsconfig

jycouet:
SvelteKit 3 no longer writes `.svelte-kit/tsconfig.json` - the generated config now lives at `node_modules/$app/tsconfig` and no longer carries `include`. Existing projects break on upgrade (e.g. `better-auth` can't read its config, the `drizzle` addon's `typescript.config` hook throws on `config.include.push`).

New `tsconfig` task in the `sveltekit-3` migration:

- points `extends` at `$app/tsconfig` and gives the project its own `include`
- drops compiler options that just duplicate the generated ones
- moves paths added by the now-deprecated `typescript.config` hook into `include`, flagging hooks that do more

It will help: https://github.com/sveltejs/cli/pull/1185


jycouet:
I move it to draft because we need https://github.com/sveltejs/cli/pull/1196 before IMO

manuel3108 (approved):
Not sure how a migration script should help #1185, but this is required either way. 

LGTM

PR #1201, opened 2026-07-30
fix(better-auth): use the `auth` CLI binary in the auth:schema script

jycouet:
The addon installs the `auth` package, which ships both `auth` and `better-auth` bins. The generated `auth:schema` script now calls `auth generate` instead of the legacy `better-auth` alias.

Snapshots updated accordingly.

### DIFF
--- a/packages/sv/src/addons/better-auth.ts
+++ b/packages/sv/src/addons/better-auth.ts
@@ -181,7 +181,7 @@ export default defineAddon({
 					/**
 					 * DO NOT USE!
 					 *
-					 * This instance is used by the \`better-auth\` CLI for schema generation ONLY.
+					 * This instance is used by the \`auth\` CLI for schema generation ONLY.
 					 * To access \`auth\` at runtime, use \`event.locals.auth\`.
 					 */
 					export const auth = createAuth(null${ts('!')});`;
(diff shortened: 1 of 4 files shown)


packages/sv/src/addons/better-auth.ts
```diff
@@ -214,7 +212,7 @@ export default defineAddon({
 				json.packageScriptsUpsert(
 					data,
 					'auth:schema',
-					`better-auth generate --config ${authConfigPath} --output ${authSchemaPath} --yes`
+					`auth generate --config ${authConfigPath} --output ${authSchemaPath} --yes`
 				);
 			})
 		);
```
AdrianGonz97, line 218: I wonder if we even need to dedicate a package script slot just for this command? It _seems_ to be something that's only ever executed once (when you're getting started with `better-auth`) and will probably never be ran again

Perhaps just adding the prompt: 
```sh
Add-on steps
# other steps...
Run `pnpm dlx auth generate --config src/lib/server/auth.ts --output src/lib/server/db/auth.schema.ts --yes`
```
in `nextSteps` would be sufficient?

sacrosanctic: It needs to run everytime the better-auth config changes that affect the db tables

jycouet: Yes, as it's a command that will be used multiple times it's almost acting as docs to me _(and to AI)_.
Also, storing it make it possible to change path & co in a "documented" way.

Issue #1092, opened 2026-05-15
Add `"types": ["node"]` in `tsconfig.json` when needed


hyunbinseo:
### Describe the bug

When `pnpm check` is ran in the default minimal template:

```sh
pnpm dlx sv@0.15.3 create --template minimal --types ts --install pnpm ./
```

This warning is logged, which is valid since `@types/node` is not installed.

```
d:\_\tsconfig.json:1:1
Warn: Cannot find type definition file for 'node'.
  The file is in the program because:
    Entry point of type library 'node' specified in compilerOptions
```

Why not move `"types": ["node"],` to the `tsconfig.json` generated by the SV CLI?

```diff
- .svelte-kit\tsconfig.json
+ tsconfig.json
```

e.g. When ESLint add-on is selected which relies on `node:path`, this can be done:

1. `@types/node` is installed
2. `"types": ["node"],` is added

### Reproduction

```sh
pnpm dlx sv@0.15.3 create --template minimal --types ts --install pnpm ./
pnpm check

pnpm i @types/node -D
pnpm check
```

### Logs

```Shell

```

### System Info

```Shell
System:
  OS: Windows 11 10.0.26200
  CPU: (8) x64 Intel(R) Core(TM) Ultra 7 258V
  Memory: 2.85 GB / 31.48 GB
Binaries:
  Node: 24.15.0 - C:\Users\hyunb\AppData\Local\fnm_multishells\28376_1778829182710\node.EXE
  Yarn: 1.22.22 - C:\Users\hyunb\AppData\Local\fnm_multishells\28376_1778829182710\yarn.CMD
  npm: 11.12.1 - C:\Users\hyunb\AppData\Local\fnm_multishells\28376_1778829182710\npm.CMD
  pnpm: 11.1.2 - C:\Users\hyunb\AppData\Local\fnm_multishells\28376_1778829182710\pnpm.CMD
  Deno: 2.4.0 - C:\Users\hyunb\.deno\bin\deno.EXE
Browsers:
  Chrome: 148.0.7778.97
  Edge: Chromium (140.0.3485.54)
  Firefox: 150.0.3 - C:\Program Files\Mozilla Firefox\firefox.exe
npmPackages:
  @sveltejs/adapter-auto: ^7.0.1 => 7.0.1
  @sveltejs/kit: ^2.57.0 => 2.59.1
  @sveltejs/vite-plugin-svelte: ^7.0.0 => 7.1.2
[...]

PR #1202, opened 2026-07-30
fix(eslint): use `recommendedTypeChecked`

### LINKED ISSUE #938: Consider using `recommendedTypeChecked` with `typescript-eslint`
https://typescript-eslint.io/getting-started/typed-linting/

https://github.com/sveltejs/kit/issues/11909#issuecomment-1967835976

sacrosanctic:
Closes #938

### Description

- [ ] address https://github.com/sveltejs/cli/issues/1092
- [ ] fix `vite-ts` test

### Checklist


packages/sv/src/addons/better-auth.ts
```diff
@@ -364,9 +364,9 @@ export default defineAddon({
 						},
 						signUpEmail: async (event) => {${d1AuthLine}
 							const formData = await event.request.formData();
-							const email = formData.get('email')?.toString() ?? '';
-							const password = formData.get('password')?.toString() ?? '';
-							const name = formData.get('name')?.toString() ?? '';
+							const email = (formData.get('email')${ts(' as string')}) ?? '';
+							const password = (formData.get('password')${ts(' as string')}) ?? '';
+							const name = (formData.get('name')${ts(' as string')}) ?? '';
```
AdrianGonz97: I would suggest reverting these back to their `?.toString()` state. 

The type assertion is not only more verbose in our template literal, it's also incorrect. Try isolating the expression without the nullish coalescing operator (`??`):
```ts
const email = (formData.get('name') as string)
//    ^? const email: string
```
it falsely asserts that the type is a `string` rather than `string | null`

sacrosanctic: Can't revert, it fails linting.

sacrosanctic: Thanks, I replaced it with something more robust

sacrosanctic:
```
 ❯  addons  tests/eslint/test.ts (4 tests | 1 failed) 185289ms
   ✓ eslint 'kit-js'  113307ms
   ✓ eslint 'kit-ts'  113306ms
   ✓ eslint 'vite-js'  113306ms
   × eslint 'vite-ts' 173888ms (retry x3)
```
I don't understand how this project is structured. It has 2 `tsconfig.json`'s.

AdrianGonz97:
While looking into the lint failure for https://github.com/sveltejs/cli/pull/1202#discussion_r3694155435, I started to go through the [new lint rules](https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/src/configs/eslintrc/recommended-type-checked-only.ts) (this list only contains what's added by `recommendedTypeChecked` to `recommended`). 

I'm left wondering if the preset rules for `recommendedTypeChecked` may be too opinionated, or perhaps contain some questionable defaults?

For example, I can already foresee [`no-floating-promises`](https://typescript-eslint.io/rules/no-floating-promises/) becoming a problem when it comes to [deferred query refreshes](https://svelte.dev/docs/kit/remote-functions#Single-flight-mutations-Server-driven-refreshes) with remote functions. 

something like this:
```ts
const foo = form(() => {
	// ...stuff happens
	
	query1().refresh() // lint error!
	query2().refresh() // lint error!
})
```
could erroneously lead people down the very wrong path of `await`ing these refreshes:
```ts
const foo = form(() => {
	// ...stuff happens

	// oops! these no longer run concurrently!
	await query1().refresh()
	await query2().refresh()
})
```

however, while adding `void query1().refresh()` may dismiss the lint error, that 'fix' will not be everyone's first instinct.

[`only-throw-error`](https://typescript-eslint.io/rules/only-throw-error) _could_ be an issue with thrown a `redirect`, though i haven't tested it so i can't be certain at the moment. 

I also _vaguely remember_ [`require-await`](https://typescript-eslint.io/rules/require-await/) being the source of some annoying issues when we first tried to enable it for this repo, but I'm not able to recall exactly why. Do you guys (@manuel3108, @benmccann) happen to remember?

sacrosanctic:
Good stuff, definitely things to consider, hopefully this PR can create the discussion needed to merge this or to close the original issue. Will have to do more exploration with the new APIs.

> however, while adding void query1().refresh() may dismiss the lint error, that 'fix' will not be everyone's first instinct.

It is part of the lint error, maybe we can overwrite it with our own custom error msg?

> [only-throw-error](https://typescript-eslint.io/rules/only-throw-error) could be an issue with thrown a redirect

Don't think this is an issue, we don't throw `error` or `redirect`.

manuel3108:
> I also _vaguely remember_ [`require-await`](https://typescript-eslint.io/rules/require-await/) being the source of some annoying issues when we first tried to enable it for this repo, but I'm not able to recall exactly why. Do you guys (@manuel3108, @benmccann) happen to remember?

Sadly i don't remember any of this. But I see your points.

Do  I see that correctly, that this PR tries to change the setup for new projects (`sv create`) and our own codebase? I mean it totally makes sense, just trying to understand the scope of this PR

manuel3108:
Changed target branch and resolved produced merge conflicts

PR #1188, opened 2026-07-23
fix: improve demo experience

sacrosanctic:
Closes #1169

### Description

I started out using the new `addOption` API, but those aren't available as CLI args. The alternative is to add a `template` property in the condition

```diff
.add('demo', {
  type: 'boolean',
  default: true,
  question: 'Do you want to include a demo?',
+  condition: ({ template }) => template !== 'demo'
})
```

@jycouet please advise.

### Checklist

- Update [snapshots](../CONTRIBUTING.md#update-snapshots) (if applicable)
- Add a [changeset](../CONTRIBUTING.md#generating-changelogs) (if applicable)
- Allow maintainers to edit this PR
- I care about what I'm doing, no matter the tool I use (Notepad, Sublime, VSCode, AI...)

PR #1203, opened 2026-07-30
feat(sv-utils): demo helpers

### LINKED ISSUE #1169: update demo experience
- [ ] On the demo template, add a fourth header called `addon` which links to the `/demo` page.
- [ ] when `sv create`, addons should not prompt to add demo if the template is `demo`
- [ ] rename `/demo` to `/addon`

sacrosanctic:
Closes #1169

### Description

Reopened from #1188 due to a glitch.

### Checklist


packages/sv/src/addons/better-auth.ts
```diff
@@ -27,7 +27,8 @@ const options = defineAddonOptions()
 			{ value: 'password', label: 'Email & Password' },
 			{ value: 'github', label: 'GitHub OAuth' }
 		],
-		required: false
+		required: false,
+		condition: ({ template }) => template !== 'demo'
```
sacrosanctic: @jycouet the general idea of this PR is this, thoughts?

jycouet: - I find it strange to not allow better-auth demo in demo template

packages/sv/src/addons/tests/better-auth/test.ts
```diff
@@ -27,7 +27,7 @@ test.concurrent.for(testCases)('better-auth $variant', async (testCase, { page,
 	const language = testCase.variant.includes('ts') ? 'ts' : 'js';
 
 	// Verify that we have a demo login
-	const loginPage = path.resolve(cwd, `src/routes/demo/better-auth/login/+page.svelte`);
+	const loginPage = path.resolve(cwd, `src/routes/addon/better-auth/login/+page.svelte`);
```
jycouet: I'm not sure I like the `/demo` -> `/addon`. (I can probably be convinced!)

But it's like "Do you want a demo?" here is a page `/addon`

packages/sv/src/core/options.ts
```diff
@@ -37,7 +37,7 @@ export type BaseQuestion<Args extends OptionDefinition> = {
 	 * When this condition explicitly returns `false`, the question's value will
 	 * always be `undefined` and will not fallback to the specified `default` value.
 	 */
-	condition?: (options: OptionValues<Args>) => boolean;
+	condition?: (context: { values: OptionValues<Args>; template?: string }) => boolean;
```
jycouet: This is a big breaking change! no?!

All this to not allow better-auth in the demo template? (Or somthing else more?)

jycouet (commented):
All in all, I'm not a big fan.

I re read the issue a few times, and: 
> On the demo template, add a fourth header called addon which links to the /demo page.

It could also be `/demo` as a 4th tab?

Actually, in the minimal template, I would love to have a link to the demo page from the root layout! (It's maybe VERY intrusive... but very visible for people adding something in their app!)

> when sv create, addons should not prompt to add demo if the template is demo

I think that this is not necessary. Why doing "soooo much work" for so small outcome? Maybe I'm missing somehting?

> rename /demo to /addon

Possible yes, but I don't really see the cool thing about it

packages/sv/src/addons/better-auth.ts
sacrosanctic: The logic got lost during the rebasing. It's supposed to be "when template is demo, don't ask, always use template"

jycouet: > The logic got lost during the rebasing. It's supposed to be "when template is demo, don't ask, always use template"

I guess you meant `when template is demo, don't ask, always use demo` ?
I don't really see the value there and it's doing a big breaking, and it's only for create.

packages/sv/src/core/options.ts
sacrosanctic: removed the breaking change

packages/sv/src/addons/tests/better-auth/test.ts
sacrosanctic: reversed it to reduce the git diff, and it seems `/demo` was the better choice in all cases

sacrosanctic:
changes:
- removed the breaking change by making it additive
- `/addon` back to `/demo`

Overall, the cumulative diff has been reduced so its easier to see the intent. I think less prompting is better. Especially when it comes to demos, which is used by new users. And it's a redundant question.

api changes
- Regardless of this PR, I do think we should have access to `template` from within the addon
- widening of type for `svelte.addFragment`

sacrosanctic:
```ts
const demo = createDemoPage('paraglide', language, directory.kitRoutes);
sv.file(...demo.listing);
sv.file(...demo.header);
```

Im also interested in making this a `sv-utils` so the community can easily add their demo to the demopage.

sacrosanctic:
I would like this merged if a maintainer can provide some more feedback.

packages/sv/src/addons/better-auth.ts
```diff
@@ -44,14 +45,14 @@ export default defineAddon({
...
 		const lib = resolveLibPrefix(dependencyVersion('@sveltejs/kit'));
 		const svelteVersion = dependencyVersion('svelte');
 		const svelte5 = !!svelteVersion && coerceVersion(svelteVersion).major === 5;
 		const [ts, s5] = createPrinter(language === 'ts', svelte5);
 
-		const demoPassword = options.demo.includes('password');
-		const demoGithub = options.demo.includes('github');
+		const demoPassword = template === 'demo' || options.demo?.includes('password');
+		const demoGithub = template === 'demo' || options.demo?.includes('github');
 		const hasDemo = demoPassword || demoGithub;
```
jycouet: Why forcing the `password` demo and the `github` demo in the demo template?

packages/sv/src/addons/better-auth.ts
```diff
@@ -559,18 +559,18 @@ export default defineAddon({
...
 		];
-		if (options.demo.includes('github')) {
+		if (options.demo?.includes('github')) {
 			steps.push(
 				`Set your ${color.env('GITHUB_CLIENT_ID')} and ${color.env('GITHUB_CLIENT_SECRET')} in ${color.path('.env')}`
 			);
 		}
-		if (options.demo.length > 0) {
+		if (template === 'demo' || (options.demo && options.demo.length > 0)) {
 			steps.push(`Visit ${color.route('/demo/better-auth')} route to view the demo`);
```
jycouet, line 574:
```suggestion
			steps.push(`Visit ${color.route(demo.addonPath)} route to view the demo`);
```

jycouet (commented):
I'm not a big fan of this new `template` param that is available only in `create` scenario.
And I don't see the value of skipping demo questions in the demo template.
Am I missing something?

packages/sv/src/addons/better-auth.ts
sacrosanctic: We should prompt the user instead of forcing.

packages/sv/src/addons/better-auth.ts
sacrosanctic: `demo` is scoped to `run` and is not available in `nextSteps`

sacrosanctic:
> I'm not a big fan of this new template param that is available only in create scenario.

Originally tried [dynamic options](https://github.com/sveltejs/cli/pull/1188/changes/3b5319d676aca0079136a7ab7fa0613c9345cae4#diff-371b8370462b488027defaa41df3ddd226598cf121a4decb3bfcd45f018b0bdbR33-R44) but didn't work. `sv create` is the only time template value is valid, there's no template in `sv add` 

> And I don't see the value of skipping demo questions in the demo template.

What is the value of asking? If you select demo template and addons that has demos, you want a playground to mess around with.

|  | want demos | dont want demos |
|--------|--------|--------|
| demo template | new users | who |
| minimal template | tryout addon | experienced users |

jycouet:
> > And I don't see the value of skipping demo questions in the demo template.
> 
> What is the value of asking? If you select demo template and addons that has demos, you want a playground to mess around with.
> 
> want demos	dont want demos
> demo template	new users	who
> minimal template	tryout addon	experienced users

- maybe a new user want to know only the code about `password` and not `github`.
- I like to not have too many hidden `if` in addon. To avoid "I ran this add-on but now it not behaving the same, why?"
- With your matrix, you are guessing what question to ask depending on "user level". I think that it's too much guessing over predictability.

sacrosanctic (commented):
I see where you're coming from. On the other side, it feels like you're getting asked the same question again.
"Which template do you want?" "Demo" "Do you want to include a demo?" "Didnt i just answer this question?"

I have reversed the changes to complex questions. We only skip questions that are binary like 'Do you want to include a demo?'. If you remain unconvinced, then we can move on.

manuel3108:
I think the committed implementation at this point is quite harmless. I don't have a strong opinion here but I think I would be slightly in favor of proceeding with this. Basically the current implementation skips the demo question if the sveltekit demo was used. Thats a total of two skipped questions or something.

But I will leave the final decision to @jycouet

jycouet:
I'm sorry, but I strongly think that this API change adding template to remove one question only in create scenario is not worth it. Plus it's creating issue with `paraglide=demo:false` that would not be respected. Plus create & add would have different questions and that's not really consistent.

I like the idea of moving the helper to sv-utils to add more demos there by the community. 
I'm moving it now to then merge like this.

jycouet (approved):

packages/sv-utils/src/demo-page.ts
```diff
@@ -0,0 +1,89 @@
...
+ * Both transforms bail out once their link exists, so re-running an add-on
+ * won't duplicate entries. The header transform also bails on templates whose
+ * layout has no nav list.
+ */
+export function defineDemoPage(name: string, language: 'ts' | 'js', kitRoutes: string): DemoPage {
+	const listing = transforms.svelteScript({ language }, ({ ast, js, svelte }) => {
+		if (hasDemoLink(ast.fragment.nodes, name)) return false;
+
+		js.imports.addNamed(ast.instance.content, { imports: ['resolve'], from: '$app/paths' });
+		svelte.addFragment(ast, `<a href={resolve('/demo/${name}')}>${name}</a>`, { mode: 'prepend' });
```
sacrosanctic, line 66: Should this be append?

sacrosanctic:
What happened to this PR. Did I close it by accident?

jycouet:
What if we go even further?
1. upsert in `DemoLinks.svelte` demo links.
2. upsert in +layout.svelte the `<DemoLinks />`

Like this, no need to explain that there is a `/demo` route to visit.
Easy opt-out, delete the component :)

sacrosanctic (commented):
i don't follow. You want to upsert <DemoLinks /> into routes/+layout.svelte? this would be for the non demo template? what about the demo template?

jycouet:
> i don't follow. You want to upsert into routes/+layout.svelte? this would be for the non demo template? what about the demo template?

I think that I never want to do something different depending on the template.
Yes, adding to the +layout.svelte to have cool discoverability & easy removal!
Not something OMG beatiful, but also not too ugly :)

I can POC if you want to on another branch

PR #1187, opened 2026-07-22
fix: recognize the nub package manager

colinhacks:
### Description

`sv` delegates package-manager detection and command construction to [`package-manager-detector`](https://github.com/antfu/package-manager-detector). That library added the [`nub`](https://nubjs.com) package manager in v1.8.0 — its agent table, the `nub.lock` lockfile mapping, and the command set (`nub install`, `nub run`, `nubx` for `execute`, `nub exec` for `execute-local`).

`@sveltejs/sv-utils` bundles `package-manager-detector` at build time (`tsdown` `onlyBundle`), so the published bundle only knows about nub once rebuilt against ≥1.8.0. The lockfile currently pins 1.6.0, so `sv create` in a nub project misdetects it as npm.

This bumps the range to `^1.8.0` in `@sveltejs/sv-utils` and `svelte-migrate` and regenerates the lockfile. No source changes — detection is fully delegated, and `AGENTS`/`AGENT_NAMES` derive from the library.

Verified against a fresh build of `packages/sv-utils/dist/index.mjs`:

- `AGENTS` includes `nub`
- `resolveCommand('nub', 'run', ['test'])` → `nub run test`
- `resolveCommand('nub', 'install', [])` → `nub install`
- `resolveCommand('nub', 'execute', ['sv'])` → `nubx sv`
- `resolveCommand('nub', 'execute-local', ['sv'])` → `nub exec sv`

### Checklist

- Update [snapshots](../CONTRIBUTING.md#update-snapshots) (if applicable) — n/a
- Add a [changeset](../CONTRIBUTING.md#generating-changelogs) (if applicable) — done
- Allow maintainers to edit this PR
- I care about what I'm doing, no matter the tool I use (Notepad, Sublime, VSCode, AI...)

PR #1207, opened 2026-08-01
fix(cli): update pm layout

sacrosanctic:
| Before | After |
|--------|--------|
|  |  |

Closes #

### Description

Don't know if it affects anyone else, but `pm` are invisible to me, I also added some padding.

### Checklist

### DIFF
--- a/packages/sv/src/core/package-manager.ts
+++ b/packages/sv/src/core/package-manager.ts
@@ -32,25 +32,17 @@ export async function packageManagerPrompt(cwd: string): Promise<AgentName | und
 	// There is no need to prompt in that case.
 	if (!process.stdout.isTTY) return agent;
 
-	const maxAgentLength = AGENT_NAMES.reduce((a, c) => (c.length > a ? c.length : a), 0);
-	// installed ones first
 	const agentOptions = [
 		{ label: 'None', value: undefined },
-		...AGENT_NAMES.map((agent) => {
-			const installed = isInstalled(agent);
-			return {
-				value: agent,
-				label: installed ? agent : `${agent.padEnd(maxAgentLength)} (not installed)`,
-				installed
-			};
-		}).sort((a, b) => Number(b.installed) - Number(a.installed))
+		...AGENT_NAMES.flatMap((agent) => (isInstalled(agent) ? { value: agent } : []))
 	];
 
 	const pm = await p.select({
-		message: 'Which package manager do you want to install dependencies with?',
+		message: 'Detected package managers. Which one should we use to install dependencies?',
 		options: agentOptions,
 		initialValue: agent
 	});
+
 	if (p.isCancel(pm)) {
 		p.cancel('Operation cancelled.');
 		process.exit(1);
--- a/packages/sv/src/core/package-manager.ts
+++ b/packages/sv/src/core/package-manager.ts
@@ -34,7 +34,7 @@ export async function packageManagerPrompt(cwd: string): Promise<AgentName | und
 
 	const agentOptions = [
 		{ label: 'None', value: undefined },
-		...AGENT_NAMES.flatMap((agent) => (isInstalled(agent) ? { value: agent } : []))
+		...AGENT_NAMES.filter(isInstalled).map((agent) => ({ value: agent }))
 	];
 
 	const pm = await p.select({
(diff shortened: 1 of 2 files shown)


packages/sv/src/core/package-manager.ts
```diff
@@ -32,14 +32,15 @@ export async function packageManagerPrompt(cwd: string): Promise<AgentName | und
...
 		{ label: 'None', value: undefined },
 		...AGENT_NAMES.map((agent) => {
 			const installed = isInstalled(agent);
 			return {
 				value: agent,
-				label: installed ? agent : color.dim(`${agent} (not installed)`),
+				label: installed ? agent : `${agent.padEnd(maxAgentLength)} (not installed)`,
 				installed
 			};
 		}).sort((a, b) => Number(b.installed) - Number(a.installed))
```
AdrianGonz97: hmmm, shouldn't we just not include non-installed package managers in the list? something like...
```ts
...AGENT_NAMES.filter(isInstalled).map((agent) => {
	// ...
});
```

it feels a bit cluttered if we show _every_ possible option, even when they're not installed on the system

sacrosanctic: This was discussed in the previous #1190, we decided to make them visible (so users are aware of the capability) and selectable with no install fallback (for a better DX). Though subsequently, #1187 was merged and introduced additional `pm`.

AdrianGonz97: yea, i don't necessarily agree that showing non-installed PMs is a better DX. if the user is interested in yarn, bun, deno, etc. then they would probably have them installed already (and those that _are_ actually interested will see it as an option anyway)

it also doesn't really solve the core issue of #1159 since those non-installed options are still selectable/fat-fingerable. 

i think it would be best just to omit them entirely and declutter the prompt*

jycouet: A few random thoughts:

a/ `dim` is not working in your terminal? :o What are you using @sacrosanctic ?
b/ Again a new package manager `aube` ?! :/ And `pnpm-rush` ? :o
c/  It's kinka expressive (in time) to check if pm is here... and now we ahve to do if for 8 things?! :(
d/ The argument for showing even if not installed was more in a `add` context when you are in a project using bun and you don't have bun yet. But you are probably right @AdrianGonz97 it's very niche. we could just hide in UI what you don't have and that's it. (and keep handling the nice error when you do it via cmd)

sacrosanctic: image


a/ `alacritty`+`zsh`, i checked in the default terminal and that does show, but barely.
b/ Ya, we didn't catch that because the version bump was in a different PR.
c/ lol, it's ok. They're in parallel, right?
d/ You want to hide it in `create` and `add`?

This kind of goes back to what I was saying earlier. Do we even need to ask when using `add`? Just install it.

sacrosanctic: @AdrianGonz97 I agree, we can update the description to mentioned "detected PM's".

AdrianGonz97: > This kind of goes back to what I was saying earlier. Do we even need to ask when using `add`? Just install it.

I would say that we should always prompt it. 

I don't remember if we ever did ship auto-installing with the detected PM in the past in either `svelte-add` or `sv` (or if it was one of my other CLIs?), but I distinctly remember it becoming a PIA in some cases. 

The auto-pm-detect feature is a bit easy to fool. The typical case I saw was when the user had a lockfile in a parent directory. For instance, the most common report was about `yarn` being used because a `yarn.lock` file was found in their home directory. I'm not sure how it gets there in the first place, but it's happened enough times for several people to report it.

As for only doing it in `add`: there's no guarantee that they ran `install` during create, or that they have a lockfile in the first place, so I would still prompt for it anyway.

jycouet: Yes, I would like to keep the question anyway.
And just remove not installed stuff in the UI.
image

If the default select is on something no installed, just go to None.

AdrianGonz97 (approved):

jycouet (approved):

sacrosanctic:
@AdrianGonz97 Is `flatmap` less performant?

AdrianGonz97:
No, it's just a readability change. This is a tiny list of elements that's not in a hot path, so the performance cost is negligible

PR #1208, opened 2026-08-01
chore: knip informed clean up

sacrosanctic:
Closes #

### Checklist


knip.jsonc
```diff
@@ -0,0 +1,28 @@
...
+			],
+			"ignore": ["src/**/tests/**", "src/create/templates/**", "src/create/shared/**"]
+		},
+		"packages/migrate": {
+			"entry": ["migrations/**/index.js"]
+		}
+	},
+	"ignoreExportsUsedInFile": { "type": true },
+	"ignoreBinaries": ["wait"]
+}
```
jycouet: Happy to see that it catched only 1 or 2 nitpicks!
_It's also because a lot of stuff are ignored, maybe there are more stuff to knip await ^^_

jycouet (commented):
I like knip a lot, but I'm not sure it's very helpful at the moment here.
Some other views?

---

_BTW, would be good to link issue and/or description to give a little more context._

knip.jsonc
sacrosanctic: - tests are ignored, but could be refined to just ignore the test output
- `/templates` and `/shared` shouldnt be parsed, they're not actually a part of the project with regards to imports and exports
- the last 2 are knip bugs, or they require a lot of refactoring to fix

sacrosanctic:
> I like knip a lot, but I'm not sure it's very helpful at the moment here.
Some other views?

We don't have to add it, we can just keep the changes it found.

> BTW, would be good to link issue and/or description to give a little more context.

No issue, doing it showcases whether its worth it or not. Also feels pretty self explanatory along the lines of `prettier` and `eslint`.

packages/migrate/migrations/app-state/migrate.spec.js
```diff
@@ -163,3 +163,23 @@ test('Does not update $app/store #3', () => {
...
+		result,
+		`{
+	"name": "svelte-app",
+	"version": "1.0.0",
+	"dependencies": {
+		"@sveltejs/kit": "^2.12.0"
+	}
+}`
+	);
+});
```
jycouet: I'm not sure this test is really needed. same in packages/migrate/migrations/sveltekit-2/migrate.spec.js

sacrosanctic: removed

jycouet (approved):

manuel3108 (approved):

PR #1209, opened 2026-08-01
chore: replace `tar-fs` with `modern-tar`

sacrosanctic:
Closes #

### Description

should be a dropin replacement

### Checklist


tsdown.config.ts
```diff
@@ -37,7 +37,7 @@ const svDeps = {
 		'text-decoder',
 		'streamx',
 		'tar-stream',
-		'tar-fs',
+		'modern-tar',
```
jycouet: I think that we can remove more entries here (not just `tar-fs`).
If you comment the list & build you will see what is needed if I remember correctly.

Would be great to get rid of all that at some point! (not now)

packages/sv/src/core/fetch-packages.ts
```diff
@@ -125,8 +125,8 @@ export async function downloadPackage(options: DownloadOptions): Promise<AddonDe
 	await pipeline(
 		data.body,
 		createGunzip(),
-		extract(NODE_MODULES, {
-			map: (header: any) => {
+		unpackTar(NODE_MODULES, {
+			map: (header) => {
 				// file paths from the tarball will always have a `package/` prefix,
 				// so we'll need to replace it with the name of the package
 				header.name = header.name.replace('package', pkg.name);
```
jycouet: I think that we can do a bit more here with the new `unpackTar`.

Looking at https://github.com/ayuhito/modern-tar#filtering-and-transformation we could have something like
```ts
unpackTar(path.join(NODE_MODULES, pkg.name), { strip: 1 })
```

tsdown.config.ts
sacrosanctic: Had to use an LLM.

jycouet (approved):

AdrianGonz97 (approved):
We're missing a changeset for this, but aside from that, everything looks good!

Conduitry:
I don't think we really need a changeset, unless we're worried it's _not_ actually a drop-in replacement and mentioning this in the changelog will make people waste less time scratching their heads over why stuff is breaking.

AdrianGonz97:
hmmm, unless I'm blind, their [README](https://github.com/ayuhito/modern-tar/blob/main/README.md) doesn't explicitly say that it's a drop-in replacement for `tar-fs`. In that case, I think adding a changeset should be fine just in case issues do crop up.

PR #1212, opened 2026-08-01
feat(sv): allow custom package manager in testing commands

GauBen:
### Description

As of now, `pnpm` is hardcoded in the testing setup. This PR enables setting a custom package manager.

### Checklist

### DIFF
--- a/.changeset/violet-coins-like.md
+++ b/.changeset/violet-coins-like.md
@@ -1,0 +1,5 @@
+---
+"sv": patch
+---
+
+feat: allow custom package manager in testing commands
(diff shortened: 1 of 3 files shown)


AdrianGonz97:
I think this change requires a bit more nuance. We chose `pnpm` as the default specifically because of workspaces. `setupTest` inserts a `pnpm-workspace.yaml` file into the test's root directory so that the test variants can have their dependencies installed all at once. 

In order to support other package managers, we would have to add their respective workspace configurations as well to match our current `pnpm-workspace.yaml` (e.g. `npm`/`yarn`/`bun` has `workspaces` defined in the `package.json`, while `deno` may have it as `workspace` in `deno.json`).

GauBen:
Indeed I missed that aspect

I fixed my original issue another way:

- `file:` linking in tests https://github.com/GauBen/vite-plugin-svelte-md/blob/677795d62bf2aef903f2735b41188fc6da345c62/src/sv/index.ts#L23-L25
- installing `pnpm` in CI https://github.com/GauBen/vite-plugin-svelte-md/blob/677795d62bf2aef903f2735b41188fc6da345c62/.github/workflows/NodeCI.yml#L38

It could probably be more elegant but it's one less thing to worry about here

PR #1216, opened 2026-08-01
feat(cli): Community add-ons no longer require scoped package names

### LINKED ISSUE #1101: npx sv create --template addon shouldn't require an organization
Maybe make it an option?

sacrosanctic:
Closes #1101

### Description

It currently accepts `@<org>`, `<pkg>`, and `@<org>/<pkg>`. Also went a little overboard with the name validation. We can pare it down.

### Checklist


packages/sv/src/cli/tests/cli.ts
```diff
@@ -61,6 +61,11 @@ describe('cli', () => {
 			projectName: '@my-org/sv',
 			template: 'addon',
 			args: []
+		},
+		{
+			projectName: 'my-addon',
+			template: 'addon',
+			args: []
 		}
```
jycouet: Do we need a test for this? (Right now, CI is failing probably because snapshots are missing).

I think that having 1 test for addon is ok (if it's only the name update)

jycouet (changes requested):

packages/sv/src/core/validate-project-name.ts
```diff
@@ -0,0 +1,32 @@
+// https://npmx.dev/package/validate-npm-package-name#user-content-naming-rules
+export const validateProjectName = (value: string | undefined): string | undefined => {
```
jycouet: This validation of package name is not really belonging to this PR (tests as well)

jycouet:
Here we would need to adjust docs & a changeset (minor) as it's a significative change.

sacrosanctic:
Somebody with a bigger computer update the snapshots.

packages/sv/src/cli/create.ts
```diff
@@ -272,29 +272,24 @@ export async function createProject(cwd: ProjectPath, options: Options) {
 	const parentDirName = path.basename(path.dirname(projectPath));
 	let projectName = parentDirName.startsWith('@') ? `${parentDirName}/${basename}` : basename;
 
-	if (template === 'addon' && !projectName.startsWith('@')) {
```
sacrosanctic, line 275: @jycouet The issue with the tests is that previously, prompt was skipped if the project path is valid, defined by `@my-org/sv'` which produces `.test-output/cli/@my-org/sv`. Now the range of valid names have expanded and so I removed this change.

2 directions to go:
- add a `--addon-name` arg so it can be defined in the test (plus users can use it for promptless)
- add validation based on path (a little clunky, me thinks)

sacrosanctic: I implemented `--addon-name` just to get a passing CI, still looking for maintainer to decide on a direction

jycouet: ok I see 👍

jycouet (approved):

PR #1223, opened 2026-08-05
chore: run addon tests on request

manuel3108:
The addon tests are the only ones that take quite a lot of time to run. So we run them in a separate runner to get faster general feedback. Additionally most of the changes we do, do not affect the addons itself. In such cases it is wasteful to run the addon tests. Therefore we add a new label [`needs-addon-integration-tests`](https://github.com/sveltejs/cli/labels) similar to the `needs-platform-tests` in kit, that we can add to the prs on a per need basis. Changeset PR's automatically receive that label.

### Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -64,7 +64,6 @@ jobs:
           node-version: 22
           cache: pnpm
       - run: pnpm install --frozen-lockfile
-      - run: pnpm exec playwright install chromium
       - run: pnpm build
       - run: pnpm test:ci
   addon-integration-tests:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,6 +14,7 @@ on:
       - 'LICENSE'
       - 'README.md'
   pull_request:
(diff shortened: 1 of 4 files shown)


jycouet:
No luck with the 11 minutes? or?!
In average I think that it's ~7 min the longest... maybe we should just split on more machine but all the time?

I did try https://www.blacksmith.sh/ on a private project and was amazed by the speed gain! should we have an eye on it?

Label can work technically, but is not that practical.

manuel3108:
Not sure what exactly you mean? Our average CI pipeline takes slightly below 7 minutes. The testes without the addons only take two. As soon as you want the addons to be tested you add the label and wait another 3-5 minutes. Seems like the 11 minutes you were seeing were just a really unlucky macos runner. Just re-ran it and it came out at 5 minutes.

I don't think we need to consider alternatives here. We just re-use the available resources respectfully. And since we do have the addon snapshots included in the cli tests, i don't think running the addon tests everytime is useful.

jycouet (approved):
We can try this flow 👍

PR #1224, opened 2026-08-05
fix(addon): relax community addon requirements

### LINKED ISSUE #1210: Relax `sv add` requirements
Hi there!

I'm trying to create a `sv add` addon for https://github.com/ota-meshi/vite-plugin-svelte-md but it would make my life simpler if some requirements were relaxed:

- The package being published as `vite-plugin-svelte-md` on npm, I would like `npm i` and `sv add` command to point to the same name, *i.e.* `sv add vite-plugin-svelte-md`

  Covered by https://github.com/sveltejs/cli/issues/1101

- Allow a non-empty `dependencies` record in package.json. I know that they won't be installed, but the `./sv` entrypoint will be self-contained. The `dependencies` record is for the root entrypoint.

  https://github.com/sveltejs/cli/blob/78f6b7af93c9ae61cb8ae2eb3293a560298cbc75/documentation/docs/30-add-ons/99-community.md?plain=1#L153

  https://github.com/sveltejs/cli/blob/78f6b7af93c9ae61cb8ae2eb3293a560298cbc75/packages/sv/src/core/fetch-packages.ts#L29-L33

  Also, as [@sveltejs/sv-utils](https://npmx.dev/package/@sveltejs/sv-utils) is a dependency of `sv`, it should be possible not to bundle it as well

You can find my WIP here: https://github.com/ota-meshi/vite-plugin-svelte-md/pull/157

sacrosanctic:
Closes #1210

### Description

My initial take is to just relax the check for multi export repos. This doesn't really flag the problem if the addon dev decides to use those dependencies.

I considered static analysis, but it seems complicated to do.

### Checklist

### DIFF
--- a/packages/sv/src/core/fetch-packages.ts
+++ b/packages/sv/src/core/fetch-packages.ts
@@ -1,4 +1,4 @@
-import { color, coerceVersion, downloadJson } from '@sveltejs/sv-utils';
+import { color, coerceVersion, dedent, downloadJson } from '@sveltejs/sv-utils';
 import { unpackTar } from 'modern-tar/fs';
 import fs from 'node:fs';
 import { platform } from 'node:os';
@@ -13,9 +13,18 @@ import type { AddonDefinition, AddonReference } from './config.ts';
 // path to the `node_modules` directory of `sv`
 const NODE_MODULES = fileURLToPath(new URL('../../node_modules', import.meta.url));
 
+function isStandalone(addonPkg: Record<string, any>): boolean {
+	const exports = addonPkg.exports;
+	if (!exports) return true;
+	if (typeof exports === 'string') return true;
+	if (typeof exports === 'object') return Object.keys(exports).length === 1;
+	return true;
+}
+
 function verifyPackage(addonPkg: Record<string, any>, specifier: string): string | undefined {
 	const peerDeps = { ...addonPkg.peerDependencies };
 	const deps = { ...addonPkg.dependencies };
+	const standalone = isStandalone(addonPkg);
 
 	// valid addons should always have `sv` as a peerDependency
 	const addonSvVersion = peerDeps['sv'];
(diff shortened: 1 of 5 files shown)


packages/sv/src/core/fetch-packages.ts
```diff
@@ -25,10 +34,12 @@ function verifyPackage(addonPkg: Record<string, any>, specifier: string): string
...
 
-	// addons should not have any dependencies (everything should be bundled)
-	if (Object.keys(deps).length > 0) {
+	// standalone addons must not have dependencies (everything should be bundled)
+	if (standalone && Object.keys(deps).length > 0) {
 		throw new Error(
-			`Invalid add-on package detected: '${specifier}'\nCommunity add-ons should not have any 'dependencies'. Use 'peerDependencies' for 'sv' and bundle everything else`
+			dedent`
+				Invalid add-on package detected: '${specifier}'
+				Standalone add-ons must not have 'dependencies' in package.json. Everything should be bundled with your add-on.`
```
AdrianGonz97, line 42: This check long predates the `./sv` export condition. Now that add-ons can co-exist with other libraries in the same package, this check is very much out-of-date.

I think it would be best to just remove this check altogether. However, it would also be nice to add it elsewhere. Perhaps our add-on template can have a post-build check built into it to help guide add-on devs in the right direction?

sacrosanctic: > a post-build check

like a warning? I was thinking we could have an `sv validate` cmd for this

AdrianGonz97:
@sacrosanctic opened #1225 as an alternate way of checking for unbundled deps. this is a different approach from https://github.com/sveltejs/cli/pull/1224#discussion_r3722463659

jycouet:
I think that we should close here in favor of https://github.com/sveltejs/cli/pull/1225 direction _(not mergable yet)_

PR #1231, opened 2026-08-07
fix(migrate): correctness fixes in the migrate CLI and sveltekit-3 tasks

jycouet:
Six independent bugs found while reviewing `feat/update-templates-to-kit-3`. None are caused by that branch - they all live on `version-1` already.

- **Ctrl+C at the migrate confirm prompt ran the migration anyway.** clack returns a truthy symbol on cancel, so `!proceed` was never true.
- The sveltekit-3 env task emitted `@sveltejs/kit/hooks` (the pre-kit-3 specifier) instead of `@sveltejs/kit/env`, and hardcoded `src/env.ts` in JS projects.
- The tsconfig task bailed - silently dropping migrated `include` paths - when a prerequisite task had already retargeted `extends`.
- `derefStores` renamed identifiers by name only, so `obj.$page` and `{ $page: x }` were rewritten too.
- A `-1` `splice` in the snapshot test dropped an unrelated entry, masking the assertion it exists for.
- The ai-tools warning listed `AGENTS.md` once per selected IDE.

`--project migrate --project sv-utils`: 289 passed.

One further issue is **not** fixed here: `sveltekit-3/tasks/svelte-config.ts` unlinks `svelte.config` outside the `sv` API, so a `--files` filter that suppresses the `vite.config` write leaves the project with no config at all. A correct fix needs an `sv.remove` that honours `filesFilter`.

### DIFF
--- a/packages/sv/src/cli/migrate.ts
+++ b/packages/sv/src/cli/migrate.ts
@@ -172,7 +172,7 @@ async function determineTasks(
 	};
 
 	try {
-		migration.setup(setupOptions);
+		await migration.setup(setupOptions);
 	} catch (err) {
 		common.errorAndExit(err instanceof Error ? err.message : String(err));
 		return;
@@ -196,7 +196,7 @@ async function determineTasks(
 			}
 		}
 	};
-	migration.collect(collectOptions);
+	await migration.collect(collectOptions);
 
 	if (allTasks.length === 0) {
 		common.errorAndExit(`Migration "${migration.id}" did not return any tasks to run.`);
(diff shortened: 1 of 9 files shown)


packages/sv/src/cli/migrate.ts
```diff
@@ -251,6 +251,12 @@ async function determineTasks(
 			initialValue: false
 		});
 
+		// a cancel is a truthy symbol, so it has to be checked before the falsy "no"
+		if (typeof proceed === 'symbol') {
```
manuel3108: any reason for not using this here?

```suggestion
		if (isCancel(proceed)) {
```

Seems like we are doing this mistake multiple times in the codebase.

manuel3108 (approved):
LGTM

PR #1235, opened 2026-08-07
chore: replace `prettier` for `oxfmt`

AdrianGonz97:
### Description

Most files should stay the same with perhaps a minor formatting difference in some. 

A notable change is the difference in field sorting for the `package.json`s. `oxfmt` uses it's [own sorting algorithm](https://github.com/oxc-project/sort-package-json#field-ordering) that's not very customizable at the moment, but it still works pretty well out of the box.

I also had to change the add-on template's package name placeholder to a slightly different variation due to `oxfmt` trying to format `~SV-PROTOCOL-NAME-TODO~` into `~~SV-PROTOCOL-NAME-TODO~~` in markdown files. The new placeholder name is now: `$SV-PROTOCOL-NAME-TODO$`.

### Checklist


packages/sv-utils/package.json
```diff
@@ -1,29 +1,32 @@
...
 	"repository": {
 		"type": "git",
 		"url": "git+https://github.com/sveltejs/cli.git",
 		"directory": "packages/sv-utils"
 	},
-	"homepage": "https://svelte.dev",
-	"scripts": {
-		"check": "tsgo",
-		"format": "pnpm lint --write",
-		"lint": "prettier --check . --config ../../prettier.config.js --ignore-path ../../.gitignore --ignore-path ../../.prettierignore"
```
jycouet, line 16: We don't format & lint at package level anymore? Only from root?

packages/sv/src/addons/experimental.ts
```diff
@@ -8,8 +10,6 @@ import {
 	svelteConfig,
 	transforms
 } from '@sveltejs/sv-utils';
-import fs from 'node:fs';
-import path from 'node:path';
 import { defineAddon, defineAddonOptions } from '../core/config.ts';
```
jycouet: For these formating we used `@trivago/prettier-plugin-sort-imports`. There is not something equivalent yet?

packages/sv-utils/package.json
AdrianGonz97: correct! if you're cd'd into a specific package, you can just run `pnpm -w format` to run the `format` script from a sub package. same goes for linting

packages/sv/src/addons/experimental.ts
AdrianGonz97: it does support import supporting, but the default sorting algo is slightly different. by default, imports for native node api's are sorted to the top of the list

sacrosanctic:
@E021ntox This is for the repo itself. It will be invisible to the users of the CLI.

jycouet:
> Got it. But I wanted to let you know that users actually need oxlint and oxfmt more.

Please, stop commenting non-stop on random things random stuff. We got your point.

teemingc (approved):

teemingc:
We should probably switch the add-on from prettier to oxfmt at some point otherwise the template files will be formatted to a different style after the user runs prettier on them in their app.

sacrosanctic:
Should be fine, don't we format the files after an add-on install?

jycouet (approved):
It looks all good, thx

PR #1236, opened 2026-08-07
feat: more helpers for analyzing and manipulating ASTs

dummdidumm:
Analyzed what the migration tasks use, and extracted stuff from it that can be reused across them


manuel3108 (approved):
Super cool, thanks!

PR #1239, opened 2026-08-08
feat(migrate): `collect-migration-instructions` for `sk3`

manuel3108:
This is something that was discussed weeks ago in discord. The idea is to provide the ai with hints what still needs to be migrated, because it was not statically analyzable or not enough users will be encountering this.

If we decide to move forward with this, we should follow this list: https://github.com/sveltejs/cli/issues/1124#issuecomment-5218979158

The question is - do we want to move forward with this? I like the idea, but im not too sure the acceptance will be there and its worth the effort. It will still require us to go through the relevant breaking changes and curate them manually. The two migration tasks for testing were generated by ai, but both had fatal flaws that i needed to correct to be correct and clear checks.

Also i hate the current wording, `prepare-ai-migration` is the best i came up, but i still hate it. Also the "migration tasks" feel weired, since this is already a task so we would technically have subtasks, which are probably going to cause even more confusion.

What do you think?
If someone wants to pick it up from here - feel free.

### Checklist


dummdidumm:
I think this is a great idea. That way you could run a migration from this prompt: "Use npx sv migrate sveltekit-3 to migrate. This will automatically update code for some breaking changes and give you instructions for the rest. Follow these instructions to complete the migration.".

I would call this "collect-migration-instructions".

dummdidumm (approved):

PR #1241, opened 2026-08-09
feat(migrate): move Kit 3 setup out of the experimental add-on

manuel3108:
Relates #1124 

### Description

Given that `sv@1` now targets `kit@3`, the `experimental` addon needs to change. Well not directly, but given that we will publish `sv migrate sveltekit-3` in the same step, we can safely assume that the user is either on `kit@3` or will use the migration to get there. Therefore the migration no longer needs to bump the peer deps or change any `$lib`/`#lib` stuff. Both things are therefore moved into its own migration task.

The only remaining task of the `experimental` addon is to enable `experimental` options. Thats it. As it should have been.

### Checklist

### DIFF
--- a/packages/sv/src/migrate/migrations/sveltekit-3/tasks/lib-alias.ts
+++ b/packages/sv/src/migrate/migrations/sveltekit-3/tasks/lib-alias.ts
@@ -1,7 +1,7 @@
 import { transforms } from '@sveltejs/sv-utils';
 import { defineMigrationTask } from '../../../index.ts';
 
-const LIB_ALIAS = /\$lib(?=\/|['"`])/g;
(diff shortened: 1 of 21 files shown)


packages/sv/src/migrate/migrations/sveltekit-3/tasks/lib-alias.ts
```diff
@@ -0,0 +1,32 @@
+import { transforms } from '@sveltejs/sv-utils';
+import { defineMigrationTask } from '../../../index.ts';
+
+const LIB_ALIAS = /\$lib(?=\/|['"`])/g;
```
dummdidumm:
```suggestion
const LIB_ALIAS = /(?<=['"`])\$lib(?=\/|['"`])/g;
```

dummdidumm (approved):

PR #1244, opened 2026-08-10
fix: adjust paraglide add-on

dummdidumm:
See https://github.com/opral/paraglide-js/issues/737 - this is necessary to adjust since the event object has readonly properties in SvelteKit 3

### DIFF
--- a/.changeset/flat-pianos-create.md
+++ b/.changeset/flat-pianos-create.md
@@ -1,0 +1,5 @@
+---
+'sv': patch
+---
+
+fix: adjust paraglide add-on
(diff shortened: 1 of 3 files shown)


.changeset/flat-pianos-create.md
```diff
@@ -0,0 +1,5 @@
+---
+'sv': patch
+---
+
+fix: adjust paraglide add-on
```
AdrianGonz97:
```suggestion
fix(paraglide): adjust paraglide add-on
```

AdrianGonz97 (approved):

PR #1250, opened 2026-08-11
feat(migrate): imports migration

dummdidumm:
Various imports were moved in SvelteKit 3, this updates usage sites

WIP because there may be more changes


manuel3108 (approved):
LGTM

PR #1252, opened 2026-08-11
fix(migrate): various migration fixes

dummdidumm:
See the individual commits


manuel3108:
svelte.dev check failing, but unrelated to this change, so ignoring.

manuel3108 (approved):

PR #1271, opened 2026-08-14
fix(drizzle): bump better-sqlite3 types to v9

### LINKED ISSUE #1253: Bump `@types/better-sqlite3` from v7 to v9
> No API changes; only the version field is updated.

- https://github.com/DefinitelyTyped/DefinitelyTyped/pull/75271

Still confused why the `@types` version lags the package version this much, but bumping seems fine.

manuel3108:
Closes #1253

## What changed

- bump the generated `@types/better-sqlite3` dependency from v7 to v9 for the Drizzle `better-sqlite3` setup
- add a patch changeset for `sv`

DefinitelyTyped now publishes the existing declarations under the v9 package line. There are no declaration API changes, but generated projects were still pinned to the older v7 line.

## Validation

- `pnpm --filter sv check`
- Prettier check for the changed files
- confirmed generated Drizzle projects resolve `@types/better-sqlite3@9.6.0`

The full Drizzle integration suite reached dependency installation, but pnpm stopped because the existing `better-sqlite3` native build script is not approved in the generated test workspace.

### DIFF
--- a/.changeset/bright-dragons-smile.md
+++ b/.changeset/bright-dragons-smile.md
@@ -1,0 +1,5 @@
+---
+'sv': patch
+---
+
+fix(drizzle): bump `@types/better-sqlite3` to v9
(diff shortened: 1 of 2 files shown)


manuel3108:
What the heck codex, don't you think one pr of the same is enough?

PR #1274, opened 2026-08-14
fix(migrate): allow `@sveltejs/kit` to be a dependency

### LINKED ISSUE #1270: @sveltejs/kit is not a devDependency
`npx sv@next migrate sveltekit-3 --tasks all --confirm`

outputs:

```
j@scha:~/dev/arm/systemkollektiv/config$ npx sv@next migrate sveltekit-3 --tasks all --confirm
Need to install the following packages:
sv@1.0.0-next.1
Ok to proceed? (y) y
npm notice run @systemkollektiv/config@0.0.233 npx
npm notice run 'sv' migrate sveltekit-3 --tasks all --confirm

HINT: Run "sv --help" to get the full list of commands, add-ons, and examples to one-shot and skip interactive prompts.
┌  Welcome to the Svelte CLI! (v1.0.0-next.1)
│
■  @sveltejs/kit is not a devDependency in package.json - this doesn't look like a SvelteKit project.
│  Point to one with --cwd <path>, or see sv migrate --help.
│
└  Operation failed.
```

package.json dependencies includes it though:
```
"dependencies": {
    "@eslint/js": "10.0.1",
    "@magic/format": "0.0.75",
    "@sveltejs/adapter-node": "5.5.7",
    "@sveltejs/adapter-static": "3.0.10",
    "@sveltejs/kit": "2.70.2",
    "@sveltejs/package": "2.5.8",
    "@sveltejs/vite-plugin-svelte": "7.3.0",
    "@systemkollektiv/i18n": "0.0.134",
    "@tailwindcss/vite": "4.3.3",
    "@types/eslint": "9.6.1",
    "@types/node": "26.2.0",
    "@types/three": "0.185.4",
    "eslint": "10.8.1",
    "eslint-plugin-svelte": "3.22.0",
    "esm-env": "1.2.2",
    "eventsourcemock": "2.0.0",
    "globals": "17.9.0",
    "prettier-plugin-svelte": "4.1.1",
    "publint": "0.3.23",
    "svelte": "5.56.8",
    "svelte-check": "4.7.5",
    "tailwindcss": "4.3.3",
    "typescript": "npm:@typescript/typescript6",
    "typescript-eslint": "8.66.0",
    "vite": "8.2.1"
  },
```

kinda clear why this errors, but i think this should be a supported usecase:
@systemkollektiv/config is our central svelte package that all our other packages and pages depend on, it handles one central svelte/vite/eslint/tests/tailwind/prettier/typescript configuration for around 40 packages, reducing the boilerplate we have to change for updates to the config and making sure our various config files do not drift apart.

maybe sv should check for existence of the @sveltejs/kit dir in node_modules instead? or read package-lock.json if @sveltejs/kit is a dependency of a dependency?

i guess the next issue would be that we have vite.js and svelte.js files in src/lib, our configuration just exports from there:

```js
// svelte.config.js
import { svelte } from '@systemkollektiv/config'
export default svelte

// vite.config.js
import { vite } from '@systemkollektiv/config'
export default vite
```
this works suprisingly well for development, can even be simply extended to create custom aliases or other special settings that the projects might need, but it means sv will throw errors, the devDepdency error only being the first, afterwards:

```
▲  Task 'tsconfig' failed: Unable to process 'svelte.config.js'. Reason: Unable to find exported variable 'svelte'
│
└  Migration failed.
```

ps: unfortunately the @systemkollektiv namespace lives on our private gitlab instance, if there is interest in fixing this i can create two reproduction repositories, i totally understand if this is not a usecase that svelte wants to support, it's a very special case that has a lot of downsides too (harder to extend, indirection when trying to debug config issues, have to publish a new config version to fix issues in downstream, making testing harder etc)

sacrosanctic:
Closes #1270

### Description

A migration is not the right place to police this.

### Checklist

### DIFF
--- a/.changeset/fair-games-tickle.md
+++ b/.changeset/fair-games-tickle.md
@@ -1,0 +1,5 @@
+---
+"sv": patch
+---
+
+fix(migrate): allow `@sveltejs/kit` to be a dependency when migrating
--- a/packages/sv/src/migrate/migrations/app-state/index.ts
+++ b/packages/sv/src/migrate/migrations/app-state/index.ts
@@ -8,9 +8,9 @@ export default defineMigration({
 	setup: ({ pkg }) => {
 		const kitPackageName = '@sveltejs/kit';
 
-		if (!pkg.devDependencies?.[kitPackageName])
+		if (!pkg.devDependencies?.[kitPackageName] && !pkg.dependencies?.[kitPackageName])
 			throw new Error(
-				`${color.command(kitPackageName)} is not a devDependency in package.json - this doesn't look like a SvelteKit project.\n` +
+				`${color.command(kitPackageName)} is not in package.json - this doesn't look like a SvelteKit project.\n` +
 					`Point to one with ${color.command('--cwd <path>')}, or see ${color.command('sv migrate --help')}.`
 			);
 	},
--- a/packages/sv/src/migrate/migrations/sveltekit-3/index.ts
+++ b/packages/sv/src/migrate/migrations/sveltekit-3/index.ts
@@ -20,13 +20,14 @@ export default defineMigration({
 	setup: ({ pkg, requires }) => {
 		const kitPackageName = '@sveltejs/kit';
 
-		if (!pkg.devDependencies?.[kitPackageName])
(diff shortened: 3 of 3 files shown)


manuel3108:
I don't think this is necessary. Also the issue describes a really special case that we should not support IMHO.

PR #1278, opened 2026-08-16
fix(addon): harden add-on entry point resolution

jycouet:
Builds on #1233. Same direction (detect the real failure instead of proxying it via `dependencies`), with the resolution path tightened.

- **Try `./sv` then `.`.** A module-not-found on `./sv` currently throws before `.` is attempted, so a package whose `./sv` file is missing from the tarball fails even when `.` loads fine. The docs already promise this fallback.
- **Stop rejecting valid `exports` shapes.** A bare string (`"exports": "./dist/index.mjs"`), a conditions-only object, and an absent `exports` (resolved via `main`) are all importable but were failing the pre-check. `exports` is now only consulted to decide whether to probe `./sv`, never to reject.
- **Better diagnosis.** CJS entries report `MODULE_NOT_FOUND` without the `ERR_` prefix, so the unbundled-deps hint never fired for them. Node's message is kept so the missing module is named, and the "no default export" case no longer throws an error with an empty bullet.

Also documents that `sv` provides `@sveltejs/sv-utils`, so an add-on may leave it unbundled at the cost of tracking whatever version `sv` ships. Nothing verifies it, by design.

Tests cover each `exports` shape, the `./sv` → `.` fallback, both module-not-found codes, and the missing-default case.


packages/sv/src/core/common.ts
```diff
@@ -278,7 +278,7 @@ export function errorAndExit(message: string) {
 	const [firstLine, ...restLines] = message.split('\n');
 
 	p.log.error(firstLine);
-	// Fixes issue where the first line of the error message is not the same color as the rest of the lines
+	// detail lines are dimmed so the first line reads as the headline
```
AdrianGonz97: i think the original comment is clearer in its purpose of fixing an issue
```suggestion
	// Fixes issue where the first line of the error message is not the same color as the rest of the lines
```

without the additional coloring, we get a mismatch of colors like this:
terminal output

PR #1161, opened 2026-07-04
docs: refactor the docs to remove the h4 header

sacrosanctic:
Related to https://github.com/sveltejs/svelte.dev/pull/2064

### Checklist

- Update [snapshots](../CONTRIBUTING.md#update-snapshots) (if applicable)
- Add a [changeset](../CONTRIBUTING.md#generating-changelogs) (if applicable)
- Allow maintainers to edit this PR
- I care about what I'm doing, no matter the tool I use (Notepad, Sublime, VSCode, AI...)

PR #1279, opened 2026-08-16
feat(create): make the npm org optional for the `addon` template

### LINKED ISSUE #1101: npx sv create --template addon shouldn't require an organization

jycouet:
Closes #1101.

The org prompt in `sv create --template addon` now accepts an empty answer and scaffolds an unscoped package. Publishing under an org stays the recommendation, since naming the package `sv` within it is what enables the `sv add @my-org` shorthand.

Also corrects the `### Package names` docs. They marked `npx sv add my-lib` as invalid, but `sv add` has parsed unscoped specifiers since the feature shipped and has never had a scope check. Two add-ons published today under the `sv-add` keyword are unscoped (`sveltekit-temporal`, `svforge`) and install fine.

For context on how the docs drifted: the requirement was only ever enforced in `create` (added in 618f1ad, relaxed to a prompt by #1031), while the docs stated it globally. A later docs refactor (#1161) then turned the hedged "unscoped packages are not supported yet" note into a hard `✗ BAD npx sv add my-lib` example, pointing at the one command that never enforced it.

Note that the prompt still blocks in non-interactive runs; that is pre-existing and left alone here.

### DIFF
--- a/.changeset/olive-hounds-repeat.md
+++ b/.changeset/olive-hounds-repeat.md
@@ -1,0 +1,5 @@
+---
+'sv': patch
+---
+
+feat(create): make the npm org optional for the `addon` template
--- a/documentation/docs/30-add-ons/99-community.md
+++ b/documentation/docs/30-add-ons/99-community.md
@@ -177,18 +177,15 @@ Your add-on must have `sv` as a peer dependency and **no** `dependencies` in `pa
 
 ### Package names
 
-Packages must be published under an npm org:
+Any npm package name works:
 
 ```sh
-# ✓ GOOD
 npx sv add @my-org/sv
 npx sv add @my-org/core
-
-# ✗ BAD
 npx sv add my-lib
 ```
 
-If your package is published with the `sv` scope, it can be omitted. The following all resolves to the same package:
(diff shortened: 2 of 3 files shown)


AdrianGonz97:
is it intentional that running `sv create @foobar/sv --template addon` creates this directory structure:
```
@foobar
└── sv
    ├── package.json
    ├── README.md
    └── ...other files
```

...where the org is its own directory.

if not, then I wonder what we should do instead? perhaps it should just create the `sv` dir without the org scope? perhaps `@foobar-sv` instead? i'm not sure what the right answer is at the moment

jycouet:
I'm glad that you bring this up, because I also don't know the best answer _(but I remember that this topic it not that trivial)_

I think that we should be able to distinguish 2 things: directory & package name.
Ideally, directoty: `sv` and package name: `@foobar/sv`

But `sv create I/want/this/location` is creating nested folder... so my changing?

Maybe if it's starting with `@` and having only one `/` we could create a folder `sv` only ?

_just thinking loud, I'm not sure_

sacrosanctic:
We can also distinguish between an empty `cwd` and an existing monorepo. Also, the current implementation has too much magic under the hood. We should explain it if we're going to create multiple nested folders.

jycouet:
I'm closing in favor of https://github.com/sveltejs/cli/pull/1216

PR #1290, opened 2026-08-21
fix: workaround for missing zimmerframe types

### LINKED ISSUE #1289: `zimmerframe` types are missing
it is declared as `zimmerframe`
https://github.com/sveltejs/cli/blob/4abd0900ea4b261a5c307db79f722914d3785b85/packages/sv-utils/api-surface.md?plain=1#L214
but exported as `index_d_exports`
https://github.com/sveltejs/cli/blob/4abd0900ea4b261a5c307db79f722914d3785b85/packages/sv-utils/api-surface.md?plain=1#L1016

sacrosanctic:
Closes #1289

### Checklist

### DIFF
--- a/.changeset/wise-deserts-press.md
+++ b/.changeset/wise-deserts-press.md
@@ -1,0 +1,6 @@
+---
+"@sveltejs/sv-utils": patch
+"sv": patch
+---
+
+fix: add zimmerframe types
--- a/packages/sv-utils/api-surface.md
+++ b/packages/sv-utils/api-surface.md
@@ -211,7 +211,7 @@ interface Dedent {
 }
 type CreateDedent = (options: DedentOptions) => Dedent;
 declare const dedent: Dedent;
-declare namespace index_d_exports {
+declare namespace zimmerframe {
 	export function walk<
 		T extends {
 			type: string;
@@ -1013,7 +1013,7 @@ export {
 	type SvelteConfigLocation,
 	type SvelteConfigObjects,
 	type TransformFn,
-	index_d_exports as Walker,
+	zimmerframe as Walker,
 	type YamlDocument,
 	coerceVersion,
 	color,
--- a/packages/sv-utils/api-surface.md
+++ b/packages/sv-utils/api-surface.md
@@ -211,7 +211,7 @@ interface Dedent {
 }
 type CreateDedent = (options: DedentOptions) => Dedent;
 declare const dedent: Dedent;
-declare module 'zimmerframe' {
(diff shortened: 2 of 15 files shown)


jycouet (commented):
Right now, it string-replaces generated output and hardcodes `index_d_exports`, a name `rolldown-plugin-dts` owns. I don't think that it's the way to go.

---

I think that we have 3 personas: 
- cli devs
- community add-on devs
- cli & add-on users

This issue is annoying for `community add-on devs`. We could instruct them what to install as devDeps to have a better DX?

- we should probably fix https://github.com/sveltejs/svelte/issues/17520 ASAP this is the REAL way to go.
- We could get rid of tsdown (using rolldown-plugin-dts) and move to JSDoc?

sacrosanctic:
Oops, did not mean to represent this as a through fix. Also made some adjustment to see if this is a passable short term fix if we're blocked upstream.

jycouet (approved):
I have to say that a few types inferred in this PR are a nice positive thing.

But it's not really fixing anything and I'm not sure how much is this a "problem" today.
You had an issue with zimmerframe and that's why you want to improve it? Because other libs have the same problem... (because of the same root cause) it's ok to not fix them?
I'm a bit worried that we add code that we don't remember that we can remove!

Anyhow, I'm approving because it's removing a lot of `as ___` but I would like to not generalize this, so if you think it's a big deal and other maintainers agree it will be merged.

sacrosanctic:
It does work a little better...but we still have to cast the  first param as `AstTypes.Node`. There isn't a lot of documentation around walking the `ast` so getting intellisense on the methods is nice.

sacrosanctic:
https://github.com/sacrosanctic/sv-addon/blob/f8fd5c1d45fc6c9b24c476a2f30f2fc85a0a261c/packages/svelte-meta-tags/src/index.js#L182-L185

For reference, gotta write out jsdoc comments.

jycouet:
I put this workaround in draft for now. Not sure it's worth to proceed. Let us know when we want to close or more forward.

sacrosanctic:
I mean I opened the PR so I want it in. I'm not seeing the major downside. Wouldn't it be a noop when it is fixed upstream?

jycouet:
One downside is this `scripts/generate-api-surface.js` tweak (that could change when other deps change). 
If you add a comment on top pointing to the real fix, and fix conflicts, we can maybe merge :)

jycouet:
I narrowed the change and throw if it's not the number we expect. Let's see! :=)

PR #1292, opened 2026-08-21
fix(migrate): specify full version ranges for `@next` packages

### LINKED ISSUE #1283: npm doesn't install `next` versions after migration
Might just be an `npm` thing but running `npx sv@next migrate sveltekit-3 --tasks all --confirm` on an old demo doesn't install the `next` versions of SvelteKit and the adapter until I delete the `package-lock.json` file and run `npm i`

AdrianGonz97:
Closes #1283

One small quirk is if they have a package that has `@sveltejs/kit@^2` as a peer-dependency (e.g. `better-auth`), then an install with `npm` will fail without the `--legacy-peer-deps` flag.

We _could_ pass it automatically as part of `npm install` (which _can_ work), however, all subsequent `npm install`'s by the user will fail if they don't provide the flag themselves, which can be confusing if they're not educated about it. I wonder if we should log out some info telling them about it? ...but then we'd have to be able to detect if the install failed for that very reason, so we may be falling down an annoying rabbit hole just for that.

### Checklist


jycouet (approved):
TIL ` ^x.y.z-next.0` is working!
Nice

PR #1300, opened 2026-08-27
feat(cli): Add `@sveltejs/enhanced-img` addon

### LINKED ISSUE #496: Add `@sveltejs/enhanced-img`
I had been holding off because I thought one of the issues might necessitate changing the API, but it looks like that's not necessary and things should be pretty stable now.

There's a `webp` and `png` in the demo app: https://github.com/sveltejs/cli/blob/d1f4432c0e31677ae754d50394547eeff820ddfe/packages/create/templates/demo/src/routes/%2Bpage.svelte#L3

We should just have the `png` and `enhanced-img` will generate the webp and avif automatically.

It would also be nice to add to the skeleton project as I think no one should be shipping unoptimized images. For the skeleton app, it potentially could be a question like:

What image optimization should be used?
- [ ] build-time optimization with `@sveltejs/enhanced-img`
- [ ] fetch from CDN with `@unpic/svelte`
- [ ] none

Finally, I don't really like that images are currently stored in `$lib/images`. This seems like it will just become a huge unorganized catch-all. Images can live in the folder where they're used alongside the `+page.svelte` that includes them. This should help keep things a bit more organized. At least this was one of Rich's original arguments in favor of `+page.svelte`. But we don't really demonstrate that and guide users down that path today

### LINKED ISSUE #1306: `detectPnpmMajor` checks `cwd` instead of `target` path
If you're working on this repo and runs `pn build && pn sv ../test --install pnpm`. It will create a `onlyBuiltDependencies` instead of `allowBuilds` because this repo is using pnpm 10.

https://github.com/sveltejs/cli/blob/311c4d6b3991a6a5ffdeb5bada71e383f863deb7/package.json#L43

This can be generalized to any situation where the `cwd` and the `target` doesn't have matching versions via https://pnpm.io/package_json#devenginespackagemanager.

It was also bugged on the `pnpm` side where child process [will inherit the parent version](https://github.com/pnpm/pnpm/issues/14309) but is now fixed.

### proposed fix

- pass in cwd `allowBuilds(packages,{cwd})`
- make the versions explicit `allowBuilds(packages, {pnpmVersion})`
- default to `os.tmpdir()`

`os.tmpdir()` will not fix `pnx sv add ...`, only `pnx sv create ...`.

```diff
export function detectPnpmMajor(): number | undefined {
	try {
		const out = execSync('pnpm --version', {
			encoding: 'utf-8',
			stdio: ['ignore', 'pipe', 'ignore'],
+			cwd: os.tmpdir()
		});
		return coerceVersion(out.trim()).major;
	} catch {
		return undefined;
	}
}
```

sacrosanctic:
Closes #496 #1306
Merge after #1304

### Description

`check.ts` always use `createProject()` instead of the low level `create()` and `add()`. This caused a circular error so I had to move the shared options to a different file.

This PR now also attempts to fix #1306 by accepting `cwd` in `pnpm.allowBuilds` which is a breaking change.

### Checklist


packages/sv/src/cli/create.ts
```diff
@@ -332,6 +334,12 @@ export async function createProject(cwd: ProjectPath, options: Options) {
 
 		loadedAddons = result.loadedAddons;
 		answers = result.answers;
+
+		if (template === 'demo' && !loadedAddons.some((a) => a.addon.id === 'enhancedImg')) {
+			const addon = getAddonDetails('enhanced-img');
+			loadedAddons.push(createLoadedAddon(addon));
```
teemingc: This doesn't seem to currently work. The test output projects for the failed CI show that the `enhanced` plugin is missing from the Vite config.

teemingc (dismissed):

packages/sv/src/cli/create.ts
sacrosanctic: Now that you metnion it. I think the test is bugged.

jycouet (approved):
LGTM now :)

@sacrosanctic you wanna check before merging?

packages/sv/src/addons/enhanced-img.ts
```diff
@@ -0,0 +1,51 @@
...
+		sv.files(
+			{ include: 'src/**/*.svelte', where: (content) => content.includes('<img') },
+			transforms.text(({ content }) => {
+				let next = content;
+				for (const [statement, binding, src] of content.matchAll(REGEX_IMPORTED_IMG)) {
+					const tag = new RegExp(`<img([^>]*?)src=\\{${binding}\\}`, 'g');
+					if (!tag.test(next)) continue;
+					// `src` can be resolve import paths
+					next = next
+						.replaceAll(tag, `<enhanced:img$1src="${src}"`)
```
sacrosanctic: This one is not quite right. 
imports uses `'./path/to/your/image.jpg?enhanced'`. I believe `?enhanced` can be omited when it is inlined.

packages/sv/src/addons/enhanced-img.ts
```diff
@@ -0,0 +1,51 @@
...
+						.replace(`\n\t${statement}`, '');
```
sacrosanctic: `\n\t` is this always correct?

jycouet: It should be better now

packages/sv/src/addons/enhanced-img.ts
jycouet: Should be good.

jycouet (approved):

Issue #1312, opened 2026-09-02
Community add-on caching bugs


sacrosanctic:
### describe the issue

https://github.com/sveltejs/cli/blob/03fc69b9a811b8e2213b4e9b0070766952408c99/packages/sv/src/core/fetch-packages.ts#L14

This constant is only correct for `sv`'s dev-checkout layout (`src/core/…` → `packages/sv/node_modules`).
In the published bundle the chunk lives at `sv/dist/engine-*.mjs`, which produces three problems:
1. Doubled `node_modules` path.
`…/node_modules/sv/dist/../../node_modules` resolves to `…/node_modules/node_modules`.
2. Global pnpm store pollution.
With `pnpm dlx sv`, `sv`'s install dir is inside the content
store (`~/.local/share/pnpm/store/v11/links/@/sv/<version>/<hash>/`), which is meant to be
immutable, content-addressed, and state-free. Every community add-on download writes mutable
state (`node_modules/node_modules/<addon>/`, extracted tarballs and symlinks) into the store:
    - shared across all projects and never invalidated (not on sv upgrade, not by pnpm store prune);
    - concurrent sv runs race on the same paths (one run's rmSync deletes what another extracts).
3. Stale symlink
It's not cleaned up due to the fork in the control flow

### Reproduction
```sh
# 1. add a local add-on by path — creates <cache>/node_modules/node_modules/<name> symlink
pnpm dlx sv create ./a --template minimal --types ts --no-install --no-download-check --add file:/path/to/my-addon
# 2. add the same package name from npm — extract collides with the symlink
pnpm dlx sv create ./b --template minimal --types ts --no-install --no-download-check --add my-addon
# => "Failed to resolve my-addon … /node_modules/node_modules/my-addon is not a valid path"

# observe store pollution:
ls "$(pnpm store path)"/links/@/sv/*/*/node_modules/node_modules
```

sacrosanctic:
Note that the fix does not resolve number 2. But I don't think it will come up.

PR #1315, opened 2026-09-03
fix(cli): install community add-ons in the user cache dir

jycouet:
Closes #1312.

Community add-ons were installed into `sv`'s own directory via a path computed for the source layout, which in the published bundle (`sv/dist/`) resolves to `node_modules/node_modules`, and under `pnpm dlx` sits inside pnpm's immutable content store.

They now install under `<cache>/sv/add-ons/<sv version>/`, each add-on in its own root (keyed by `name@version` for npm, by path for `file:`) containing the add-on, a link to the running `sv`, and a stub it is imported through. Roots are built in a temp dir and moved into place, so concurrent runs no longer race, npm add-ons are reused across runs, and a local add-on can no longer collide with a published one of the same name.

### DIFF
--- a/.changeset/brave-pugs-shave.md
+++ b/.changeset/brave-pugs-shave.md
@@ -1,0 +1,5 @@
+---
+'sv': patch
+---
+
+fix(cli): install community add-ons in the user cache dir instead of next to `sv`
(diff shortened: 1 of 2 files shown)


jycouet:
It was a POC / Demo

PR #1319, opened 2026-09-09
chore(sv): refactor old code (WIP)

### LINKED ISSUE #956: rework `shared` folders
> I'm thinking that the shared folder could even go away since we have a lot of utils now
> https://github.com/sveltejs/cli/pull/953#issuecomment-3977275187

- doesn't support folders
- output has duplicated entries
- `-typescript` is the only folder with a minus

sacrosanctic:
Closes #956

### Description

This is WIP, will be chipping away at this slowly. I wanted to use the CI and maintainers can take a peek to guide the direction.

The goal is the write code that effects `shared` and `templates`.

### Checklist


manuel3108:
Hey!
can you potentially layout a bit more of what you are planning to do? The current PR does not really reveal anything. 
I do completely agree that the `create` code in this repo is the worst one that is remaining, now that we get rid of the old `migrate`. And given that the migrations use the same `sv.file` api and stuff i do think it would feel natural to do the something similar for create and the different create templates. But migrations are not exactly an `addon` they just use a similar api.

Something similar should work for create I think, but I would prefer if we would discuss it first, before you start running in one direction. Do you have any concrete ideas how you would like to implement this?

Or did I misunderstand you and are you trying to do something completely different here?

sacrosanctic:
No concrete plans yet, just feeling it out. My original idea was some sort of internal add-on. `migrate` might be a better mental model.

Everything is really tangled together so it's hard to make atomic changes. I might start with rewriting the template construction logic first. As it has a lot of baggage of wanting them to be running projects.

PR #1320, opened 2026-09-11
fix(sv-utils): detect pnpm version from the target project, not the invoker cwd

### LINKED ISSUE #1306: `detectPnpmMajor` checks `cwd` instead of `target` path

cpruijsen:
Closes #1306

### Description

`pnpm.allowBuilds` now takes `{ cwd, pnpmVersion }` and detects the pnpm major from the target project, not the invoker cwd. `sv create` / `sv add` (`addPnpmAllowBuilds`) and the Cloudflare adapter pass the target path. When `cwd` is omitted, detection runs in `os.tmpdir()` so the invoker's pin is not inherited.

Previously, `pnpm.allowBuilds` chose between `allowBuilds` (pnpm 11+) and `onlyBuiltDependencies` (pnpm 10) by running `pnpm --version` with no `cwd`, so Node used the invoker's working directory. From this repo (`packageManager: pnpm@10.33.4`), `sv create ../test --install pnpm` wrote the legacy key even when the new project would run pnpm 11.

`pnpm --version` follows `package.json#packageManager` / `devEngines.packageManager` by walking up from `cwd`. This repo reports 10.33.4; `os.tmpdir()` reports 11.26.0; a directory pinned to `pnpm@11.0.0` reports 11.0.0 (same numbers under `pnpm exec`).

Decision: pass target `{ cwd }` from first-party callers, accept `{ pnpmVersion }` for explicit writes, default omitted cwd to `os.tmpdir()`.
Alternative: only `cwd: os.tmpdir()` inside `detectPnpmMajor` (the one-line diff in the issue).
Why: tmpdir-only does not fix `pnx sv add` in a project whose pin differs from global pnpm (as the issue notes). Can drop the tmpdir default and/or `{ pnpmVersion }` for a smaller public API; first-party `{ cwd }` alone fixes `sv create` and `sv add`.

### Checklist

- [x] Update snapshots (if applicable): `api-surface.md` regenerated; CLI create snapshots unchanged (in-repo test apps still see this repo's pnpm 10 pin)
- [x] Add a changeset (if applicable)


jycouet (approved):
Thank you @cpruijsen 
LGTM

PR #1330, opened 2026-09-19
fix(sv): key `officialAddons` by add-on id

### LINKED ISSUE #1310: questions about `sv` addon API
### 1. `dependsOn` and `runsAfter` restricted to `OfficialAddons`

Are they supposed to support community addons or not yet.

### 2. Casing mismatch between `OfficialAddons` keys and `addon.id`
The contents of `OfficialAddons` is camelCase while the contents of `addon.id` is kebab-case. Should we align them. This matters cause they're both used explicitly.

### 3. `runAfter` is missing from documentation
```ts
setup: ({ dependsOn, unsupported, addOption, isKit }) => {
```
https://svelte.dev/docs/cli/sv#defineAddon

jycouet:
Closes #1310

### Description

`officialAddons` keys are now the add-on ids, which are what `dependsOn` / `runsAfter` actually match against - so `runsAfter('sveltekitAdapter')` was a silent no-op. Both callbacks now also accept community add-on ids, and `runsAfter` is documented.

### Checklist


documentation/docs/50-api/10-sv.md
```diff
@@ -69,6 +70,8 @@ export default defineAddon({
 });
 ```
 
+`dependsOn` and `runsAfter` take add-on ids (`'sveltekit-adapter'`, `'better-auth'`, ...), including community ones. A missing `dependsOn` is installed automatically if it's an official add-on, otherwise it has to be part of the same command. `runsAfter` only affects ordering, and `dependsOn` implies it.
```
sacrosanctic:
```suggestion
Official add-ons included in `dependsOn` will be installed automatically. Community add-ons must be included as part of the same command.

`runsAfter` makes add-on ordering explicit, while `dependsOn` loosely infers it.

`dependsOn` and `runsAfter` both accepts official and community add-on ids. (`'sveltekit-adapter'`, `'better-auth'`, ...) 
```

manuel3108:
Ohhh didn't see the PR. The changes look good, but not sure if `dependsOn` and `runsAfter` should actually allow community addons. I did not test it, but i would have the feeling, that they might actually be able to pull in another untrusted community addon. On the other hand you could potentially also get away with this if you are bundling the other untrusted addon. Will need to spend some more time on this.

jycouet:
Every community addon passes through `resolveNonOfficialAddons` => **blocklist check** + the "not reviewed for malicious code" warning + the confirm prompt. So we can [ban](https://github.com/sveltejs/cli/blob/version-1/packages/sv/blocklist.json) the whole graph on our side when needed. 

With this PR, users need to specify all community addon. (community addon can't pull silently another community addon, we can loosen this later if we want :) ) we write: `Add it to the command explicitly.`

_Did also some other cleanups_

sacrosanctic:
Not a huge difference between first and second degree community add-on. They're both at the "please be careful level". It is nice that we are explicit about their upstream dependencies.

PR #1331, opened 2026-09-19
feat(sv-utils): add browser entry point

### LINKED ISSUE #1262: feat: Make `@sveltejs/sv-utils` browser-friendly
I would like to reuse the same logic based on `@sveltejs/sv-utils` both in the implementation of an addon and in factories that generate files for sandbox environments directly in the browser.

Currently, this is not possible because the Node.js version of the `yaml` dependency is bundled. As a result, the generated bundle contains the following import:

```js
import { createRequire } from "node:module";
```

Would it be possible to use a browser-compatible version/build of `yaml` instead, so that `@sveltejs/sv-utils` can also be used in browser environments?

jycouet:
Closes #1262

### Description

New `@sveltejs/sv-utils/browser` entry exposing the environment-agnostic subset (parsers, transforms, language tooling, pure helpers). The root entry keeps the Node-only APIs (fs, package managers, shell, colors) and re-exports the browser subset, so its surface is unchanged.

`yaml` is now resolved through its `browser` field: its `node` build is CJS, which rolldown pulls in with `import { createRequire } from 'node:module'` even when unused. Only observable delta for Node is `!!binary` decoding to `Uint8Array` instead of `Buffer`, which we never parse.

A test asserts the built browser entry and its chunks import no Node builtins, and round-trips yaml through the shipped bundle.

### Checklist


jycouet:
> Install the latest version of `sv` from [3bc2879](https://github.com/sveltejs/cli/commit/3bc2879b08344a2e3433f7a21502448195620331):
> 
> ```shell
> pnx https://pkg.svelte.dev/sv/c/3bc2879b08344a2e3433f7a21502448195620331 create
> ```
> 
> Open in `pkg.svelte.dev`: https://pkg.svelte.dev/repos/cli/pr/1331

@x0k could you try it?

x0k:
I tested it in the browser and ran the add-on locally - everything works.

manuel3108 (approved):
LGTM

Issue #1219, opened 2026-08-02
feat: Allow `dependsOn` to pass default add-on options if the user doesn't specify them


AdrianGonz97:
While authoring the sv community [add-on for `shadcn-svelte`](https://github.com/huntabyte/shadcn-svelte/pull/2826), I found myself wishing to pass predefined default options to add-on dependencies.

For example, rather than just defining an add-on dependency like this:
```ts
	// ...
	setup: ({ dependsOn }) => {
		dependsOn("tailwindcss");
	},
```

we could have something that allows default options:
```ts
	// ...
	setup: ({ dependsOn }) => {
		dependsOn("tailwindcss", { plugins: [] });
	},
```

so if the users calls the add-on _without adding the dependency_ like so:
```sh
sv add @shadcn-svelte
```
then the interactive prompt won't appear for them for `tailwindcss`.

Of course, if they do specify the dependency:
```sh
sv add @shadcn-svelte tailwindcss
```
then the prompt for `tailwindcss`'s options will appear like normal. Likewise, if the add-on just does `dependsOn("tailwindcss")`, then the prompt for options will also appear.

jycouet:
That would be great 👍 

One note: if multiple `dependsOn` has different options for the same add-on... first one is winning? A merge strategy might be complicated?

AdrianGonz97:
Great question! 

So I think we have 2 distinct cases that have to be resolved:

## Sibling add-ons with conflicting options
If 2 add-ons (`A` and `B`) are specified and they both depend on the same add-on (`tailwindcss`), then we should probably just fallback to prompting the user directly rather than silently discarding one of the add-on's defaults for the other. 

```mermaid
flowchart LR
  U["sv add A B"] --> A
  U --> B
  A -->|"plugins: []"| C["tailwindcss"]
  B -->|"plugins: ['typography']"| C
  C --> R["prompt user"]
```


## Nested add-on deps with conflicting options
If the dependency tree has nested add-ons with conflicting options, then the add-on closest to the root should take precedence (`A` in the below example), which _should_ usually be the add-on that's specified by the user.

```mermaid
flowchart LR
  U["sv add A"] --> A
  A -->|"plugins: []"| C["tailwindcss"]
  A --> B
  B -->|"plugins: ['typography']"| C
  C --> R["resolves to plugins = []\n(A wins over B)"]
```

---

### Merging non-conflicting options
[...]

PR #1333, opened 2026-09-19
feat(sv): type the options of official add-ons (alternative manual)

### LINKED ISSUE #1217: docs: `sv.add` Drizzle ORM options for `client`
Turns out this is wrong:

```ts
await sv.add({
  cwd,
  addons: {
    drizzle: sv.officialAddons.drizzle,
  },
  options: {
    drizzle: { database: "sqlite", sqlite: "client-sqlite3" },
  },
  packageManager: "pnpm",
});
```

This is the fix:

```diff
- drizzle: { database: "sqlite", client: "better-sqlite3" },
+ drizzle: { database: "sqlite", sqlite: "better-sqlite3" },
```

I've been using a `client` key to match the CLI script:

```bash
npx sv add drizzle="database:postgresql+client:postgres.js"
```

I did name this as a docs issue, but would prefer the API to match CLI.

jycouet:
Closes #1217

### Description

`officialAddons` is no longer widened to `Addon<any>`, so add-on options are typed. `add({ addons, options })` now rejects `drizzle: { database: 'sqlite', sqlite: 'client-sqlite3' }` (and `client:`, which is a CLI group name, not an option key) instead of accepting anything.

The option values are hand-written in `packages/sv/src/addons/options.ts` rather than inferred: inferring them would drag every add-on's option literals into the published `.d.ts` (`api-surface.md` 297 -> 1564 lines) and force the dts generator from oxc to tsgo. `core/tests/addon-options.ts` fails type-checking as soon as the two drift apart.

Groundwork for #1219: `dependsOn('tailwindcss', { plugins: [] })` needs these types.

### Checklist


manuel3108:
Honestly, I'm not in favor of this one at this point. I will need to have a closer look at this in the next couple of days, but moving the addon-option-types into the core just feels wrong. The whole goal of the addons is to keep everything separate, otherwise we will end up the way `svelte-add` did with all the code spread around everywhere. 

Will we potentially be able to work around this in changing the way the options in the addons are declared?

jycouet:
It's probably not well laid out! sorry :D
A long long time ago, in API mode, `sv` was typed, and then we lost this _(a bit by accident I would say)_.

I tried: 
### A/ fully inferred https://github.com/sveltejs/cli/pull/1334
Switching the generator `.d.ts` from `oxc` to `tsgo` (in `tsdown`). But this is adding a lot of other things in d.ts that we probably don't want. I will do a PR to show it's probably better :)

### B/ manual types https://github.com/sveltejs/cli/pull/1333
This is this PR. Having to author 1 manual type for all official addon. _(with a test to avoid drifts!)_
To me it's not "core", files should probably move to a better location.
I'll update the branch to see where it could go

### C/ exporting options https://github.com/sveltejs/cli/pull/1335
It's also something that I looked quickly. But yes, we should probably look at this before 1.0... In case it's better... And changing a bit the way :) PR s00n

jycouet:
In order of preference for me right now: 
- #1333 (this one)
- #1335
- #1334

@manuel3108 @AdrianGonz97 @sacrosanctic ?

sacrosanctic:
They're all a little clunky in their own way. I'm slightly leaning towards `alternative export`. What was the method used when it was last working?

jycouet:
> They're all a little clunky in their own way. I'm slightly leaning towards `alternative export`. What was the method used when it was last working?

I agree that the 3 options have their flaws. 
It was working before tsdown, not using oxc... 

`alternative export` is a serious option yes 😏

PR #1336, opened 2026-09-20
fix(migrate): update @sveltejs/package for SvelteKit 3

### LINKED ISSUE #1299: `@sveltejs/package` is not updated during SvelteKit 3 migration
It might make sense to update `@sveltejs/package` if it exists in the project when performing a SvelteKit 3 migration because the migration updates to TypeScript 6 if it exists but `@sveltejs/package` v2 still depends on a version of `svelte2tsx` which uses TypeScript 5 only. Not a huge issue but it reduces one step for library authors

erenbati:
## What changed

When the SvelteKit 3 migration upgrades TypeScript to 6, also update an installed `@sveltejs/package` older than `2.5.8`.

`@sveltejs/package@2.5.8` added TypeScript 6 support, and the current SvelteKit 3 library template already uses `^2.5.8`. This keeps the migration on the existing package major while avoiding the TypeScript 5-only dependency combination described in #1299.

The package-json migration fixture now covers `@sveltejs/package@2.5.7 -> ^2.5.8`.

Fixes #1299.

## Validation

No local test run yet. Opening as draft to use CI for validation before marking ready.


jycouet (approved):
Add only if it's there 👍 
Thank you

